PR #46, opened 2021-07-02
add comments about thread model and thread safety

wikimonkey:
**Issue #, if available:**

**Description of changes:**
Add comments to explain the threads on which the methods are executed; note down the measures to ensure thread safety

for the sake of code maintainability

no new code

### DIFF
--- a/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
@@ -115,6 +115,8 @@ public class CertificateManager {
      * @throws KeyStoreException if unable to access KeyStore
      * @throws CsrProcessingException if unable to process CSR
      */
+    // the service certificate subscribe function is invoked on the thread starting broker, i.e. nucleus
+    // lifecycle management thread
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToServerCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate> cb)
             throws KeyStoreException, CsrProcessingException {
@@ -154,6 +156,8 @@ public class CertificateManager {
      * @throws KeyStoreException if unable to access KeyStore
      * @throws CsrProcessingException if unable to process CSR
      */
+    // the client certificate subscribe funtion is invoked on the thread starting component (such as bridge), i.e.
+    // nucleus lifecycle management thread
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToClientCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate[]> cb)
             throws KeyStoreException, CsrProcessingException {
--- a/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
@@ -120,6 +120,7 @@ public class CertificateManager {
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToServerCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate> cb)
             throws KeyStoreException, CsrProcessingException {
+        logger.atInfo().log("Subscribe to service certificate update");
         // BouncyCastle can throw RuntimeExceptions, and unfortunately it is not easy to detect
         // bad input beforehand. For now, just catch and re-throw a CsrProcessingException
         try {
@@ -161,6 +162,7 @@ public class CertificateManager {
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToClientCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate[]> cb)
             throws KeyStoreException, CsrProcessingException {
+        logger.atInfo().log("Subscribe to client certificate update");
         // BouncyCastle can throw RuntimeExceptions, and unfortunately it is not easy to detect
         // bad input beforehand. For now, just catch and re-throw a CsrProcessingException
         try {
--- a/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
@@ -102,7 +102,8 @@ public class CertificateManager {
     }
 
     /**
-     * Subscribe to server certificate updates.
+     * Subscribe to server certificate updates. It registered the callback function with the component for the
+     * server certificate update event, the callback could happen on another thread before this method call return.
      * <p>
      * The certificate manager will save the given CSR and generate a new certificate under the following scenarios:
      *   1) The previous certificate is nearing expiry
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -115,9 +115,12 @@ public String getCaPassPhrase() {
      * @throws KeyStoreException if unable to access KeyStore
      * @throws CsrProcessingException if unable to process CSR
      */
+    // the service certificate subscribe function is invoked on the thread starting broker, i.e. nucleus
+    // lifecycle management thread
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToServerCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate> cb)
```
sushantAtAws, line 121: Can we also add a comment on where the callback is invoked?

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -62,19 +65,23 @@
                 VerifyClientDeviceIdentityResponse response = RetryUtils.runWithRetry(SERVICE_EXCEPTION_RETRY_CONFIG,
                         () -> clientFactory.getGreengrassV2DataClient().verifyClientDeviceIdentity(request),
                         "verify-client-device-identity", logger);
-                return Optional.of(response.clientDeviceCertificateId());
+                String certificateId = response.clientDeviceCertificateId();
+                logger.atInfo().kv(CERTIFICATE_PEM, certificatePem)
```
sushantAtAws, line 69: Not sure if we need to log the client cert PEM at Info level. We should be careful about this.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -112,13 +119,13 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
                 throw new CloudServiceInteractionException(
                         "Failed to verify certificate thing association, process got interrupted", e);
             } catch (ValidationException | ResourceNotFoundException e) {
-                logger.atDebug().cause(e).kv("thingName", thing.getThingName())
-                        .kv("certificateId", certificate.getIotCertificateId())
+                logger.atInfo().cause(e).kv(THING_NAME, thing.getThingName())
```
sushantAtAws, line 122: Should this be a Warning?

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
wikimonkey: sure

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
wikimonkey: right

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
wikimonkey: client certificate should not contain sensitive info, such as private key. But probably not very useful to log certificate here either, will remove.

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
jbutler: I don't think we should make a statement about the calling context here. It's another component calling this. This likely won't hold true for bring your own broker.

We should note in the method documentation itself that this will result in a certificate being generated and the callback will be invoked prior to this method returning. Issuing a certificate is computationally expensive, so the caller needs to make sure they call in an appropriate thread context.

I'd prefer that we document callback thread context in the code where we actually invoke the callback

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -115,9 +115,13 @@ public String getCaPassPhrase() {
      * @throws KeyStoreException if unable to access KeyStore
      * @throws CsrProcessingException if unable to process CSR
      */
+    // the service certificate subscribe function is invoked on the thread starting broker, i.e. nucleus
+    // lifecycle management thread, the callback function is invoked in CertificateGenerator class function
+    // which could happen on the thread scheduled by connectivity info monitor
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToServerCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate> cb)
             throws KeyStoreException, CsrProcessingException {
+        logger.atInfo().log("Subscribe to service certificate update");
```
jbutler, line 124: I don't think this adds value. I'd prefer more information on the caller. Once we expose this via IPC we'll have that additional info. For now, I think we log this in moquette (or should, if we don't)

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -154,9 +158,13 @@ public void subscribeToServerCertificateUpdates(@NonNull String csr, @NonNull Co
      * @throws KeyStoreException if unable to access KeyStore
      * @throws CsrProcessingException if unable to process CSR
      */
+    // the client certificate subscribe function is invoked on the thread starting component (such as bridge), i.e.
+    // nucleus lifecycle management thread, the callback function is invoked in CertificateGenerator class function
+    // which could happen on the thread scheduled by certificate expiry monitor
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void subscribeToClientCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate[]> cb)
             throws KeyStoreException, CsrProcessingException {
+        logger.atInfo().log("Subscribe to client certificate update");
```
jbutler, line 167: same comments here

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -145,6 +149,7 @@ public void stopMonitor() {
      * @param certificateGenerator CertificateGenerator instance for the certificate
      */
     public void addToMonitor(CertificateGenerator certificateGenerator) {
+        LOGGER.atInfo().log("Add server certificate generator to the watcher queue");
```
jbutler, line 152: debug? I think we need additional information for this to be useful, such as caller identity

src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateExpiryMonitor.java
```diff
@@ -65,6 +68,8 @@ private void watchForCertExpiryOnce() {
 
         for (CertificateGenerator cg = monitoredCertificateGenerators.peek(); cg != null;
              cg = monitoredCertificateGenerators.peek()) {
+            LOGGER.atInfo().kv("certificateSubject", cg.subject)
+                    .log("Checking certificate expiry, regenerate if necessary");
```
jbutler, line 72: This runs every 30 seconds. Log volume will be way too high

src/main/java/com/aws/greengrass/device/DeviceAuthClient.java
```diff
@@ -165,13 +170,14 @@ public void closeSession(String sessionId) throws AuthorizationException {
...
     public void attachThing(String sessionId, String thingName) throws AuthenticationException {
-        logger.atDebug()
+        logger.atInfo()
                 .kv("sessionId", sessionId)
                 .kv("thingName", thingName)
                 .log("Attaching thing to session");
 
         // Workaround for bridge component
         if (ALLOW_ALL_SESSION.equals(sessionId)) {
+            logger.atInfo().log("Skip attach thing action for the greengrass component");
```
jbutler, line 180: Is there value in this log? We have connection / client ID logs from the broker and the corresponding session ID.

wikimonkey: yeah, broker side should log it. But want to keep CDA side log comprehensive as well.

src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateExpiryMonitor.java
wikimonkey: removed.

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
wikimonkey: Yes, need more info on the caller identity to make it useful. But I would like it to be info so we can trace the sequences in case of some rare timing issue happening.

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
wikimonkey: Yeah, caller side should log the call, but server side should log the trace as well IMO. Inside the function call, it has more context, `CSR` should be loggable in this situation to provide more info.

PR #54, opened 2022-03-29
fix: subscribe to configuration topic instead of child topics

naikvidish:
**Issue #, if available:**

**Description of changes:**
Currently we subscribe to `ca_type` and `deviceGroups` separately.
When we reset the configuration using `[""]`, we lose the subscription in the process.
Subscribing to `configuration` allows us to preserve the subscription.

When we reset the configuration for `deviceGroups` using `[""]`, we lose the subscription in the process.
Any further updates to the config are not parsed by CDA

`mvn clean verify`

Currently there are 2 tests that are failing.
One of those tests (ClientDevicesAuthServiceTest.GIVEN_GG_with_cda_WHEN_updated_ca_type_THEN_ca_is_updated) passes when run independently but fails when running all tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,6 +215,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.0.0-M5</version>
                 <configuration>
+                    <reuseForks>false</reuseForks>
                     <skipTests>true</skipTests>
                     <trimStackTrace>false</trimStackTrace>
                     <argLine>${surefire.argLine} -Xms1m</argLine>
--- a/pom.xml
+++ b/pom.xml
@@ -215,7 +215,6 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.0.0-M5</version>
                 <configuration>
-                    <reuseForks>false</reuseForks>
                     <skipTests>true</skipTests>
                     <trimStackTrace>false</trimStackTrace>
                     <argLine>${surefire.argLine} -Xms1m</argLine>
--- a/src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
@@ -7,7 +7,6 @@ package com.aws.greengrass.device;
 
 import com.aws.greengrass.certificatemanager.CertificateManager;
 import com.aws.greengrass.certificatemanager.certificate.CertificateStore;
-import com.aws.greengrass.config.Node;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.config.WhatHappened;
@@ -108,9 +107,9 @@ public class ClientDevicesAuthService extends PluginService {
             if (whatHappened == WhatHappened.initialized) {
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
                 updateCAType(whatHappened, caTypeTopic);
-            } else if (isParentTopic(node, deviceGroupTopics)) {
+            } else if (node.childOf(deviceGroupTopics.getName())) {
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
-            } else if (isParentTopic(node, caTypeTopic)) {
+            } else if (node.childOf(caTypeTopic.getName())) {
                 if (caTypeTopic.toPOJO() == null) {
                     return;
                 }
@@ -214,10 +213,5 @@ public class ClientDevicesAuthService extends PluginService {
             throw new CloudServiceInteractionException(
                     String.format("Failed to put core %s CA certificates to cloud", thingName), e);
         }
-    }
-
-    // Check if the changed node belongs to the topics.
-    private boolean isParentTopic(Node changedNode, Node topics) {
-        return changedNode != null && changedNode.getFullName().startsWith(topics.getFullName());
     }
 }
--- a/src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
@@ -7,7 +7,6 @@ package com.aws.greengrass.device;
 
 import com.aws.greengrass.certificatemanager.CertificateManager;
 import com.aws.greengrass.certificatemanager.certificate.CertificateStore;
-import com.aws.greengrass.config.Node;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.config.WhatHappened;
@@ -61,6 +60,8 @@ public class ClientDevicesAuthService extends PluginService {
 
     private final DeviceConfiguration deviceConfiguration;
 
+    private boolean hasChildChangedAfterInitialize = false;
+
     /**
      * Constructor.
      *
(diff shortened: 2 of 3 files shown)


src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -99,9 +97,13 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager, Certif
     protected void install() throws InterruptedException {
         super.install();
         //handleConfiguration
-        this.deviceGroupsTopics = this.config.lookupTopics(CONFIGURATION_CONFIG_KEY, DEVICE_GROUPS_TOPICS);
-        this.deviceGroupsTopics.subscribe(this::handleConfigurationChange);
-        this.config.lookup(CONFIGURATION_CONFIG_KEY, CA_TYPE_TOPIC).subscribe(this::updateCAType);
+        this.config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe((whatHappened, node) -> {
```
MikeDombo, line 100: ignore whathappened's that you don't care about. Specifically `interiorAdded` and `timestampUpdated`

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -99,9 +97,13 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager, Certif
...
+            handleConfigurationChange(
+                    whatHappened, this.config.lookupTopics(CONFIGURATION_CONFIG_KEY, DEVICE_GROUPS_TOPICS)
+            );
+            updateCAType(whatHappened, this.config.lookup(CONFIGURATION_CONFIG_KEY, CA_TYPE_TOPIC));
```
jcosentino11, line 104: Looks like the above two methods will be called on every config change, rather than just on device groups or ca type. To keep functionality similar to before, we'd need to check if the desired node changed before applying the update

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
naikvidish: Addressed. Thanks!

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
naikvidish: Addressed. Thanks!

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -99,9 +98,31 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager, Certif
...
+                handleConfigurationChange(whatHappened, deviceGroupTopics);
+            } else if (CA_TYPE_TOPIC.equals(node.getName())) {
+                // Skip the first duplicate event of childChanged that is received after initialize.
+                // This event is same as the one received in initialize. Skip doing duplicate work.
+                // Parse any further childChanged events.
+                if (!hasChildChangedAfterInitialize && whatHappened == WhatHappened.childChanged) {
+                    hasChildChangedAfterInitialize = true;
+                    return;
+                }
+                updateCAType(whatHappened, caTypeTopic);
```
jbutler, line 122: This doesn't feel right - is this a workaround to current Nucleus behavior or is this part of a contract? In either case, `updateCAType` should be idempotent so there shouldn't be a problem calling this twice. Unless the CA type changes it will just reload the existing CA.

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -99,9 +98,31 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager, Certif
...
         super.install();
         //handleConfiguration
-        this.deviceGroupsTopics = this.config.lookupTopics(CONFIGURATION_CONFIG_KEY, DEVICE_GROUPS_TOPICS);
-        this.deviceGroupsTopics.subscribe(this::handleConfigurationChange);
-        this.config.lookup(CONFIGURATION_CONFIG_KEY, CA_TYPE_TOPIC).subscribe(this::updateCAType);
+        this.config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe((whatHappened, node) -> {
+            if (whatHappened == WhatHappened.timestampUpdated || whatHappened == WhatHappened.interiorAdded) {
+                return;
+            }
+            logger.atTrace().kv("why", whatHappened).kv("node", node).log();
```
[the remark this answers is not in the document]
naikvidish: Done

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
naikvidish: Ahhh I over complicated it. 
It was a workaround.
It was just a matter of checking if the `caTypeTopic` is null. Updated it in the new commit.

jbutler (approved):

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -121,13 +136,13 @@ public CertificateManager getCertificateManager() {
     }
 
     @SuppressWarnings("PMD.UnusedFormalParameter")
-    private void handleConfigurationChange(WhatHappened whatHappened, Node childNode) {
+    private void handleConfigurationChange(WhatHappened whatHappened, Topics deviceGroupsTopics) {
```
jcosentino11, line 139: nit: method could use a more specific name, might want it to match w/ `updateCAType`:  `updateDeviceGroups`

pom.xml
```diff
@@ -215,6 +215,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.0.0-M5</version>
                 <configuration>
+                    <reuseForks>false</reuseForks>
```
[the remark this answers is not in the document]
MikeDombo: Setting this is definite a smell, it means that typically something `static` is in a bad state from a previous test execution.

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -99,9 +96,27 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager, Certif
...
+
+            if (whatHappened == WhatHappened.initialized) {
+                handleConfigurationChange(whatHappened, deviceGroupTopics);
+                updateCAType(whatHappened, caTypeTopic);
+            } else if (DEVICE_GROUPS_TOPICS.equals(node.getName())) {
+                handleConfigurationChange(whatHappened, deviceGroupTopics);
+            } else if (CA_TYPE_TOPIC.equals(node.getName())) {
+                if (caTypeTopic.toPOJO() == null) {
+                    return;
+                }
```
sushantAtAws, line 115: When can this case happen?
Can this happen for deviceGroupTopics too? I didnt see that check

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
naikvidish: Fixed

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -199,4 +215,9 @@ private void uploadCoreDeviceCAs(List<String> certificatePemList) {
                     String.format("Failed to put core %s CA certificates to cloud", thingName), e);
         }
     }
+
+    // Check if the changed node belongs to the topics.
+    private boolean isParentTopic(Node changedNode, Node topics) {
```
MikeDombo, line 220: you don't need this.

`childOf()` already exists

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
naikvidish: It happens after initialization when the structure of the config is being generated. A null child is added initially.

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
naikvidish: Fixed

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -121,13 +136,13 @@ public CertificateManager getCertificateManager() {
     }
 
     @SuppressWarnings("PMD.UnusedFormalParameter")
```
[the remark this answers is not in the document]
naikvidish: Removed

pom.xml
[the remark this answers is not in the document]
naikvidish: Not needed. Thanks @MikeDombo for the fix!

MikeDombo (approved):

sushantAtAws (dismissed):

jcosentino11 (approved):

MikeDombo (approved):

jbutler (approved):

PR #56, opened 2022-04-01
fix(cert-sub): deduplicate server certificate callbacks

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Deduplicates server certificate subscriptions based on the pair of csr and callback. Also implements the unsubscribe method in order to remove the subscription.

### DIFF
--- a/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
@@ -28,6 +28,8 @@ import java.security.cert.CertificateEncodingException;
 import java.security.cert.X509Certificate;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import javax.inject.Inject;
 
@@ -41,6 +43,9 @@ public class CertificateManager {
     private final CertificateExpiryMonitor certExpiryMonitor;
 
     private final CISShadowMonitor cisShadowMonitor;
+
+    private final Map<Consumer<X509Certificate>, CertificateGenerator> serverCertSubscriptions =
+            new ConcurrentHashMap<>();
 
     /**
      * Constructor.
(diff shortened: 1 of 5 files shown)


naikvidish (approved):

naikvidish (approved):

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -114,30 +120,55 @@ public String getCaPassPhrase() {
...
      * @throws CsrProcessingException if unable to process CSR
+     * @throws RuntimeException for any unknown runtime error
      */
-    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.ExceptionAsFlowControl"})
     public void subscribeToServerCertificateUpdates(@NonNull String csr, @NonNull Consumer<X509Certificate> cb)
             throws KeyStoreException, CsrProcessingException {
-        // BouncyCastle can throw RuntimeExceptions, and unfortunately it is not easy to detect
-        // bad input beforehand. For now, just catch and re-throw a CsrProcessingException
+        // deduplicate subscriptions with the same callback and csr
```
jbutler, line 128: Thoughts on ignoring the CSR and keying off just the callback? Technically the CSR gets re-generated on each request. It should be deterministic, but it still feels less right than keying off just the callback.

This API is poorly designed all around, so I think we'll want to move away from it anyway. I think I'd rather enforce 1 callback per cert and avoid potential issues with the CSR changing, especially since I can't think of a good way for a caller to actually use this API to generate multiple server certs with the same callback.

MikeDombo: Yes, that seems perfectly reasonable to me

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -114,30 +119,55 @@ public String getCaPassPhrase() {
...
-            certExpiryMonitor.addToMonitor(certificateGenerator);
-            cisShadowMonitor.addToMonitor(certificateGenerator);
-
-            certificateGenerator.generateCertificate(connectivityInfoProvider::getCachedHostAddresses);
-        } catch (KeyStoreException e) {
-            logger.atError().setCause(e).log("unable to subscribe to certificate update");
+            serverCertSubscriptions.compute(cb, (k, v) -> {
+                // A subscription already exists, we will replace it so that a new certificate is generated immediately
+                if (v != null) {
+                    unsubscribeFromServerCertificateUpdates(cb);
```
sushantAtAws, line 132: Lets not modify the map while in the compute function. 

`Attempts to compute a mapping for the specified key and its current mapped value (or null if there is no current mapping). The entire method invocation is performed atomically. Some attempted update operations on this map by other threads may be blocked while computation is in progress, so the computation should be short and simple, and must not attempt to update any other mappings of this Map.`

MikeDombo: Done.

sushantAtAws (approved):

jbutler (approved):

sushantAtAws (approved):

src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateExpiryMonitor.java
```diff
@@ -97,4 +104,20 @@ public void stopMonitor() {
             monitorFuture.cancel(true);
         }
     }
+
+    /**
+     * Remove cert from cert expiry monitor.
+     * @param cg CertificateGenerator instance for the certificate
+     */
+    public synchronized void removeFromMonitor(CertificateGenerator cg) {
```
jbutler, line 112: Note: adding/removing cert generators from a PQ is just asking for concurrency issues. Let's just add CGs to a CopyOnWriteSet and let the monitor loop check iterate through them all. There will realistically only be 1-2 CGs in it so it's probably cheaper anyway and avoids this nonsense

PR #64, opened 2022-04-21
feat: support for mqtt device credentials

jbutler:
**Issue #, if available:**

**Description of changes:**
Adds basic session management code for creating sessions based on device credentials as opposed to certificates. It introduces a `SessionCreator` which can be used to instantiate sessions based on credentials of differing types, as well as a factory for creating based on MQTT credentials.

This also removes some unused code and adds unit tests.

This change is needed in order to create sessions based on MQTT device credentials.

Unit testing. Manually tested integration with Moquette

This change is functional, but not complete. There are a couple of TODOs in the code that need to be addressed still:
1. Need to properly handle session creation for internally issued certificates
2. Need to add protection mechanism to protect against memory leaks if clients misbehave and do not close sessions
3. Add `SessionNotFound` exception for when sessions can be evicted

### DIFF
--- a/src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
@@ -39,8 +39,8 @@ public class AbstractSessionFactory {
             throws AuthenticationException {
         SessionFactory sessionFactory = SessionFactorySingleton.INSTANCE.factoryMap.get(credentialType);
         if (sessionFactory == null) {
-            logger.atWarn().kv("credentialType", credentialType).kv("credentialMap", credentialMap)
-                    .log("no registered handler to process device credentials");
+            logger.atWarn().kv("credentialType", credentialType)
+                .log("no registered handler to process device credentials");
             throw new IllegalArgumentException("unknown credential type");
         }
         return sessionFactory.createSession(credentialMap);
--- a/src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AbstractSessionFactory {
+    private static final Logger logger = LogManager.getLogger(AbstractSessionFactory.class);
+
+    @SuppressWarnings("PMD.UnusedPrivateField")
+    private final Map<String, SessionFactory> factoryMap;
+
+    private AbstractSessionFactory() {
+        factoryMap = new ConcurrentHashMap<>();
+    }
+
+    private static class SessionFactorySingleton {
+        @SuppressWarnings("PMD.AccessorClassGeneration")
+        private static final AbstractSessionFactory INSTANCE = new AbstractSessionFactory();
+    }
+
+    /**
+     * Create a client device session.
+     *
+     * @param credentialType type of credentials provided
+     * @param credentialMap  map of client credentials
+     * @return new session if the client can be authenticated
+     * @throws AuthenticationException if the client fails to be authenticated
+     */
+    public static Session createSession(String credentialType, Map<String, String> credentialMap)
+            throws AuthenticationException {
+        SessionFactory sessionFactory = SessionFactorySingleton.INSTANCE.factoryMap.get(credentialType);
+        if (sessionFactory == null) {
+            logger.atWarn().kv("credentialType", credentialType).kv("credentialMap", credentialMap)
+                    .log("no registered handler to process device credentials");
+            throw new IllegalArgumentException("unknown credential type");
+        }
+        return sessionFactory.createSession(credentialMap);
+    }
+
+    public static void registerSessionFactory(String credentialType, SessionFactory sessionFactory) {
+        SessionFactorySingleton.INSTANCE.factoryMap.put(credentialType, sessionFactory);
+    }
+
+    public static void unregisterSessionFactory(String credentialType) {
+        SessionFactorySingleton.INSTANCE.factoryMap.remove(credentialType);
+    }
+}
--- a/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
@@ -11,10 +11,12 @@ import com.aws.greengrass.device.iot.IotAuthClient;
 import com.aws.greengrass.device.iot.Thing;
 
 import java.util.Map;
+import javax.inject.Inject;
 
 public class MqttSessionFactory implements SessionFactory {
     private final IotAuthClient iotAuthClient;
 
+    @Inject
     public MqttSessionFactory(IotAuthClient iotAuthClient) {
         this.iotAuthClient = iotAuthClient;
     }
--- a/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+import com.aws.greengrass.device.iot.Certificate;
+import com.aws.greengrass.device.iot.IotAuthClient;
+import com.aws.greengrass.device.iot.Thing;
+
+import java.util.Map;
+
+public class MqttSessionFactory implements SessionFactory {
+    private final IotAuthClient iotAuthClient;
+
+    public MqttSessionFactory(IotAuthClient iotAuthClient) {
+        this.iotAuthClient = iotAuthClient;
+    }
+
+    @Override
+    public Session createSession(Map<String, String> credentialMap) throws AuthenticationException {
+        // TODO: replace with jackson object mapper
+        MqttCredential mqttCredential = new MqttCredential(credentialMap);
+
+        // TODO: support internal client certificates.
+        //   Internally issued certificates need to be handled by the entity creating sessions since
+        //   we don't yet have the ability to authorize actions for those clients. So, for now, assume
+        //   this is an IoT Thing and components will be specially handled elsewhere.
+        Thing thing = new Thing(mqttCredential.clientId);
+        Certificate cert = new Certificate(mqttCredential.certificatePem);
+        if (!iotAuthClient.isThingAttachedToCertificate(thing, cert)) {
+            throw new AuthenticationException("unable to authenticate device");
+        }
+
+        Session session = new SessionImpl(cert);
+        session.putAttributeProvider(Thing.NAMESPACE, thing);
+        return session;
+    }
+
+    private static class MqttCredential {
+        private final String clientId;
+        private final String certificatePem;
+        private final String username;
+        private final String password;
+
+        public MqttCredential(Map<String, String> credentialMap) {
+            this.clientId = credentialMap.get("clientId");
+            this.certificatePem = credentialMap.get("certificatePem");
+            this.username = credentialMap.get("username");
+            this.password = credentialMap.get("password");
+        }
+    }
+}
--- a/src/main/java/com/aws/greengrass/device/session/SessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/SessionFactory.java
@@ -1,0 +1,14 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+
+import java.util.Map;
+
+public interface SessionFactory {
+    Session createSession(Map<String, String> credentialMap) throws AuthenticationException;
+}
--- a/src/main/java/com/aws/greengrass/device/session/SessionManager.java
+++ b/src/main/java/com/aws/greengrass/device/session/SessionManager.java
@@ -83,13 +83,13 @@ public class SessionManager {
 
     // Returns a session ID which can be returned to the client
     private synchronized String addSessionInternal(Session session) {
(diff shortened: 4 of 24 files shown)


src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
```diff
@@ -0,0 +1,55 @@
...
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+import com.aws.greengrass.device.iot.Certificate;
+import com.aws.greengrass.device.iot.IotAuthClient;
+import com.aws.greengrass.device.iot.Thing;
+
+import java.util.Map;
+
+public class MqttSessionFactory implements SessionFactory {
```
jbutler, line 15: TODO: this factory needs to be registered

src/main/java/com/aws/greengrass/device/session/SessionFactory.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+
+import java.util.Map;
+
+public interface SessionFactory {
+    Session createSession(Map<String, String> credentialMap) throws AuthenticationException;
```
MikeDombo, line 13: you want to enforce string->string? I think you may want to be a bit more generic, String->Object, or even make it generic String->T

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -0,0 +1,118 @@
...
+    private final Map<String, String> externalToInternalSessionMap = new ConcurrentHashMap<>();
+
+    /**
+     * find session by id.
+     *
+     * @param sessionId session identifier
+     * @return session or null
+     */
+    public Session findSession(String sessionId) {
+        String internalSessionId = externalToInternalSessionMap.get(sessionId);
```
MikeDombo, line 45: why do we need indirection of internal and external session IDs?

src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
```diff
@@ -0,0 +1,56 @@
...
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AbstractSessionFactory {
```
MikeDombo, line 15: not actually abstract?

src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param credentialType type of credentials provided
+     * @param credentialMap  map of client credentials
+     * @return new session if the client can be authenticated
+     * @throws AuthenticationException if the client fails to be authenticated
+     */
+    public static Session createSession(String credentialType, Map<String, String> credentialMap)
+            throws AuthenticationException {
+        SessionFactory sessionFactory = SessionFactorySingleton.INSTANCE.factoryMap.get(credentialType);
+        if (sessionFactory == null) {
+            logger.atWarn().kv("credentialType", credentialType).kv("credentialMap", credentialMap)
```
MikeDombo, line 42: you don't want to log the actual credentials. You do the lookup based on the credential type, so that should give enough info in the log

src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private static class MqttCredential {
+        private final String clientId;
+        private final String certificatePem;
+        private final String username;
+        private final String password;
+
+        public MqttCredential(Map<String, String> credentialMap) {
+            this.clientId = credentialMap.get("clientId");
+            this.certificatePem = credentialMap.get("certificatePem");
```
MikeDombo, line 50: do we need to hold onto the whole pem, or can we hash or otherwise reduce our need for it?

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -0,0 +1,118 @@
...
+        if (internalSessionId == null) {
+            return null;
+        }
+        return sessionMap.getOrDefault(internalSessionId, null).getRight();
```
MikeDombo, line 49: get or default is useless here since if it doesn't exist it will return null already. And you'll then cause a NPE.

src/test/java/com/aws/greengrass/device/session/AbstractSessionFactoryTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.HashMap;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@ExtendWith({MockitoExtension.class})
```
MikeDombo, line 24: ggextension

src/test/java/com/aws/greengrass/device/session/MqttSessionFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Map;
+
+import org.hamcrest.core.IsNull;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith({MockitoExtension.class})
```
MikeDombo, line 26: ggextension

src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
MikeDombo: More of a SPI right?

jbutler: Sure, can rename

src/main/java/com/aws/greengrass/device/session/AbstractSessionFactory.java
jbutler: Good call - will remove

src/main/java/com/aws/greengrass/device/session/SessionFactory.java
jbutler: Didn't want to overthink this one too much to start. It should be easy enough to change later

src/main/java/com/aws/greengrass/device/session/SessionManager.java
jbutler: Yep, good catch

src/test/java/com/aws/greengrass/device/session/AbstractSessionFactoryTest.java
jbutler: What does that do again? This class doesn't really have anything to do with gg. I'm not sure why we're using it everywhere

src/main/java/com/aws/greengrass/device/session/SessionFactory.java
MikeDombo: It is our common base. It includes several extensions including a log exception protector and extra thread protector. Just good to have as best practice everywhere, even in unit tests which don't use GG stuff.

src/main/java/com/aws/greengrass/device/session/SessionManager.java
jbutler: I could be convinced that it's not needed. But I'm hesitant to return a "predictable" session ID to clients - similar to guidance given for pagination tokens. Worth a quick discussion with the security team, I suppose

src/test/java/com/aws/greengrass/device/session/AbstractSessionFactoryTest.java
MikeDombo: It is our common base. It includes several extensions including a log exception protector and extra thread protector. Just good to have as best practice everywhere, even in unit tests which don't use GG stuff.


(I had put this reply in the wrong place)

src/main/java/com/aws/greengrass/device/session/SessionManager.java
MikeDombo: I'm ok with it, but if removing the indirection will simplify anything then I'd push for it. Pagination token needs to be kept more obscure AFAIK because it can directly include the actual value of the last scanned key, so you're leaking info. Just having a session ID which is a hash of the inputs should be opaque enough.

src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
MikeDombo: How are the tests passing if this isn't registered? Does that imply that the integration tests have insufficient coverage?

jbutler: Only have unit tests at this point. Will add an integ test to cover this

src/main/java/com/aws/greengrass/device/session/SessionManager.java
jbutler: You're right, the reasons for opaque pagination tokens don't quite apply here. I think the complexity can be contained within this class, so I'm not too too concerned with it. But we should definitely think harder if it starts leaking out.

Other reasons for returning unique IDs:
1. Improved debuggability/tracability - I can see not being able to differentiate between two sessions created with the same credentials being annoying
2. Less chance of a client side bug (using wrong auth token to authorize an action)

We should at least make sure the format is the same, however. I'll make a change to fix that so we have the option to easily change this behavior in the future without risk of breaking anything.

src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
MikeDombo: No doubt it'll be fine later, but I'm surprised that if this PR completely replaces the session management, that the existing integ tests (which should be exercising sessions) are still passing even when the MQTT factory isn't used. The existing integ tests are mocked or stubbed too much perhaps.

src/main/java/com/aws/greengrass/device/session/SessionManager.java
MikeDombo: I'm fine with it being uuids, no concern there.

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -0,0 +1,122 @@
...
+        Session session = sessionMap.get(internalSessionId).getRight();
+        sessionMap.remove(internalSessionId, session);
+    }
+
+    /*
+     * Returns an external session ID, if one was present and was evicted. Else, null
+     */
+    private synchronized String addSessionInternal(String externalId, String internalId, Session session) {
+        Pair<String, Session> existingPair = sessionMap.get(internalId);
+        sessionMap.put(internalId, new Pair<>(externalId, session));
```
MikeDombo, line 114: put returns the old value, so you can just get the existing value from here.

src/test/java/com/aws/greengrass/device/session/AbstractSessionFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+    private MqttSessionFactory mqttSessionFactory;
+
+    @AfterEach
+    void afterEach() {
+        AbstractSessionFactory.unregisterSessionFactory(mqttCredentialType);
+    }
+
+    @Test
+    void GIVEN_noRegisteredFactories_WHEN_createSession_THEN_throwsException() {
+        Assertions.assertThrows(IllegalArgumentException.class,
```
MikeDombo, line 40: Illegal argument is unchecked, so callers will need to be careful to handle it fully. Our coding guidelines also say to prefer checked exceptions where possible, so consider that as well.

src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
jbutler: We need the full pem to verify that it's attached to a Thing. But we're not actually holding a reference to this object anywhere, so should be okay

MikeDombo: > We're not actually holding a reference

You mean that this MqttCredential object is just used temporarily during authN?

src/main/java/com/aws/greengrass/device/session/SessionManager.java
jbutler: I removed this logic for now. The code path that Moquette uses currently creates sessions based on certificates, and we can't de-dup based on this alone (since multiple clients could use the same cert)

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+public class SessionManager {
+    private static final Logger logger = LogManager.getLogger(SessionManager.class);
+    private static final String SESSION_ID = "SessionId";
+
+    // TODO: Add mechanism to prevent session leaks
+    // This implementation currently relies on clients to properly close sessions. If they don't, then sessions
+    // will be leaked. We can add this once sessions are created with the entire set of device credentials. However,
+    // Moquette is currently creating sessions using just the certificate. Since multiple clients could use the same
+    // certificate, we can't de-dup based on this alone.
```
MikeDombo, line 30: what will we dedupe on in the future? Client id + cert?

If the cert is the identity, then is there actually a problem in using the same session since the authZ decisions would be the same?

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -0,0 +1,102 @@
...
+        logger.atDebug().kv(SESSION_ID, sessionId).log("Closing session");
+        closeSessionInternal(sessionId);
+    }
+
+    private synchronized void closeSessionInternal(String sessionId) {
+        sessionMap.remove(sessionId);
+    }
+
+    /*
+     * Returns external session ID
```
MikeDombo, line 85: accurate javadoc?

src/main/java/com/aws/greengrass/device/session/SessionManager.java
jbutler: The cert won't be the identity. Moquette needs to be changed to use mqtt device credentials.

The current code "attaches" the thing after the fact for AuthZ decisions. You can just trust me on this one ;)

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/device/session/SessionManager.java
jbutler: Do ppl generally write proper javadocs for private methods?

MikeDombo: No, but if there is a comment it needs to be accurate. I was just asking if this comment is actually correct.

MikeDombo (approved):

sushantAtAws (approved):

PR #65, opened 2022-04-25
feat: expose verifying client device identity over ipc

saranyailla:
**Issue #, if available:**

**Description of changes:**
Exposes verifying client device identity over IPC. 

Facilitate any authorized non-plugin component (such as new MQTT 5 broker or BYOB) to verify client device identity over IPC.

Added integ tests. 

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
@@ -1,0 +1,260 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.integrationtests.ipc;
+
+import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.device.ClientDevicesAuthService;
+import com.aws.greengrass.device.exception.CloudServiceInteractionException;
+import com.aws.greengrass.device.iot.IotAuthClient;
+import com.aws.greengrass.lifecyclemanager.GlobalStateChangeListener;
+import com.aws.greengrass.lifecyclemanager.GreengrassService;
+import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.logging.impl.config.LogConfig;
+import com.aws.greengrass.testcommons.testutilities.GGExtension;
+import com.aws.greengrass.testcommons.testutilities.UniqueRootPathExtension;
+import com.aws.greengrass.util.GreengrassServiceClientFactory;
+import com.aws.greengrass.util.Pair;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.io.TempDir;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCClient;
+import software.amazon.awssdk.aws.greengrass.VerifyClientDeviceIdentityResponseHandler;
+import software.amazon.awssdk.aws.greengrass.model.ClientDeviceCredential;
+import software.amazon.awssdk.aws.greengrass.model.ServiceError;
+import software.amazon.awssdk.aws.greengrass.model.UnauthorizedError;
+import software.amazon.awssdk.aws.greengrass.model.VerifyClientDeviceIdentityRequest;
+import software.amazon.awssdk.aws.greengrass.model.VerifyClientDeviceIdentityResponse;
+import software.amazon.awssdk.eventstreamrpc.EventStreamRPCConnection;
+import software.amazon.awssdk.services.greengrassv2data.GreengrassV2DataClient;
+import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
+import software.amazon.awssdk.services.greengrassv2data.model.ValidationException;
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
+import static com.aws.greengrass.testcommons.testutilities.TestUtils.asyncAssertOnConsumer;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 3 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
```diff
@@ -0,0 +1,181 @@
...
+            VerifyClientDeviceIdentityRequest request = new VerifyClientDeviceIdentityRequest()
+                    .withCredential(new ClientDeviceCredential()
+                            .withClientDeviceCertificate("VALID PEM"));
+            Pair<CompletableFuture<Void>, Consumer<VerifyClientDeviceIdentityResponse>> cb =
+                    asyncAssertOnConsumer((m) -> {
+                    });
+            Exception err = Assertions.assertThrows(Exception.class, () -> {
+                verifyClientIdentity(ipcClient, request, cb.getRight());
+            });
+            assertThat(err.getCause(), is(instanceOf(UnauthorizedError.class)));
```
MikeDombo, line 178: you have to use `cb.getLeft().get` to ensure it was called and only called the expected number of times. If you aren't doing this, then there really isn't any point in using `asyncAssertOnConsumer`

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
```diff
@@ -0,0 +1,105 @@
...
+        super(context);
+        this.iotAuthClient = iotAuthClient;
+        serviceName = context.getAuthenticationData().getIdentityLabel();
+        this.authorizationHandler = authorizationHandler;
+    }
+
+    @SuppressWarnings("PMD.PreserveStackTrace")
+    @Override
+    public VerifyClientDeviceIdentityResponse handleRequest(VerifyClientDeviceIdentityRequest request) {
+        return translateExceptions(() -> {
```
MikeDombo, line 61: formatting

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
```diff
@@ -0,0 +1,105 @@
...
+            try {
+                doAuthorizationForClientDevIdentity();
+            } catch (AuthorizationException e) {
+            logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+            throw new UnauthorizedError(e.getMessage());
+        }
+            String certificate = getCertificateFromCredential(request.getCredential());
+            Optional<String> certificateId = iotAuthClient.getActiveCertificateId(certificate);
```
MikeDombo, line 69: catch and handle errors from iot auth client

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static com.aws.greengrass.ipc.common.ExceptionUtil.translateExceptions;
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.VERIFY_CLIENT_DEVICE_IDENTITY;
+
+public class VerifyClientDeviceIdentityOperationHandler
+        extends GeneratedAbstractVerifyClientDeviceIdentityOperationHandler {
+    private static final Logger logger = LogManager.getLogger(VerifyClientDeviceIdentityOperationHandler.class);
+    private static final String COMPONENT_NAME = "componentName";
+    private static final String UNAUTHORIZED_ERROR = "Not Authorized";
+    private static final String NO_DEVICE_CREDENTIAL_ERROR = "Client device credential is required.";
```
MikeDombo, line 35: remove period

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
```diff
@@ -0,0 +1,181 @@
...
+        kernel.getContext().put(GreengrassServiceClientFactory.class, clientFactory);
+
+
+        when(clientFactory.getGreengrassV2DataClient()).thenReturn(client);
+        lenient().when(iotAuthClient.getActiveCertificateId("VALID PEM")).thenReturn(
+                Optional.of("SOME-CERT")
+        );
+        lenient().when(iotAuthClient.getActiveCertificateId("INVALID PEM")).thenReturn(
+                Optional.empty()
+        );
```
MikeDombo, line 87: setup these mocks in the individual test case so that you aren't using `lenient`

Have a test case which includes the auth client throwing an exception

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
```diff
@@ -0,0 +1,254 @@
...
+        ignoreExceptionOfType(context, CloudServiceInteractionException.class);
+        try (EventStreamRPCConnection connection = IPCTestUtils.getEventStreamRpcConnection(kernel,
+                "BrokerSubscribingToCertUpdates")) {
+            GreengrassCoreIPCClient ipcClient = new GreengrassCoreIPCClient(connection);
+            VerifyClientDeviceIdentityRequest request = new VerifyClientDeviceIdentityRequest()
+                    .withCredential(new ClientDeviceCredential()
+                            .withClientDeviceCertificate("Generic Exception PEM"));
+
+            Exception err = Assertions.assertThrows(Exception.class, () -> {
+                verifyClientIdentity(ipcClient, request,null);
```
MikeDombo, line 249: `,` formatting.

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    public VerifyClientDeviceIdentityResponse handleRequest(VerifyClientDeviceIdentityRequest request) {
+        return translateExceptions(() -> {
+            try {
+                doAuthorizationForClientDevIdentity();
+            } catch (AuthorizationException e) {
+                logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+                throw new UnauthorizedError(e.getMessage());
+            }
+            String certificate = getCertificateFromCredential(request.getCredential());
+            VerifyClientDeviceIdentityResponse response = new VerifyClientDeviceIdentityResponse();
```
MikeDombo, line 71: [nit]

Move this response creation and return inside the `try`. Seems a bit more clear that it will return or throw. ie. the `catch` won't ever return a response except for an exception.

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
```diff
@@ -0,0 +1,113 @@
...
+            try {
+                Optional<String> certificateId = iotAuthClient.getActiveCertificateId(certificate);
```
MikeDombo, line 73: @jbutler 

This call goes to the cloud. Performing a synchronous cloud lookup will block the SDK thread. This is scary to me, even given limited retries and timeouts.

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
MikeDombo: Move the response inside the try. Variables should have the minimum possible scope

src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java
MikeDombo: Discussed offline with Joe and we will be changing this, but not for this commit right now.

MikeDombo (approved):

PR #66, opened 2022-04-26
refactor: modify credentials from map to object

saranyailla:
**Issue #, if available:**

**Description of changes:**
Use Credential and MqttCredential object instead of Map. 

### DIFF
--- a/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
@@ -9,8 +9,9 @@ import com.aws.greengrass.device.exception.AuthenticationException;
 import com.aws.greengrass.device.iot.Certificate;
 import com.aws.greengrass.device.iot.IotAuthClient;
 import com.aws.greengrass.device.iot.Thing;
+import com.aws.greengrass.device.session.credentials.Credential;
+import com.aws.greengrass.device.session.credentials.MqttCredential;
 
-import java.util.Map;
 import javax.inject.Inject;
 
 public class MqttSessionFactory implements SessionFactory {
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
```diff
@@ -22,16 +23,15 @@ public MqttSessionFactory(IotAuthClient iotAuthClient) {
     }
 
     @Override
-    public Session createSession(Map<String, String> credentialMap) throws AuthenticationException {
-        // TODO: replace with jackson object mapper
-        MqttCredential mqttCredential = new MqttCredential(credentialMap);
+    public Session createSession(Credential credential) throws AuthenticationException {
+        MqttCredential mqttCredential = (MqttCredential) credential;
```
MikeDombo, line 27: do not blindly cast, make sure it is the right type first

MikeDombo (commented):
What is the motivation for this change?

MikeDombo:
I do not think that we want this because it seems to force the credential object creation into IPC, rather than abstracting it and letting the session implementation handle it on its own. For example, if we wanted to support some new auth scheme then the current interface allows for scheme (ex. mqtt) registration and then the new authenticator can parse the input in a way that it understands.

src/main/java/com/aws/greengrass/device/session/credentials/Credential.java
```diff
@@ -0,0 +1,6 @@
+package com.aws.greengrass.device.session.credentials;
```
jbutler, line 1: copyright

src/main/java/com/aws/greengrass/device/session/credentials/MqttCredential.java
```diff
@@ -0,0 +1,17 @@
+package com.aws.greengrass.device.session.credentials;
```
jbutler, line 1: copyright

src/main/java/com/aws/greengrass/device/session/SessionCreator.java
```diff
@@ -22,28 +23,23 @@ private SessionCreator() {
...
             logger.atWarn().kv("credentialType", credentialType)
-                .log("no registered handler to process device credentials");
+                    .log("no registered handler to process device credentials");
             throw new IllegalArgumentException("unknown credential type");
         }
-        return sessionFactory.createSession(credentialMap);
+        return sessionFactory.createSession(credential);
     }
 
     public static void registerSessionFactory(String credentialType, SessionFactory sessionFactory) {
```
jbutler, line 45: Credential registration will break with this change. Please update PR description once tested

PR #69, opened 2022-04-26
feat: expose get client device auth token over ipc

saranyailla:
**Issue #, if available:**

**Description of changes:**
Expose get client device auth token over IPC. 


src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,129 @@
...
+        return translateExceptions(() -> {
+            try {
+                doAuthorizationForClientDevAuthToken();
+            } catch (AuthorizationException e) {
+                logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+                throw new UnauthorizedError(e.getMessage());
+            }
+            Map<String, String> credentialMap = mapOfMqttCredential(request.getCredential());
+            try {
+                String sessionId = sessionManager.createSession(MQTT_CREDENTIAL_TYPE, credentialMap);
```
MikeDombo, line 78: catch all exceptions from createSession and throw a nice-ish error that we couldn't create a session

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,128 @@
...
+                GetClientDeviceAuthTokenResponse response = new GetClientDeviceAuthTokenResponse();
+                return response.withClientDeviceAuthToken(sessionId);
+            } catch (Exception e) {
```
MikeDombo, line 80: if it throws a more specific error and we can give the user a better indication of what went wrong, then we should do that. I just want to be complete and make sure you additionally have a catch all.

MikeDombo: AuthenticationException should be treated differently - we return false to the client, rather than throwing the client an exception right

MikeDombo: ok

MikeDombo: But it isn't a ServiceError, it should be a properly modeled exception in that case.

MikeDombo: `InvalidCredentialError` is what is in the design doc

jbutler: +1 The session manager throws an `AuthenticationException` which should be caught and translated here

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static com.aws.greengrass.ipc.common.ExceptionUtil.translateExceptions;
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.GET_CLIENT_DEVICE_AUTH_TOKEN;
+
+public class GetClientDeviceAuthTokenOperationHandler
+        extends GeneratedAbstractGetClientDeviceAuthTokenOperationHandler {
+    private static final Logger logger = LogManager.getLogger(GetClientDeviceAuthTokenOperationHandler.class);
+    private static final String COMPONENT_NAME = "componentName";
+    private static final String UNAUTHORIZED_ERROR = "Not Authorized";
+    private static final String MQTT_CREDENTIAL_TYPE = "mqtt";
```
jbutler, line 38: Fine for now, but ideally this layer wouldn't need to be aware of credential types. Will probably need some tweaks to the SessionManager

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @SuppressWarnings("PMD.PreserveStackTrace")
+    private MQTTCredential validateMqttCredentials(CredentialDocument credentialDocument) {
+        if (credentialDocument == null) {
+            throw new InvalidArgumentsError(NO_DEVICE_CREDENTIAL_ERROR);
+        }
+        MQTTCredential mqttCredential = credentialDocument.getMqttCredential();
+        if (mqttCredential == null) {
+            throw new InvalidArgumentsError(NO_VALID_MQTT_CREDENTIAL_ERROR);
```
jbutler, line 95: I'd keep this basic. "Invalid device credentials"

saranyailla: L91 gives `Client device credential document is required` and L95 gives `Valid mqtt credential document is required`. Are you suggesting `Invalid mqtt credentials`  for L95?

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+            Map<String, String> credentialMap = mapOfMqttCredential(request.getCredential());
+            try {
+                String sessionId = sessionManager.createSession(MQTT_CREDENTIAL_TYPE, credentialMap);
+                GetClientDeviceAuthTokenResponse response = new GetClientDeviceAuthTokenResponse();
+                return response.withClientDeviceAuthToken(sessionId);
+            } catch (AuthenticationException e) {
+                logger.atError().cause(e).log("Unable to authenticate the client device.");
+                throw new ServiceError(
+                        "Getting client device auth token failed. Check Greengrass log for details.");
```
MikeDombo, line 84: this isn't a serviceerror, should be a modeled `InvalidCredentialError`

saranyailla: Still working on that :)

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                throw new UnauthorizedError(e.getMessage());
+            }
+            Map<String, String> credentialMap = mapOfMqttCredential(request.getCredential());
+            try {
+                String sessionId = sessionManager.createSession(MQTT_CREDENTIAL_TYPE, credentialMap);
+                GetClientDeviceAuthTokenResponse response = new GetClientDeviceAuthTokenResponse();
+                return response.withClientDeviceAuthToken(sessionId);
+            } catch (AuthenticationException e) {
+                logger.atError().cause(e).log("Unable to authenticate the client device with the given credentials.");
+                throw new InvalidCredentialError(
```
MikeDombo, line 84: in this case, I think passing through the error message would be appropriate. Rather than just directing them to the logs.

MikeDombo (approved):

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
saranyailla: Added tests as well.

MikeDombo (approved):

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
jbutler: I'm saying not to return detailed information to the caller about what failed. You can log extra information here, but just return `Invalid device credentials` to the caller.

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            Map<String, String> credentialMap = mapOfMqttCredential(request.getCredential());
+            try {
+                String sessionId = sessionManager.createSession(MQTT_CREDENTIAL_TYPE, credentialMap);
+                GetClientDeviceAuthTokenResponse response = new GetClientDeviceAuthTokenResponse();
+                return response.withClientDeviceAuthToken(sessionId);
+            } catch (AuthenticationException e) {
+                logger.atError().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName)
+                        .log("Unable to authenticate the client device with the given credentials");
+                throw new InvalidCredentialError(e.getMessage());
```
jbutler, line 85: Don't include the exception message. We don't want to return that to the client.

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+            Map<String, String> credentialMap = mapOfMqttCredential(request.getCredential());
+            try {
+                String sessionId = sessionManager.createSession(MQTT_CREDENTIAL_TYPE, credentialMap);
+                GetClientDeviceAuthTokenResponse response = new GetClientDeviceAuthTokenResponse();
+                return response.withClientDeviceAuthToken(sessionId);
+            } catch (AuthenticationException e) {
+                logger.atError().cause(e).log("Unable to get client device auth token from the session");
+                throw new InvalidCredentialError(
+                        "Getting client device auth token failed. Check Greengrass log for details.");
```
MikeDombo, line 84: I think something like "Unable to authenticate client device with given credentials" is more appropriate here.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.PreserveStackTrace"})
+    @Override
+    public GetClientDeviceAuthTokenResponse handleRequest(GetClientDeviceAuthTokenRequest request) {
+        return translateExceptions(() -> {
+            try {
+                doAuthorizationForClientDevAuthToken();
+            } catch (AuthorizationException e) {
+                logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+                throw new UnauthorizedError(e.getMessage());
```
jbutler, line 74: Same thing here. Log the cause but return a boiler plate message

MikeDombo: The message is already boilerplate defined in the nucleus. All our ipc services do this the same way. It should not be changed.

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/GetClientDeviceAuthTokenTest.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @AfterEach
+    void afterEach() {
+        LogConfig.getRootLogConfig().reset();
+        kernel.shutdown();
+    }
+
+    @Test
+    void GIVEN_broker_WHEN_get_client_device_auth_token_THEN_get() throws Exception {
```
jbutler, line 119: Rename all these test cases to be a little more clear. It's assumed that a "broker" is the thing calling these APIs, so dig in one level further. We're talking about device credentials, or component IPC policies. Make it clear from the test name what is being tested.

In this case, it's 
`GIVEN_validMqttCredentialMap_WHEN_GetClientDeviceAuthToken_THEN_returnsValidToken`

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/GetClientDeviceAuthTokenTest.java
```diff
@@ -0,0 +1,247 @@
...
+                    asyncAssertOnConsumer((m) -> {
+                        assertEquals("some-session-uuid", m.getClientDeviceAuthToken());
+                    });
+            clientDeviceAuthToken(ipcClient, request, cb.getRight());
+            cb.getLeft().get(10, TimeUnit.SECONDS);
+        }
+    }
+
+    @Test
+    void GIVEN_broker_WHEN_get_client_device_auth_token_with_invalid_input_THEN_throw_error()
```
jbutler, line 149: `GIVEN_invalidMqttCredentialMap_WHEN_GetClientDeviceAuthToken_THEN_throwsInvalidArgumentsError`

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/GetClientDeviceAuthTokenTest.java
```diff
@@ -0,0 +1,247 @@
...
+                clientDeviceAuthToken(ipcClient, request, null);
+            });
+            assertEquals(err.getCause().getClass(), InvalidArgumentsError.class);
+            assertThat(err.getCause().getMessage(), containsString("Invalid client device credentials"));
+
+        }
+    }
+
+    @Test
+    void GIVEN_broker_with_no_config_WHEN_get_client_device_auth_token_THEN_throw_error()
```
jbutler, line 169: `GIVEN_missingIpcPermission_WHEN_GetClientDeviceAuthToken_THEN_throwsUnauthorizedError` ?

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
jbutler: Ah, right, this is core IPC stuff. As long as that error message is controlled in Nucleus (which it looks like it is) then that's ok

src/integrationtests/resources/com/aws/greengrass/integrationtests/ipc/cda.yaml
```diff
@@ -86,3 +87,22 @@ services:
...
+          sleep 1
+    configuration:
+      accessControl:
+        aws.greengrass.clientdevices.Auth:
+          GetClientDeviceAuthTokenPolicy:
+            policyDescription: access to certificate updates
+            operations:
+              - 'aws.greengrass#GetClientDeviceAuthToken'
+            resources:
+              - '*'
```
MikeDombo, line 108: EOF.

Configure your IDE to always end with a newline

MikeDombo (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/GetClientDeviceAuthTokenTest.java
saranyailla: Updated test case names. Still kept broker in the name as these are integration tests.

jbutler (approved):

PR #70, opened 2022-04-28
feat: expose authz client device action over IPC

saranyailla:
**Issue #, if available:**

**Description of changes:**
This change exposes authorizing client device action over IPC. Returns either True or False if a valid session exists and is able to authorize certain operation on a resource. 
If a valid session is not found, then `InvalidClientDeviceAuthTokenError` is thrown. 
For all other exceptions, `ServiceError` is thrown.

Needed for mqtt5/byob  integration. 

Added integrations tests. 

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
@@ -8,7 +8,7 @@ package com.aws.greengrass.integrationtests.ipc;
 import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.device.ClientDevicesAuthService;
 import com.aws.greengrass.device.DeviceAuthClient;
-import com.aws.greengrass.device.exception.AuthorizationException;
+import com.aws.greengrass.device.exception.InvalidSessionException;
 import com.aws.greengrass.lifecyclemanager.GlobalStateChangeListener;
 import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.lifecyclemanager.Kernel;
@@ -68,8 +68,8 @@ class AuthorizeClientDeviceActionTest {
     private DeviceAuthClient deviceAuthClient;
 
     private static void authzClientDeviceAction(GreengrassCoreIPCClient ipcClient,
-                                               AuthorizeClientDeviceActionRequest request,
-                                               Consumer<AuthorizeClientDeviceActionResponse> consumer)
+                                                AuthorizeClientDeviceActionRequest request,
+                                                Consumer<AuthorizeClientDeviceActionResponse> consumer)
             throws Exception {
         AuthorizeClientDeviceActionResponseHandler handler =
                 ipcClient.authorizeClientDeviceAction(request, Optional.empty());
@@ -179,7 +179,7 @@ class AuthorizeClientDeviceActionTest {
     void GIVEN_brokerWithInvalidAuthToken_WHEN_AuthorizeClientDeviceAction_THEN_throwInvalidClientDeviceAuthTokenError
             (ExtensionContext context) throws Exception {
         kernel.getContext().put(DeviceAuthClient.class, deviceAuthClient);
-        ignoreExceptionOfType(context, AuthorizationException.class);
+        ignoreExceptionOfType(context, InvalidSessionException.class);
         startNucleusWithConfig("cda.yaml");
         try (EventStreamRPCConnection connection = IPCTestUtils.getEventStreamRpcConnection(kernel,
                 "BrokerWithAuthorizeClientDeviceActionPermission")) {
@@ -188,7 +188,7 @@ class AuthorizeClientDeviceActionTest {
                     .withOperation("some-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-invalid-token");
-            when(deviceAuthClient.canDevicePerform(any())).thenThrow(AuthorizationException.class);
+            when(deviceAuthClient.canDevicePerform(any())).thenThrow(InvalidSessionException.class);
             Exception err = assertThrows(Exception.class, () -> {
                 authzClientDeviceAction(ipcClient, request, null);
             });
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
@@ -215,12 +215,7 @@ class AuthorizeClientDeviceActionTest {
                     .withOperation("some-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-token");
-            AuthorizationRequest authzReq = AuthorizationRequest.builder()
-                    .sessionId(request.getClientDeviceAuthToken())
-                    .operation(request.getOperation())
-                    .resource(request.getResource())
-                    .build();
-            when(deviceAuthClient.canDevicePerform(authzReq)).thenThrow(IllegalArgumentException.class);
+            when(deviceAuthClient.canDevicePerform(any())).thenThrow(IllegalArgumentException.class);
             Exception err = assertThrows(Exception.class, () -> {
                 authzClientDeviceAction(ipcClient, request, null);
             });
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
@@ -6,6 +6,7 @@
 package com.aws.greengrass.integrationtests.ipc;
 
 import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.device.AuthorizationRequest;
 import com.aws.greengrass.device.ClientDevicesAuthService;
 import com.aws.greengrass.device.DeviceAuthClient;
 import com.aws.greengrass.device.exception.InvalidSessionException;
@@ -51,7 +52,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.when;
 
 @ExtendWith({GGExtension.class, UniqueRootPathExtension.class, MockitoExtension.class})
@@ -63,7 +63,7 @@ class AuthorizeClientDeviceActionTest {
     @Mock
     private GreengrassServiceClientFactory clientFactory;
     @Mock
-    private GreengrassV2DataClient client;
+    private GreengrassV2DataClient v2DataClient;
     @Mock
     private DeviceAuthClient deviceAuthClient;
 
@@ -73,7 +73,7 @@ class AuthorizeClientDeviceActionTest {
             throws Exception {
         AuthorizeClientDeviceActionResponseHandler handler =
                 ipcClient.authorizeClientDeviceAction(request, Optional.empty());
-        AuthorizeClientDeviceActionResponse response = handler.getResponse().get(10, TimeUnit.SECONDS);
+        AuthorizeClientDeviceActionResponse response = handler.getResponse().get(5, TimeUnit.SECONDS);
         consumer.accept(response);
     }
 
@@ -84,7 +84,7 @@ class AuthorizeClientDeviceActionTest {
         kernel = new Kernel();
         kernel.getContext().put(GreengrassServiceClientFactory.class, clientFactory);
 
-        when(clientFactory.getGreengrassV2DataClient()).thenReturn(client);
+        when(clientFactory.getGreengrassV2DataClient()).thenReturn(v2DataClient);
 
     }
 
@@ -125,13 +125,18 @@ class AuthorizeClientDeviceActionTest {
                     .withOperation("some-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-token");
-            when(deviceAuthClient.canDevicePerform(any())).thenReturn(true);
+            AuthorizationRequest authzReq = AuthorizationRequest.builder()
+                    .sessionId(request.getClientDeviceAuthToken())
+                    .operation(request.getOperation())
+                    .resource(request.getResource())
+                    .build();
+            when(deviceAuthClient.canDevicePerform(authzReq)).thenReturn(true);
             Pair<CompletableFuture<Void>, Consumer<AuthorizeClientDeviceActionResponse>> cb =
                     asyncAssertOnConsumer((m) -> {
                         assertTrue(m.isIsAuthorized());
                     });
             authzClientDeviceAction(ipcClient, request, cb.getRight());
-            cb.getLeft().get(10, TimeUnit.SECONDS);
+            cb.getLeft().get(5, TimeUnit.SECONDS);
         }
     }
 
@@ -146,13 +151,18 @@ class AuthorizeClientDeviceActionTest {
                     .withOperation("some-invalid-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-token");
-            when(deviceAuthClient.canDevicePerform(any())).thenReturn(false);
+            AuthorizationRequest authzReq = AuthorizationRequest.builder()
+                    .sessionId(request.getClientDeviceAuthToken())
+                    .operation(request.getOperation())
+                    .resource(request.getResource())
+                    .build();
+            when(deviceAuthClient.canDevicePerform(authzReq)).thenReturn(false);
             Pair<CompletableFuture<Void>, Consumer<AuthorizeClientDeviceActionResponse>> cb =
                     asyncAssertOnConsumer((m) -> {
                         assertFalse(m.isIsAuthorized());
                     });
             authzClientDeviceAction(ipcClient, request, cb.getRight());
-            cb.getLeft().get(10, TimeUnit.SECONDS);
+            cb.getLeft().get(5, TimeUnit.SECONDS);
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
```diff
@@ -0,0 +1,131 @@
...
+                logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+                throw new UnauthorizedError(e.getMessage());
+            }
+            AuthorizationRequest authorizationRequest = getAuthzRequest(request);
+            try {
+                boolean isAuthorized = deviceAuthClient.canDevicePerform(authorizationRequest);
+                AuthorizeClientDeviceActionResponse response = new AuthorizeClientDeviceActionResponse();
+                return response.withIsAuthorized(isAuthorized);
+            }  catch (com.aws.greengrass.device.exception.AuthorizationException e) {
+                logger.atError().cause(e).log("Unable to find a valid session with the given auth token");
```
MikeDombo, line 79: this is the only possible cause of that exception type? I think that should have a more specific error

saranyailla: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/main/java/com/aws/greengrass/device/DeviceAuthClient.java#L171  and `findSession` is checking for the session in a map.

MikeDombo: That's not the only cause though: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/main/java/com/aws/greengrass/device/DeviceAuthClient.java#L180

DeviceAuthClient should throw a subclass for a bad auth token as we'd like to tell those apart from other issues.

saranyailla: Actually, I posted wrong line reference previously. Updated it later.

MikeDombo: Yes I noticed. My comment still stands though, it should be a subtype so we can tell these issues apart.

src/main/java/com/aws/greengrass/device/DeviceAuthClient.java
```diff
@@ -196,9 +196,9 @@ private Thing getValidThing(String thingName, Certificate certificate) {
      *
      * @param request authorization request including operation, resource, sessionId, clientId
      * @return if device is authorized
-     * @throws AuthorizationException if session is invalid
+     * @throws InvalidSessionException if session is invalid
      */
-    public boolean canDevicePerform(AuthorizationRequest request) throws AuthorizationException {
+    public boolean canDevicePerform(AuthorizationRequest request) throws InvalidSessionException {
```
MikeDombo, line 201: do not change the signature of this method.

src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
```diff
@@ -0,0 +1,132 @@
...
+                logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+                throw new UnauthorizedError(e.getMessage());
+            }
+            AuthorizationRequest authorizationRequest = getAuthzRequest(request);
+            try {
+                boolean isAuthorized = deviceAuthClient.canDevicePerform(authorizationRequest);
+                AuthorizeClientDeviceActionResponse response = new AuthorizeClientDeviceActionResponse();
+                return response.withIsAuthorized(isAuthorized);
+            } catch (InvalidSessionException e) {
+                logger.atError().cause(e).log("Unable to find a valid session with the given auth token");
```
MikeDombo, line 80: since the exception is so specific here, don't log the `cause` field as that will emit a stacktrace we don't care about. The error message here is clear enough, do include the exception's error message though.

src/main/java/com/aws/greengrass/device/exception/InvalidSessionException.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device.exception;
+
+public class InvalidSessionException extends Exception {
```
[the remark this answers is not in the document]
saranyailla: Right, I thought about it but not sure why an invalid session should be tied to an AuthorizationException. But if we want to avoid refactoring moquette code for this, I'll make it subtype of `AuthorizationException`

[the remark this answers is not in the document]
MikeDombo: It is a type of authorization exception. At a high level, we cannot authorize the session. The reason why we can't authorize it is because there isn't any session. Adding it as a subclass, IMO is correct object modeling, and has the benefit of keeping the method's interface the same.

src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
saranyailla: I'll remove the cause but keep the error message same as the exception's error message talks about invalid session where customers provided auth token in the request.

MikeDombo (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -0,0 +1,257 @@
...
+@ExtendWith({GGExtension.class, UniqueRootPathExtension.class, MockitoExtension.class})
+class AuthorizeClientDeviceActionTest {
+    private static GlobalStateChangeListener listener;
+    @TempDir
+    Path rootDir;
+    private Kernel kernel;
+    @Mock
+    private GreengrassServiceClientFactory clientFactory;
+    @Mock
+    private GreengrassV2DataClient client;
```
jbutler, line 66: Nit: v2DataClient

src/main/java/com/aws/greengrass/device/exception/InvalidSessionException.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device.exception;
+
+public class InvalidSessionException extends AuthorizationException {
+    static final long serialVersionUID = -1L;
```
jbutler, line 9: nit: private

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html

src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
```diff
@@ -0,0 +1,132 @@
...
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.AUTHORIZE_CLIENT_DEVICE_ACTION;
+
+public class AuthorizeClientDeviceActionOperationHandler
+        extends GeneratedAbstractAuthorizeClientDeviceActionOperationHandler {
+    private static final Logger logger = LogManager.getLogger(AuthorizeClientDeviceActionOperationHandler.class);
+    private static final String COMPONENT_NAME = "componentName";
+    private static final String UNAUTHORIZED_ERROR = "Not Authorized";
+    private static final String NO_AUTH_TOKEN_ERROR = "Auth token is required";
+    private static final String NO_OPERATION_ERROR = "Mqtt operation is required";
+    private static final String NO_RESOURCE_ERROR = "Mqtt resource is required";
```
jbutler, line 39: Remove "mqtt". This is operation agnostic

src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
```diff
@@ -0,0 +1,132 @@
...
+                logger.atWarn().kv("error", e.getMessage()).kv(COMPONENT_NAME, serviceName).log(UNAUTHORIZED_ERROR);
+                throw new UnauthorizedError(e.getMessage());
+            }
+            AuthorizationRequest authorizationRequest = getAuthzRequest(request);
+            try {
+                boolean isAuthorized = deviceAuthClient.canDevicePerform(authorizationRequest);
+                AuthorizeClientDeviceActionResponse response = new AuthorizeClientDeviceActionResponse();
+                return response.withIsAuthorized(isAuthorized);
+            } catch (InvalidSessionException e) {
+                logger.atError().log("Unable to find a valid session with the given auth token");
```
jbutler, line 80: warn

src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
```diff
@@ -0,0 +1,132 @@
...
+                throw new InvalidClientDeviceAuthTokenError(
+                        "Unable to find a valid session with the given auth token. Check Greengrass log for details.");
+            } catch (Exception e) {
+                logger.atError().cause(e).log("Unable to authorize the client device action");
```
jbutler, line 84: This isn't really about not authorizing the client device action. It's a legitimate failure that shouldn't occur. Log something like "Unhandled exception while authorizing client device action" to make it more clear

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -0,0 +1,257 @@
...
+        kernel.getContext().put(DeviceAuthClient.class, deviceAuthClient);
+        startNucleusWithConfig("cda.yaml");
+        try (EventStreamRPCConnection connection = IPCTestUtils.getEventStreamRpcConnection(kernel,
+                "BrokerWithAuthorizeClientDeviceActionPermission")) {
+            GreengrassCoreIPCClient ipcClient = new GreengrassCoreIPCClient(connection);
+            AuthorizeClientDeviceActionRequest request = new AuthorizeClientDeviceActionRequest()
+                    .withOperation("some-action")
+                    .withResource("some-resource")
+                    .withClientDeviceAuthToken("some-token");
+            when(deviceAuthClient.canDevicePerform(any())).thenReturn(true);
```
jbutler, line 128: This expectation is too broad. This test should change `any()` to ensure the correct auth request is passed. It's okay for the other tests to use `any()`

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -0,0 +1,257 @@
...
+            Pair<CompletableFuture<Void>, Consumer<AuthorizeClientDeviceActionResponse>> cb =
+                    asyncAssertOnConsumer((m) -> {
+                        assertTrue(m.isIsAuthorized());
+                    });
+            authzClientDeviceAction(ipcClient, request, cb.getRight());
+            cb.getLeft().get(10, TimeUnit.SECONDS);
```
jbutler, line 134: Does this really need a full 10 seconds?

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -0,0 +1,257 @@
...
+                    .withOperation("some-invalid-action")
+                    .withResource("some-resource")
+                    .withClientDeviceAuthToken("some-token");
+            when(deviceAuthClient.canDevicePerform(any())).thenReturn(false);
+            Pair<CompletableFuture<Void>, Consumer<AuthorizeClientDeviceActionResponse>> cb =
+                    asyncAssertOnConsumer((m) -> {
+                        assertFalse(m.isIsAuthorized());
+                    });
+            authzClientDeviceAction(ipcClient, request, cb.getRight());
+            cb.getLeft().get(10, TimeUnit.SECONDS);
```
jbutler, line 155: Same here

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -0,0 +1,257 @@
...
+    void GIVEN_brokerWithInvalidAuthzRequest_WHEN_AuthorizeClientDeviceAction_THEN_throwInvalidArgumentsError()
+            throws Exception {
+        kernel.getContext().put(DeviceAuthClient.class, deviceAuthClient);
+        startNucleusWithConfig("cda.yaml");
+        try (EventStreamRPCConnection connection = IPCTestUtils.getEventStreamRpcConnection(kernel,
+                "BrokerWithAuthorizeClientDeviceActionPermission")) {
+            GreengrassCoreIPCClient ipcClient = new GreengrassCoreIPCClient(connection);
+            AuthorizeClientDeviceActionRequest request = new AuthorizeClientDeviceActionRequest()
+                    .withOperation("some-action")
+                    .withResource(null)
```
jbutler, line 169: What if it's empty?

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
saranyailla: It shouldn't. Updating it to 5.

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
saranyailla: It is considered same as `null`.

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -73,7 +73,7 @@ private static void authzClientDeviceAction(GreengrassCoreIPCClient ipcClient,
             throws Exception {
         AuthorizeClientDeviceActionResponseHandler handler =
                 ipcClient.authorizeClientDeviceAction(request, Optional.empty());
-        AuthorizeClientDeviceActionResponse response = handler.getResponse().get(10, TimeUnit.SECONDS);
+        AuthorizeClientDeviceActionResponse response = handler.getResponse().get(5, TimeUnit.SECONDS);
```
jbutler, line 76: I'm surprised this would take longer than a couple milliseconds. What's so slow?

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -125,13 +125,18 @@ void GIVEN_brokerWithAuthorizedAction_WHEN_AuthorizeClientDeviceAction_THEN_retu
                     .withOperation("some-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-token");
-            when(deviceAuthClient.canDevicePerform(any())).thenReturn(true);
+            AuthorizationRequest authzReq = AuthorizationRequest.builder()
```
jbutler, line 128: What you have here is good - keep this. You want to ensure that the IPC layer passes the correct arguments, and this achieves that

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
saranyailla: I don't see it being slow for any reason but how does that impact the tests? Are we trying to fail-fast if there's actually an issue that makes it take more than a few milliseconds?

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -146,13 +151,18 @@ void GIVEN_brokerWithUnauthorizedAction_WHEN_AuthorizeClientDeviceAction_THEN_re
                     .withOperation("some-invalid-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-token");
-            when(deviceAuthClient.canDevicePerform(any())).thenReturn(false);
+            AuthorizationRequest authzReq = AuthorizationRequest.builder()
```
jbutler, line 154: Remove this one though.

Only mock the behavior you want to test. You've already tested that the correct arguments are passed. So now you're just interested in showing that IPC behaves correctly when it receives the various responses. The mock should return `false` no matter what. This is just unnecessarily increasing coupling

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
```diff
@@ -188,7 +198,12 @@ void GIVEN_brokerWithInvalidAuthzRequest_WHEN_AuthorizeClientDeviceAction_THEN_t
                     .withOperation("some-action")
                     .withResource("some-resource")
                     .withClientDeviceAuthToken("some-invalid-token");
-            when(deviceAuthClient.canDevicePerform(any())).thenThrow(InvalidSessionException.class);
+            AuthorizationRequest authzReq = AuthorizationRequest.builder()
```
jbutler, line 201: Same here

src/main/java/com/aws/greengrass/ipc/AuthorizeClientDeviceActionOperationHandler.java
```diff
@@ -35,7 +35,7 @@
     private static final String COMPONENT_NAME = "componentName";
     private static final String UNAUTHORIZED_ERROR = "Not Authorized";
     private static final String NO_AUTH_TOKEN_ERROR = "Auth token is required";
-    private static final String NO_OPERATION_ERROR = "Mqtt operation is required";
+    private static final String NO_OPERATION_ERROR = "Operation is required";
     private static final String NO_RESOURCE_ERROR = "Mqtt resource is required";
```
jbutler, line 39: Same here

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/AuthorizeClientDeviceActionTest.java
jbutler: Failing fast, yes. But also, if it really is taking 1+ seconds, then that's a real issue.

jbutler (approved):

MikeDombo (approved):

PR #72, opened 2022-05-04
fix: add state to update shadow request

jcosentino11:
**Issue #, if available:**

**Description of changes:**

This change fixes the handling of CIS shadow delta updates within `CISShadowMonitor` and adds more exception logging:

* `UpdateShadowRequest#state` field is now populated. Previously, this was causing a `NullPointerException` when the request was processed by `IotShadowClient`.
* `lastVersion` is **always** set after certs have been rotated, to prevent unnecessary cert rotations if the same shadow delta version is processed twice.
* All exceptions are now logged via `CompletableFuture#exceptionally`, when previously some were silently ignored.
* `CISShadowMonitorTest` is completely refactored and uses a fake implementation of `IotShadowClient`.  This allows the test to hit the `NullPointerException` code path mentioned above.

* To prevent unnecessary certificate rotations
* To guarantee logs are provided in error cases when handling CIS shadow delta updates

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
@@ -14,6 +14,8 @@ import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.mqttclient.WrapperMqttClientConnection;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.RetryUtils;
+import lombok.AccessLevel;
+import lombok.Getter;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnection;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
 import software.amazon.awssdk.crt.mqtt.MqttException;
@@ -22,6 +24,7 @@ import software.amazon.awssdk.iot.iotshadow.IotShadowClient;
 import software.amazon.awssdk.iot.iotshadow.model.GetShadowRequest;
 import software.amazon.awssdk.iot.iotshadow.model.GetShadowSubscriptionRequest;
 import software.amazon.awssdk.iot.iotshadow.model.ShadowDeltaUpdatedSubscriptionRequest;
+import software.amazon.awssdk.iot.iotshadow.model.ShadowState;
 import software.amazon.awssdk.iot.iotshadow.model.UpdateShadowRequest;
 import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
@@ -29,6 +32,7 @@ import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingExceptio
 import java.security.KeyStoreException;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
@@ -41,7 +45,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.inject.Inject;
 
-@SuppressWarnings("PMD.ImmutableField")
+@SuppressWarnings({"PMD.ImmutableField", "PMD.AvoidDuplicateLiterals"})
 public class CISShadowMonitor {
     private static final Logger LOGGER = LogManager.getLogger(CISShadowMonitor.class);
     private static final String CIS_SHADOW_SUFFIX = "-gci";
@@ -68,6 +72,7 @@ public class CISShadowMonitor {
     private final ConnectivityInfoProvider connectivityInfoProvider;
     private final AtomicInteger nextVersion = new AtomicInteger(-1);
 
+    @Getter(AccessLevel.PACKAGE) // for unit tests
     private final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
@@ -206,7 +211,6 @@ public class CISShadowMonitor {
     private synchronized void handleNewCloudVersion(int newVersion) {
         if (newVersion == lastVersion) {
             LOGGER.atInfo().kv("version", newVersion).log("Already processed version. Skipping cert re-generation");
-            reportVersion(newVersion);
             return;
         }
 
@@ -243,20 +247,37 @@ public class CISShadowMonitor {
                 for (CertificateGenerator cg : monitoredCertificateGenerators) {
                     cg.generateCertificate(connectivityInfoProvider::getCachedHostAddresses);
                 }
-                reportVersion(version);
-                lastVersion = version;
+
+                try {
+                    reportVersion(version);
+                } catch (ExecutionException e) {
+                    Throwable cause = e.getCause() == null ? e : e.getCause();
+                    LOGGER.atWarn().kv("version", version).cause(cause).log("Unable to report CIS shadow version");
+                } catch (TimeoutException e) {
+                    LOGGER.atWarn().kv("version", version).log("Timed out while reporting CIS shadow version");
+                } catch (InterruptedException e) {
+                    LOGGER.atWarn().kv("version", version).log("Interrupted while reporting CIS shadow version");
+                } catch (Exception e) {
+                    LOGGER.atWarn().kv("version", version).cause(e).log("Unable to report CIS shadow version");
+                } finally {
+                    lastVersion = version;
+                }
             } catch (KeyStoreException e) {
                 LOGGER.atError().cause(e).log("failed to generate new certificates");
             }
         });
     }
 
-    private void reportVersion(int version) {
+    private void reportVersion(int version) throws ExecutionException, InterruptedException, TimeoutException {
         LOGGER.atInfo().kv("version", version).log("Reporting version");
         UpdateShadowRequest updateShadowRequest = new UpdateShadowRequest();
         updateShadowRequest.thingName = shadowName;
         updateShadowRequest.version = version;
-        iotShadowClient.PublishUpdateShadow(updateShadowRequest, QualityOfService.AT_LEAST_ONCE);
+        updateShadowRequest.state = new ShadowState();
+        updateShadowRequest.state.reported = new HashMap<>();
+        updateShadowRequest.state.desired = new HashMap<>();
+        iotShadowClient.PublishUpdateShadow(updateShadowRequest, QualityOfService.AT_LEAST_ONCE)
+                .get(5, TimeUnit.SECONDS);
     }
 
     private void publishToGetCISShadowTopic() {
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
@@ -261,7 +261,7 @@ public class CISShadowMonitor {
                 lastVersion = newVersion;
             }
         }, executorService).exceptionally(e -> {
-            LOGGER.atError().kv(VERSION, newVersion).cause(e).log("Unable to handle cis shadow delta");
+            LOGGER.atError().kv(VERSION, newVersion).cause(e).log("Unable to handle CIS shadow delta");
             return null;
         });
     }
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
@@ -267,7 +267,7 @@ public class CISShadowMonitor {
     }
 
     private CompletableFuture<Integer> reportVersion(int version) {
-        LOGGER.atInfo().kv(VERSION, version).log("Reporting version");
+        LOGGER.atInfo().kv(VERSION, version).log("Reporting CIS version");
         UpdateShadowRequest updateShadowRequest = new UpdateShadowRequest();
         updateShadowRequest.thingName = shadowName;
         updateShadowRequest.version = version;
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
@@ -234,7 +234,7 @@ public class CISShadowMonitor {
                 RetryUtils.runWithRetry(GET_CONNECTIVITY_RETRY_CONFIG, connectivityInfoProvider::getConnectivityInfo,
                         "get-connectivity", LOGGER);
             } catch (InterruptedException e) {
-                LOGGER.atWarn().kv(VERSION, version).cause(e)
+                LOGGER.atDebug().kv(VERSION, version).cause(e)
                         .log("Retry workflow for getting connectivity info interrupted");
                 Thread.currentThread().interrupt();
                 return;
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
@@ -45,10 +45,11 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.inject.Inject;
 
-@SuppressWarnings({"PMD.ImmutableField", "PMD.AvoidDuplicateLiterals"})
+@SuppressWarnings("PMD.ImmutableField")
 public class CISShadowMonitor {
     private static final Logger LOGGER = LogManager.getLogger(CISShadowMonitor.class);
     private static final String CIS_SHADOW_SUFFIX = "-gci";
+    private static final String VERSION = "version";
     private static final long TIMEOUT_FOR_SUBSCRIBING_TO_TOPICS_SECONDS = Duration.ofMinutes(1).getSeconds();
     private static final long WAIT_TIME_TO_SUBSCRIBE_AGAIN_IN_MS = Duration.ofMinutes(2).toMillis();
     private static final Random JITTER = new Random();
@@ -210,7 +211,7 @@ public class CISShadowMonitor {
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private synchronized void handleNewCloudVersion(int newVersion) {
         if (newVersion == lastVersion) {
-            LOGGER.atInfo().kv("version", newVersion).log("Already processed version. Skipping cert re-generation");
+            LOGGER.atInfo().kv(VERSION, newVersion).log("Already processed version. Skipping cert re-generation");
             return;
         }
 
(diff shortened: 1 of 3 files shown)


MikeDombo:

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -243,20 +247,37 @@ private synchronized void handleNewCloudVersion(int newVersion) {
...
-                lastVersion = version;
+
+                try {
+                    reportVersion(version);
+                } catch (ExecutionException e) {
+                    Throwable cause = e.getCause() == null ? e : e.getCause();
+                    LOGGER.atWarn().kv("version", version).cause(cause).log("Unable to report CIS shadow version");
+                } catch (TimeoutException e) {
+                    LOGGER.atWarn().kv("version", version).log("Timed out while reporting CIS shadow version");
+                } catch (InterruptedException e) {
```
MikeDombo, line 258: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -41,7 +45,7 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.inject.Inject;
 
-@SuppressWarnings("PMD.ImmutableField")
+@SuppressWarnings({"PMD.ImmutableField", "PMD.AvoidDuplicateLiterals"})
```
MikeDombo, line 48: don't ignore this, extract duplicate literals to a constant

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -243,20 +247,37 @@ private synchronized void handleNewCloudVersion(int newVersion) {
...
         LOGGER.atInfo().kv("version", version).log("Reporting version");
         UpdateShadowRequest updateShadowRequest = new UpdateShadowRequest();
         updateShadowRequest.thingName = shadowName;
         updateShadowRequest.version = version;
-        iotShadowClient.PublishUpdateShadow(updateShadowRequest, QualityOfService.AT_LEAST_ONCE);
+        updateShadowRequest.state = new ShadowState();
+        updateShadowRequest.state.reported = new HashMap<>();
+        updateShadowRequest.state.desired = new HashMap<>();
+        iotShadowClient.PublishUpdateShadow(updateShadowRequest, QualityOfService.AT_LEAST_ONCE)
+                .get(5, TimeUnit.SECONDS);
```
MikeDombo, line 280: I don't think we should be waiting here, this is called in a thread that shouldn't be blocked. You can chain the future to run code on success or error

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
MikeDombo: You haven't dealt with this exception correctly @jcosentino11, you must re-interrupt the thread using `Thread.currentThread().interrupt()`. Thread interruption means that someone wants your thread to shutdown, and catching the interrupted exception clears the thread interrupt flag. You must reset the flag.

jcosentino11: Will add---I think my confusion is that the interruption is happening within a completablefuture, which won't get interrupted on cancel like a regular future, so wasn't actually sure how that code path would happen.

MikeDombo: A Completable future is a future, and calling `.cancel(true)` will attempt to interrupt the thread running the task.

jcosentino11: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#cancel-boolean-
implementation is different, it'll complete the future with a cancellation exception

MikeDombo: The future completes with that exception, but the thread running the future is interrupted. This code here will be interrupted because it is waiting for a different future to complete.

jcosentino11: im probably overthinking this 😓 ,  regardless, this catch clause is now removed since the publish get call was removed 😄

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -205,58 +208,72 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+        }, executorService).exceptionally(e -> {
+            LOGGER.atError().kv(VERSION, newVersion).cause(e).log("Unable to handle cis shadow delta");
+            return null;
         });
     }
 
-    private void reportVersion(int version) {
-        LOGGER.atInfo().kv("version", version).log("Reporting version");
+    private CompletableFuture<Integer> reportVersion(int version) {
+        LOGGER.atInfo().kv(VERSION, version).log("Reporting version");
```
MikeDombo, line 269: reporting CIS version

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -205,58 +208,72 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+                        .exceptionally(e -> {
+                            LOGGER.atWarn().kv(VERSION, newVersion).cause(e)
+                                    .log("Unable to report CIS shadow version");
+                            return null;
+                        });
+            } finally {
+                lastVersion = newVersion;
             }
+        }, executorService).exceptionally(e -> {
+            LOGGER.atError().kv(VERSION, newVersion).cause(e).log("Unable to handle cis shadow delta");
```
MikeDombo, line 263: capital CIS here and anywhere else

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -205,58 +208,72 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
-                        .log("Failed to get connectivity info from cloud. Check that the core device's IoT policy "
-                                + "grants the greengrass:GetConnectivityInfo permission.");
-                throw new CloudServiceInteractionException("Failed to get connectivity info", e);
+                LOGGER.atError().kv(VERSION, newVersion).cause(e)
+                        .log("Failed to get connectivity info from cloud."
+                                + " Check that the core device's IoT policy "
+                                + "grants the greengrass:GetConnectivityInfo permission");
+                return;
+            } finally {
+                connectivityCallInProgress.set(false);
```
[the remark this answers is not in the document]
jcosentino11: was thinking about this earlier, so if we synchronize everything [1. CIS call, 2. cert gen, 3. publish shadow], then if we get a shadow delta anytime after 1., then we'd miss a rotation although probably unlikely?  

but keeping the original/current design, there could be interleaving during 1. and 2., with some (small) chance that some cert generators use the wrong connectivity info.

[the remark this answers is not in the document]
jcosentino11: a probably over-engineered solution could be to push to a shared queue on each shadow delta, when no CIS call is in progress.  a separate thread loops, picks up oldest queue message, discarding the rest, then runs CIS call/ cert gen.

would ensure we don't miss rotations, while also not over-rotating 🤷

MikeDombo (dismissed):

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -49,172 +77,507 @@
...
+
         cisShadowMonitor.startMonitor();
+        cisShadowMonitor.addToMonitor(certificateGenerator);
+
+        shadowClient.withDuplicatePublishing(true);
+
+        publishDesiredShadowState(Utils.immutableMap("newState", 1));
+
+        assertTrue(shadowDeltaUpdated.await(5L, TimeUnit.SECONDS));
+        verify(certificateGenerator).generateCertificate(any(), any());
```
vaibhavmurkute, line 191: verify generateCertificate() is called only _once_?

jcosentino11: Yep, duplicate calls to the shadow delta handler are idempotent, we'll only generate the cert the first time

vaibhavmurkute: I mean we should explicitly verify here that `generateCertificate()` is called only once:
```
verify(certificateGenerator, times(1)).generateCertificate(any(), any());
```

jcosentino11: Ahh okay.  I personally think `times(1)` is verbose, but looks like it's widely used in the codebase, will adjust :)

MikeDombo: There is also `once` which is a synonym.

jcosentino11: done

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
jbutler: @MikeDombo that's not true. We're calling this from an executor service, not the CRT thread

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -205,59 +208,73 @@ private void subscribeToShadowTopics() throws InterruptedException {
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private synchronized void handleNewCloudVersion(int newVersion) {
         if (newVersion == lastVersion) {
-            LOGGER.atInfo().kv("version", newVersion).log("Already processed version. Skipping cert re-generation");
-            reportVersion(newVersion);
+            LOGGER.atInfo().kv(VERSION, newVersion).log("Already processed version. Skipping cert re-generation");
```
jbutler, line 211: The fact that this is getting called means there is a shadow delta so you'll want to attempt to re-report

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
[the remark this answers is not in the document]
jbutler: Whenever you tell yourself "this scenarios seem unlikely", remember that this code is executing on millions of devices. At a certain scale, "can it happen?" and "will it happen?" tend to have the same answer. Be extremely paranoid about concurrency bugs. We have a lot of them.

BTW, we've had ordering bugs with this specific code on V1. It's not a theoretical issue.

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -49,172 +77,507 @@
...
-        assertThat(getSubscribeRequest.getValue().thingName, is(SHADOW_NAME));
+        IntStream.range(0, numShadowChanges).forEach(i -> {
+            publishDesiredShadowState(Utils.immutableMap("newState", i));
+        });
 
-        verify(mockShadowClient, times(1)).PublishGetShadow(publishRequest.capture(),
-                eq(QualityOfService.AT_LEAST_ONCE));
-        assertThat(publishRequest.getValue().thingName, is(SHADOW_NAME));
+        assertTrue(shadowDeltaUpdated.await(5L, TimeUnit.SECONDS));
+        verify(certificateGenerator, times(numShadowChanges)).generateCertificate(any(), any());
```
jbutler, line 176: This might be okay, but it seems fairly coupled to the implementation. Might be better to test state rather than behavior in this case

jcosentino11: I'm not sure I see the coupling---I believe this is testing the high-level "interface" of the monitor:  if we get 3 shadow deltas, we'd expect 3 rotations.

maybe we need a separate integration test?

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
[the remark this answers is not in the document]
jcosentino11: 👍  going to address this in a follow-up PR

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
jbutler: Not necessarily. If you get 3 shadow deltas, you just expect that `generateCertificate` is called with the latest connectivity info. Imagine if you add some queueing de-dup logic to the code. This test will become non-deterministic

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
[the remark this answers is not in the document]
jcosentino11: addressed in https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/76

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -202,62 +210,89 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+                LOGGER.atDebug().log("getConnectivityInfo call in progress. Skipping cert re-generation");
+                return;
+            }
             try {
                 RetryUtils.runWithRetry(GET_CONNECTIVITY_RETRY_CONFIG, connectivityInfoProvider::getConnectivityInfo,
                         "get-connectivity", LOGGER);
             } catch (InterruptedException e) {
-                LOGGER.atWarn().cause(e).log("Retry workflow for getting connectivity info interrupted");
+                LOGGER.atWarn().kv(VERSION, version).cause(e)
+                        .log("Retry workflow for getting connectivity info interrupted");
```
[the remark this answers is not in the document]
jcosentino11: fixed

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -202,62 +210,89 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+
         // NOTE: This method executes in an MQTT CRT thread. Since certificate generation is a compute intensive
         // operation (particularly on low end devices) it is imperative that we process this event asynchronously
         // to avoid blocking other MQTT subscribers in the Nucleus
-        getConnectivityFuture = CompletableFuture.supplyAsync(() -> {
+        CompletableFuture.runAsync(() -> {
+            if (connectivityCallInProgress.getAndSet(true)) {
+                LOGGER.atDebug().log("getConnectivityInfo call in progress. Skipping cert re-generation");
+                return;
+            }
```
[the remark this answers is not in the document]
jcosentino11: done 👍

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -8,213 +8,717 @@
...
-        verify(mockCertGenerator, times(1)).generateCertificate(any(), any());
-        ArgumentCaptor<UpdateShadowRequest> updateShadowRequest = ArgumentCaptor.forClass(UpdateShadowRequest.class);
-        verify(mockShadowClient, times(1)).PublishUpdateShadow(updateShadowRequest.capture(),
-                eq(QualityOfService.AT_LEAST_ONCE));
-        assertThat(updateShadowRequest.getValue().version, is(1));
+        verify(certificateGenerator, times(numShadowChanges)).generateCertificate(any(), any());
+    }
+
+    @Test
+    void GIVEN_CISShadowMonitor_WHEN_cis_shadow_changed_multiple_times_and_shadow_update_failed_THEN_new_cert_generated_multiple_times(ExtensionContext context) throws Exception {
```
jbutler, line 281: Can probably be combined with the previous test

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -8,213 +8,717 @@
...
+        CountDownLatch shadowDeltaUpdated = whenShadowDeltaUpdated(numShadowChanges * 2);
+
         cisShadowMonitor.startMonitor();
+        cisShadowMonitor.addToMonitor(certificateGenerator);
 
-        ArgumentCaptor<Consumer<ShadowDeltaUpdatedEvent>> updateSubscribeHandler = ArgumentCaptor.forClass(
-                Consumer.class);
-        verify(mockShadowClient, times(1)).SubscribeToShadowDeltaUpdatedEvents(any(), any(),
-                updateSubscribeHandler.capture(), any());
+        shadowClient.withDuplicatePublishing(true);
```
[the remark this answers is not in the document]
jcosentino11: agreed, just refactored everything out of the fake so tests are hopefully more straightforward

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -8,213 +8,717 @@
...
+        CountDownLatch shadowDeltaUpdated = whenShadowDeltaUpdated(numShadowChanges);
+
         cisShadowMonitor.startMonitor();
+        cisShadowMonitor.addToMonitor(certificateGenerator);
+
+        shadowClient.onPrePublish(topic -> {
+            if (topic.endsWith("shadow/update/delta")) {
+                shadowClient.withPublishException(true);
+            }
+        });
```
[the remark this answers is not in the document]
jcosentino11: agreed, just refactored everything out of the fake so tests are hopefully more straightforward

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -8,213 +8,586 @@
...
+        assertEquals(0, numSubscriptions.get());
+    }
+
+    /**
+     * Verify that certificates are rotated only when connectivity info changes.
+     *
+     * @throws KeyStoreException n/a
+     */
+    private void verifyCertsRotatedWhenConnectivityChanges() throws KeyStoreException {
+        verify(certificateGenerator, times(connectivityInfoProvider.getNumUniqueConnectivityInfoResponses())).generateCertificate(any(), any());
```
jbutler, line 429: Nice

jbutler (approved):

MikeDombo (approved):

PR #73, opened 2022-05-09
chore: use new fetchGreengrassV2DataClient API

J-C-L:
**Issue #, if available:**

**Description of changes:**
Updating code to use new fetchGreengrassV2DataClient API. 

This API throws DeviceConfigurationException instead of returning null when the client is not available due to a config error. This avoids an NPE.

### DIFF
--- a/src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
+++ b/src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
@@ -6,17 +6,22 @@
 package com.aws.greengrass.cisclient;
 
 import com.aws.greengrass.deployment.DeviceConfiguration;
+import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.GreengrassServiceClientFactory;
+import com.aws.greengrass.util.RetryUtils;
+import software.amazon.awssdk.services.greengrassv2data.GreengrassV2DataClient;
 import software.amazon.awssdk.services.greengrassv2data.model.ConnectivityInfo;
 import software.amazon.awssdk.services.greengrassv2data.model.GetConnectivityInfoRequest;
 import software.amazon.awssdk.services.greengrassv2data.model.GetConnectivityInfoResponse;
 import software.amazon.awssdk.services.greengrassv2data.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.greengrassv2data.model.ValidationException;
 
+import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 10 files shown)


naikvidish (approved):

src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
```diff
@@ -33,6 +38,12 @@ public class ConnectivityInfoProvider {
 
     private volatile List<String> cachedHostAddresses = Collections.emptyList();
 
+    public static final int CLIENT_RETRY_COUNT = 3;
+    private static final RetryUtils.RetryConfig CLIENT_EXCEPTION_RETRY_CONFIG =
+            RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(30))
+                    .maxRetryInterval(Duration.ofSeconds(30)).maxAttempt(CLIENT_RETRY_COUNT)
+                    .retryableExceptions(Arrays.asList(DeviceConfigurationException.class)).build();
```
MikeDombo, line 45: I don't think we should be retrying this, it isn't going to fix itself. Retries should be handled at a higher level.

J-C-L: Retrying to cover the edge case where client is requested at the same time configs are being updated, so client returns null, but an immediate retry would be successful.

sushantAtAws: That "immediate" retry will be in 30 seconds as per this though, right? That sounds too much, so does the 3 retries.

src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
```diff
@@ -59,14 +70,19 @@ public List<String> getCachedHostAddresses() {
      * Get connectivity info.
      *
      * @return list of connectivity info items
+     * @throws Exception when not able to retrieve greengrasV2DataClient
```
sushantAtAws, line 73: Throw only the `DeviceConfigurationException` exception. Do you have `Exception` because of the `RetryUtils`?

src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
J-C-L: What would be reasonable? (5 sec and 2 retries?)

src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
J-C-L: yes, RetryUtils required Exception

src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
jbutler: Retries are handled at another layer. Don't retry anything here. Just fail

PR #74, opened 2022-05-10
fix: handle null response from getGreengrassV2DataClient

J-C-L:
**Issue #, if available:**

**Description of changes:**
Handle null response from getGreengrassV2DataClient

Needed to avoid NPE when getGreengrassV2DataClient returns null


MikeDombo:
Wait for #73.

PR #76, opened 2022-05-13
fix: cis shadow monitor concurrency

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Follow up from https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/72 to address `CISShadowMonitor` concurrency issues.

By looking at the commit history in the PR, I tried two approaches:
* Initially, I just synchronized everything.  While correct, a potential issue is that many threads could queue up if the synchronized block took a long time. (see [this snippet](https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/a692fb0da93888b188a41e9212fe0e30fc8feb84/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java#L238-L285))
* To improve on this, a separate queuing mechanism is used rather than synchronized block. When `CISShadowMonitor` receives a delta, it queues up a request to `ShadowProcessor`, which in turn reads from the queue on an interval.

There is a possibility that `CISShadowMonitor` can rotate certs using outdated connectivity info, due to interleaving between getConnectivityInfo call and cert generation loop.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
@@ -14,6 +14,8 @@ import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.mqttclient.WrapperMqttClientConnection;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.RetryUtils;
+import lombok.AccessLevel;
+import lombok.Getter;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnection;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
 import software.amazon.awssdk.crt.mqtt.MqttException;
@@ -22,6 +24,7 @@ import software.amazon.awssdk.iot.iotshadow.IotShadowClient;
 import software.amazon.awssdk.iot.iotshadow.model.GetShadowRequest;
 import software.amazon.awssdk.iot.iotshadow.model.GetShadowSubscriptionRequest;
 import software.amazon.awssdk.iot.iotshadow.model.ShadowDeltaUpdatedSubscriptionRequest;
+import software.amazon.awssdk.iot.iotshadow.model.ShadowState;
 import software.amazon.awssdk.iot.iotshadow.model.UpdateShadowRequest;
 import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
@@ -29,6 +32,7 @@ import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingExceptio
 import java.security.KeyStoreException;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
@@ -41,7 +45,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.inject.Inject;
 
-@SuppressWarnings("PMD.ImmutableField")
+@SuppressWarnings({"PMD.ImmutableField", "PMD.AvoidDuplicateLiterals"})
 public class CISShadowMonitor {
     private static final Logger LOGGER = LogManager.getLogger(CISShadowMonitor.class);
     private static final String CIS_SHADOW_SUFFIX = "-gci";
@@ -68,6 +72,7 @@ public class CISShadowMonitor {
     private final ConnectivityInfoProvider connectivityInfoProvider;
     private final AtomicInteger nextVersion = new AtomicInteger(-1);
 
+    @Getter(AccessLevel.PACKAGE) // for unit tests
     private final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
@@ -206,7 +211,6 @@ public class CISShadowMonitor {
     private synchronized void handleNewCloudVersion(int newVersion) {
         if (newVersion == lastVersion) {
             LOGGER.atInfo().kv("version", newVersion).log("Already processed version. Skipping cert re-generation");
-            reportVersion(newVersion);
             return;
         }
 
@@ -243,20 +247,37 @@ public class CISShadowMonitor {
                 for (CertificateGenerator cg : monitoredCertificateGenerators) {
                     cg.generateCertificate(connectivityInfoProvider::getCachedHostAddresses);
                 }
-                reportVersion(version);
-                lastVersion = version;
+
+                try {
+                    reportVersion(version);
+                } catch (ExecutionException e) {
+                    Throwable cause = e.getCause() == null ? e : e.getCause();
+                    LOGGER.atWarn().kv("version", version).cause(cause).log("Unable to report CIS shadow version");
+                } catch (TimeoutException e) {
+                    LOGGER.atWarn().kv("version", version).log("Timed out while reporting CIS shadow version");
+                } catch (InterruptedException e) {
+                    LOGGER.atWarn().kv("version", version).log("Interrupted while reporting CIS shadow version");
+                } catch (Exception e) {
+                    LOGGER.atWarn().kv("version", version).cause(e).log("Unable to report CIS shadow version");
+                } finally {
+                    lastVersion = version;
+                }
             } catch (KeyStoreException e) {
                 LOGGER.atError().cause(e).log("failed to generate new certificates");
             }
         });
     }
 
-    private void reportVersion(int version) {
+    private void reportVersion(int version) throws ExecutionException, InterruptedException, TimeoutException {
         LOGGER.atInfo().kv("version", version).log("Reporting version");
         UpdateShadowRequest updateShadowRequest = new UpdateShadowRequest();
         updateShadowRequest.thingName = shadowName;
         updateShadowRequest.version = version;
-        iotShadowClient.PublishUpdateShadow(updateShadowRequest, QualityOfService.AT_LEAST_ONCE);
+        updateShadowRequest.state = new ShadowState();
+        updateShadowRequest.state.reported = new HashMap<>();
+        updateShadowRequest.state.desired = new HashMap<>();
+        iotShadowClient.PublishUpdateShadow(updateShadowRequest, QualityOfService.AT_LEAST_ONCE)
+                .get(5, TimeUnit.SECONDS);
     }
 
     private void publishToGetCISShadowTopic() {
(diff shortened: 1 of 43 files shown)


src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -87,38 +108,48 @@ public void onConnectionResumed(boolean sessionPresent) {
...
+                     ExecutorService executorService,
+                     String shadowName,
                      ConnectivityInfoProvider connectivityInfoProvider) {
         mqttClient.addToCallbackEvents(callbacks);
         this.connection = connection;
         this.iotShadowClient = iotShadowClient;
         this.executorService = executorService;
         this.shadowName = shadowName;
         this.connectivityInfoProvider = connectivityInfoProvider;
+        this.shadowProcessor = new ShadowProcessor(ses);
```
MikeDombo, line 144: this can, and should be injected

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
```diff
@@ -590,4 +666,82 @@ private static ConnectivityInfo connectivityInfoWithRandomHost() {
                     .build();
         }
     }
+
+    /**
+     * A ScheduledThreadPoolExecutor that executes using a provided ExecutorService.
+     */
+    static class DelegatedScheduledExecutorService extends ScheduledThreadPoolExecutor {
```
MikeDombo, line 673: why? Just create a scheduled executor using `Executors.newScheduledExecutor`

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
jcosentino11: does that work for inner classes?

MikeDombo: Yes, certainly

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
jcosentino11: previous changes removed ses, so was following the spirit of that (https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/60),  totally fine either way

MikeDombo: if you need a ses, then create one the normal way. Doing this is unnecessary and implies that there is a reason why this is needed.

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
jcosentino11: sorry meant to say non-static inner class

src/test/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitorTest.java
jcosentino11: 👍 will fix

jcosentino11: done

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
MikeDombo: Yes should be fine.

jcosentino11: Haven't been able to get this to work, it looks like a chicken-and-egg issue, since nested non-static class depends on its parent


infinite loop resolving dependencies:
```
	at com.aws.greengrass.dependency.Context$Value.getOrCreateArgInstances(Context.java:520)
	at com.aws.greengrass.dependency.Context$Value.constructObjectWithInjection(Context.java:486)
	at com.aws.greengrass.dependency.Context$Value.get(Context.java:436)
	at com.aws.greengrass.dependency.Context.get(Context.java:100)
	at com.aws.greengrass.dependency.Context$Value.getOrCreateArgInstances(Context.java:520)
	at com.aws.greengrass.dependency.Context$Value.constructObjectWithInjection(Context.java:486)
	at com.aws.greengrass.dependency.Context$Value.get(Context.java:436)
	at com.aws.greengrass.dependency.Context.get(Context.java:100)
```

MikeDombo: So make it static or make it not an inner class?

jcosentino11: yeah, but I'm not sure if it's worth it. making it static requires passing in several fields from `CISShadowMonitor` that can't be injected (without adding extra wrappers i guess): `CISShadowMonitor#monitoredCertificateGenerators`, `CISShadowMonitor#shadowName`, `CISShadowMonitor#iotShadowClient`, inner class felt like the cleanest way to accomplish this. While I could just get rid of `ShadowProcessor` alltogether, it felt like a good way to group together functionality without making `CISShadowMontior` feel cluttered 🤷

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -208,31 +252,173 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+        private Future<?> shadowProcessingTask;
+
+        /**
+         * Start processing shadow requests.
+         */
+        void start() {
+            if (shadowProcessingTask != null) {
+                shadowProcessingTask.cancel(true);
+            }
+            shadowProcessingTask = ses.scheduleWithFixedDelay(
```
MikeDombo, line 332: this check, cancel, and set needs to be synchronized. If this method can only ever be called by a single thread, then that's OK.

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -208,31 +252,173 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
         }
 
-        LOGGER.atInfo().log("New CIS version: {}", version);
+        /**
+         * Stop processing shadow requests and interrupt if one is in-progress.
+         */
+        void stop() {
+            if (shadowProcessingTask != null) {
+                shadowProcessingTask.cancel(true);
+            }
```
MikeDombo, line 346: likewise

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
jcosentino11: it's only called as part of `ClientDevicesAuthService#startup`, but it wouldn't hurt to properly synchronize so we don't have to worry about it later

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -208,31 +261,176 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+         * Start processing shadow requests.
+         */
+        void start() {
+            synchronized (shadowProcessingTaskLock) {
+                stop();
+                shadowProcessingTask = ses.scheduleWithFixedDelay(
+                        this::handleShadowProcessingRequest,
+                        0L,
+                        shadowProcessingDelayMs,
+                        TimeUnit.MILLISECONDS
```
jbutler, line 345: I don't think it would take much to get rid of this SES. It seems a bit silly to wake up every 5 seconds to check for an event that should be rare. Can you process shadow events on-demand in a regular executor service? I think it should be fairly straight forward

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -259,53 +464,54 @@ private synchronized void processCISShadow(int version, Map<String, Object> desi
...
+         */
+        @SuppressWarnings("PMD.AvoidCatchingGenericException")
+        private void notifyWorkComplete() {
+            if (onShadowProcessingWorkComplete != null) {
+                try {
+                    onShadowProcessingWorkComplete.run();
+                } catch (Exception e) {
+                    LOGGER.atDebug().cause(e).log("Unable to notify work complete");
+                }
+            }
```
jbutler, line 514: This exists solely for tests, correct? Could you remove this if you used a fake cert generator instead?

src/main/java/com/aws/greengrass/cisclient/ConnectivityInfoProvider.java
```diff
@@ -71,9 +70,10 @@ public List<ConnectivityInfo> getConnectivityInfo() {
             if (getConnectivityInfoResponse.hasConnectivityInfo()) {
                 // Filter out port and metadata since it is not needed
                 connectivityInfoList = getConnectivityInfoResponse.connectivityInfo();
-                cachedHostAddresses = new ArrayList<>(connectivityInfoList.stream()
-                        .map(ci -> ci.hostAddress())
-                        .collect(Collectors.toSet()));
+                cachedHostAddresses = connectivityInfoList.stream()
+                        .map(ConnectivityInfo::hostAddress)
+                        .distinct()
+                        .collect(Collectors.toList());
```
jbutler, line 76: Nice

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
```diff
@@ -208,31 +261,176 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+
+        /**
+         * Request from {@link CISShadowMonitor} to process shadow state.
+         */
+        private final AtomicReference<ShadowProcessingRequest> shadowProcessingRequest = new AtomicReference<>();
+
+        /**
+         * Last received CIS shadow version, used to de-duplicate processing requests.
+         */
+        private final AtomicInteger lastCISShadowVersion = new AtomicInteger();
```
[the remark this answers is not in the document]
jcosentino11: for overkill :) will fix

src/main/java/com/aws/greengrass/certificatemanager/certificate/CISShadowMonitor.java
jcosentino11: ah yeah that's a good point, will rework this

jcosentino11:
closing in favor of https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/106

PR #82, opened 2022-05-20
fix: check thing association with cert id during session creation

saranyailla:
**Issue #, if available:**

**Description of changes:**
The cloud API `VerifyClientDeviceIoTCertificateAssociationRequest`needs certificate with its ID but not certificate pem to verify thing-cert association. This change fixed the certificate object creation in the session creation code where cert pem is used instead of its ID during `isThingAttachedToCertificate` check. 

Fixed the API request parameters. 
 
Added two more unit tests and UATs are in place. 

### DIFF
--- a/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
+++ b/src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
@@ -6,11 +6,13 @@
 package com.aws.greengrass.device.session;
 
 import com.aws.greengrass.device.exception.AuthenticationException;
+import com.aws.greengrass.device.exception.CloudServiceInteractionException;
 import com.aws.greengrass.device.iot.Certificate;
 import com.aws.greengrass.device.iot.IotAuthClient;
 import com.aws.greengrass.device.iot.Thing;
 
 import java.util.Map;
+import java.util.Optional;
 import javax.inject.Inject;
 
 public class MqttSessionFactory implements SessionFactory {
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
```diff
@@ -31,14 +33,22 @@ public Session createSession(Map<String, String> credentialMap) throws Authentic
         //   we don't yet have the ability to authorize actions for those clients. So, for now, assume
         //   this is an IoT Thing and components will be specially handled elsewhere.
         Thing thing = new Thing(mqttCredential.clientId);
-        Certificate cert = new Certificate(mqttCredential.certificatePem);
-        if (!iotAuthClient.isThingAttachedToCertificate(thing, cert)) {
-            throw new AuthenticationException("unable to authenticate device");
+        Optional<String> certificateId;
+        try {
+            certificateId = iotAuthClient.getActiveCertificateId(mqttCredential.certificatePem);
```
MikeDombo, line 38: @jbutler.

By this point we've already done this same call once: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/0a726d2c354c4cefd8e827f40caf2b8cf8d3fcd1/src/main/java/com/aws/greengrass/ipc/VerifyClientDeviceIdentityOperationHandler.java#L109. 

We'd prefer not to do it again right. How do you suggest we make that work?

MikeDombo (approved):

src/main/java/com/aws/greengrass/device/session/MqttSessionFactory.java
jbutler: This is fine for now. We'll need a caching layer for certs and things, which we can add later.

jbutler (approved):

PR #83, opened 2022-05-20
Fix session

saranyailla:
**Issue #, if available:**

**Description of changes:**
Re-encode certificate PEM of the `GetClientDeviceAuthTokenRequest` IPC request if it is not encoded already. 

Same as https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/81


MikeDombo:
Rename this PR with an appropriate commitlint conforming name.

PR #86, opened 2022-05-22
fix: create session for internal components via new SessionManager

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Handle certificates coming from internal components (ie. Bridge).

### DIFF
--- a/src/main/java/com/aws/greengrass/device/DeviceAuthClient.java
+++ b/src/main/java/com/aws/greengrass/device/DeviceAuthClient.java
@@ -118,7 +118,7 @@ public class DeviceAuthClient {
     private boolean isGreengrassComponent(CertPath certPath) {
         try {
             X509Certificate caCertificate = certificateStore.getCACertificate();
-            if (caCertificate == null) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/device/DeviceAuthClient.java
```diff
@@ -216,6 +223,10 @@ public boolean canDevicePerform(AuthorizationRequest request) throws Authorizati
             throw new InvalidSessionException(
                     String.format("Invalid session ID (%s)", request.getSessionId()));
         }
+        // Allow all operations from internal components
```
jbutler, line 226: Can you add a comment that the workaround above still needs to be there since Moquette isn't using the IPC interface?

jbutler (dismissed):

src/main/java/com/aws/greengrass/device/DeviceAuthClient.java
MikeDombo: right.

jbutler (dismissed):

jbutler (approved):

jbutler (approved):

PR #88, opened 2022-06-06
feat: LRU session cache

vaibhavmurkute:
**Description of changes:** LRU session Cache with configurable capacity.

**How was this change tested:** unit tests (`mvn clean package`)

### DIFF
--- a/src/main/java/com/aws/greengrass/device/session/SessionConfig.java
+++ b/src/main/java/com/aws/greengrass/device/session/SessionConfig.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device.session;
+
+import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.util.Coerce;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class SessionConfig {
+    private static final Logger LOGGER = LogManager.getLogger(SessionConfig.class);
+    // arbitrarily set default
+    protected static final int DEFAULT_SESSION_CAPACITY = 100_000;
+    public static final String CLIENT_DEVICE_AUTH_SESSION_CAPACITY_TOPIC = "clientDeviceAuthSessionCapacity";
+
+    private final AtomicInteger clientDeviceAuthSessionCapacity = new AtomicInteger(DEFAULT_SESSION_CAPACITY);
+
+    private final Topics configuration;
+
+    /**
+     * Constructor.
+     *
+     * @param configuration Configuration topic for this service
+     */
+    public SessionConfig(Topics configuration) {
+        this.configuration = configuration;
+        this.clientDeviceAuthSessionCapacity.set(getConfiguredSessionCapacity());
+
+        this.configuration.subscribe((whatHappened, node) -> {
+            // update session capacity to the latest configured value
+            updateSessionCapacity(getConfiguredSessionCapacity());
+            LOGGER.atInfo().kv("clientDeviceAuthSessionCapacity", clientDeviceAuthSessionCapacity.get())
+                    .log("Configuration updated");
+        });
+    }
+
+    /**
+     * Get configured Client-Device-Auth Session capacity.
+     *
+     * @return session capacity
+     */
+    public int getSessionCapacity() {
+        return clientDeviceAuthSessionCapacity.get();
+    }
+
+    /**
+     * Updates Client-Device-Auth Session capacity to the desired int value.
+     *
+     * @param newCapacity desired Client-Device-Auth Session capacity
+     */
+    public void updateSessionCapacity(int newCapacity) {
+        this.clientDeviceAuthSessionCapacity.set(newCapacity);
+    }
+
+    private int getConfiguredSessionCapacity() {
+        // valid session capacity should be within range [1, Integer.MAX_VALUE]
+        return getIntConfigValueRangeCheck(CLIENT_DEVICE_AUTH_SESSION_CAPACITY_TOPIC, DEFAULT_SESSION_CAPACITY,
+                1, Integer.MAX_VALUE);
+    }
+
+    /**
+     * Helper method that handles integer configuration parameters with a range check.
+     *
+     * @param configParameter   name of the configuration parameter
+     * @param defaultValue  default value
+     * @param low   low end of the range check
+     * @param high  high end of the range check
+     * @return returns the parsed value of the parameter.
+     *      If the value passed to the parameter is not a valid integer,
+     *      or within the specified range, the default value for the parameter will be returned.
+     */
+    private int getIntConfigValueRangeCheck(String configParameter, int defaultValue, int low, int high) {
+        if (configuration == null || configuration.isEmpty()) {
+            return defaultValue;
+        }
+        Object configValue = configuration.findOrDefault(defaultValue, configParameter);
+        int intValue = Coerce.toInt(configValue);
+        if (intValue < low || intValue > high) {
+            LOGGER.warn("Illegal value {} for configuration {}. Using default value {}",
+                    configValue, configParameter, defaultValue);
+            return defaultValue;
+        }
+        return intValue;
+    }
+}
--- a/src/main/java/com/aws/greengrass/device/session/SessionManager.java
+++ b/src/main/java/com/aws/greengrass/device/session/SessionManager.java
@@ -12,9 +12,10 @@ import com.aws.greengrass.logging.impl.LogManager;
 import lombok.AccessLevel;
 import lombok.Getter;
 
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Singleton class for managing AuthN and AuthZ sessions.
@@ -22,14 +23,22 @@ import java.util.concurrent.ConcurrentHashMap;
 public class SessionManager {
     private static final Logger logger = LogManager.getLogger(SessionManager.class);
     private static final String SESSION_ID = "SessionId";
+    // arbitrarily set default
+    private static final int DEFAULT_MAXIMUM_SESSION_COUNT = 100_000;
 
-    // TODO: Add mechanism to prevent session leaks
-    // This implementation currently relies on clients to properly close sessions. If they don't, then sessions
-    // will be leaked. We can add this once sessions are created with the entire set of device credentials. However,
+    // LRU Session Cache that evicts the eldest entry (based on access order) upon reaching its size.
+    // TODO: Support time-based cache eviction (Session timeout) and Session deduping.
+    // We can add this once sessions are created with the entire set of device credentials. However,
     // Moquette is currently creating sessions using just the certificate. Since multiple clients could use the same
     // certificate, we can't de-dup based on this alone.
     @Getter(AccessLevel.PACKAGE)
-    private final Map<String, Session> sessionMap = new ConcurrentHashMap<>();
+    private final Map<String, Session> sessionMap = Collections.synchronizedMap(
+            new LinkedHashMap<String, Session>(DEFAULT_MAXIMUM_SESSION_COUNT, 0.75f, true) {
+                @Override
+                protected boolean removeEldestEntry(Map.Entry<String, Session> eldest) {
+                    return size() > DEFAULT_MAXIMUM_SESSION_COUNT;
+                }
+            });
 
     /**
      * Looks up a session by id.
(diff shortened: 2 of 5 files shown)


src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @param newCapacity desired Client-Device-Auth Session capacity
+     */
+    public void updateSessionCapacity(int newCapacity) {
+        this.clientDeviceAuthSessionCapacity.set(newCapacity);
+    }
+
+    private int getConfiguredSessionCapacity() {
+        // valid session capacity should be within range [1, Integer.MAX_VALUE]
+        return getIntConfigValueRangeCheck(CLIENT_DEVICE_AUTH_SESSION_CAPACITY_TOPIC, DEFAULT_SESSION_CAPACITY,
```
jcosentino11, line 62: values greater than Integer.MAX_VALUE will overflow, so we can simplify the check to just greater than zero.

jcosentino11: also, not sure about the behavior of using default value if out of range.  if user attempts to set to zero, choosing 100,000 instead seems like a big difference.  maybe just using the minimum value would be better, similar to this: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/bc7c041d5fc6355e082d58c60b8e17a1616a305b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificatesConfig.java#L80-L85

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.util.Coerce;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class SessionConfig {
+    private static final Logger LOGGER = LogManager.getLogger(SessionConfig.class);
+    // arbitrarily set default
+    protected static final int DEFAULT_SESSION_CAPACITY = 100_000;
```
MikeDombo, line 18: this is quite large as a default IMO. I'd go for more like 1000

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+        return clientDeviceAuthSessionCapacity.get();
+    }
+
+    /**
+     * Updates Client-Device-Auth Session capacity to the desired int value.
+     *
+     * @param newCapacity desired Client-Device-Auth Session capacity
+     */
+    public void updateSessionCapacity(int newCapacity) {
+        this.clientDeviceAuthSessionCapacity.set(newCapacity);
```
jcosentino11, line 57:
```suggestion
        clientDeviceAuthSessionCapacity.set(newCapacity);
```
nit: for consistency w/ `getSessionCapacity`

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.util.Coerce;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class SessionConfig {
+    private static final Logger LOGGER = LogManager.getLogger(SessionConfig.class);
+    // arbitrarily set default
+    protected static final int DEFAULT_SESSION_CAPACITY = 100_000;
+    public static final String CLIENT_DEVICE_AUTH_SESSION_CAPACITY_TOPIC = "clientDeviceAuthSessionCapacity";
```
MikeDombo, line 19: remove cda from this string, we know it is cda because customers configure it under the cda key.

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -77,6 +87,15 @@ public void closeSession(String sessionId) {
         closeSessionInternal(sessionId);
     }
 
+    /**
+     * Updates Session Configuration.
+     *
+     * @param sessionConfig desired session configuration
+     */
+    public void updateSessionConfig(SessionConfig sessionConfig) {
```
MikeDombo, line 95: this isn't really updating, it is setting.

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public SessionConfig(Topics configuration) {
+        this.configuration = configuration;
+        this.clientDeviceAuthSessionCapacity.set(getConfiguredSessionCapacity());
+
+        this.configuration.subscribe((whatHappened, node) -> {
+            // update session capacity to the latest configured value
+            updateSessionCapacity(getConfiguredSessionCapacity());
+            LOGGER.atInfo().kv("clientDeviceAuthSessionCapacity", clientDeviceAuthSessionCapacity.get())
+                    .log("Configuration updated");
```
MikeDombo, line 38: remove useless log

src/test/java/com/aws/greengrass/device/session/SessionConfigTest.java
```diff
@@ -0,0 +1,80 @@
...
+                .withValue(overflown);
+        configurationTopics.context.waitForPublishQueueToClear();
+        assertThat(sessionConfig.getSessionCapacity(), is(equalTo(SessionConfig.DEFAULT_SESSION_CAPACITY)));
+
+        String empty = "";
+        configurationTopics.lookup(SessionConfig.CLIENT_DEVICE_AUTH_SESSION_CAPACITY_TOPIC).withValue(empty);
+        configurationTopics.context.waitForPublishQueueToClear();
+        assertThat(sessionConfig.getSessionCapacity(), is(equalTo(SessionConfig.DEFAULT_SESSION_CAPACITY)));
+    }
+}
```
MikeDombo, line 80: EOF

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+    public int getSessionCapacity() {
+        return clientDeviceAuthSessionCapacity.get();
+    }
+
+    /**
+     * Updates Client-Device-Auth Session capacity to the desired int value.
+     *
+     * @param newCapacity desired Client-Device-Auth Session capacity
+     */
+    public void updateSessionCapacity(int newCapacity) {
```
jcosentino11, line 56:
```suggestion
    private void updateSessionCapacity(int newCapacity) {
```
can reduce scope since it's only used in this class

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -23,13 +24,22 @@ public class SessionManager {
...
     // Moquette is currently creating sessions using just the certificate. Since multiple clients could use the same
     // certificate, we can't de-dup based on this alone.
     @Getter(AccessLevel.PACKAGE)
-    private final Map<String, Session> sessionMap = new ConcurrentHashMap<>();
+    private final Map<String, Session> sessionMap = Collections.synchronizedMap(
+            new LinkedHashMap<String, Session>(getSessionCapacity(), 0.75f, true) {
+                @Override
+                protected boolean removeEldestEntry(Map.Entry<String, Session> eldest) {
+                    // check size against latest configured session capacity
+                    return size() > getSessionCapacity();
```
jcosentino11, line 38: this will always be false if session capacity is `Integer.Max_VALUE`

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
vaibhavmurkute: updated

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
vaibhavmurkute: This makes sense. 1000 sounds like a good default (or _minimum_) to return for _all_ invalid values, for simplicity sake. Is this okay?
Also updated the valid capacity range to `[1, Integer.MAX_VALUE)`.

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
vaibhavmurkute: makes sense, updated.

src/main/java/com/aws/greengrass/device/session/SessionManager.java
vaibhavmurkute: updated

src/main/java/com/aws/greengrass/device/session/SessionManager.java
```diff
@@ -23,13 +24,22 @@ public class SessionManager {
...
+    // TODO: Support time-based cache eviction (Session timeout) and Session deduping.
+    // We can add this once sessions are created with the entire set of device credentials. However,
     // Moquette is currently creating sessions using just the certificate. Since multiple clients could use the same
     // certificate, we can't de-dup based on this alone.
     @Getter(AccessLevel.PACKAGE)
-    private final Map<String, Session> sessionMap = new ConcurrentHashMap<>();
+    private final Map<String, Session> sessionMap = Collections.synchronizedMap(
+            new LinkedHashMap<String, Session>(getSessionCapacity(), 0.75f, true) {
+                @Override
+                protected boolean removeEldestEntry(Map.Entry<String, Session> eldest) {
```
MikeDombo, line 36: Sessions will be getting automatically removed. Ensure we don't need any cleanup logic when sessions are dropped.

vaibhavmurkute:
> This isn't actually a LRU is it? It will remove based on insertion order and not by usage.

The LinkedHashMap here is configured to evict based on access-order (usage). Like this: [link](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html#:~:text=A%20special%20constructor,the%20backing%20map).

> Additionally, what is the big O for LinkedHashMap access, it is a linked list, so linear right?

It has a _doubly_-linkedlist. Quoting docs here:
> Like HashMap, it provides constant-time performance for the basic operations (add, contains and remove), assuming the hash function disperses elements properly among the buckets. Performance is likely to be just slightly below that of HashMap, due to the added expense of maintaining the linked list, with one exception: Iteration over the collection-views of a LinkedHashMap requires time proportional to the size of the map, regardless of its capacity. Iteration over a HashMap is likely to be more expensive, requiring time proportional to its capacity.

MikeDombo (dismissed):

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/device/session/SessionManager.java
MikeDombo: Good idea, yes. Maybe trace, just so we have it if absolutely needed

src/test/java/com/aws/greengrass/device/session/SessionConfigTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+class SessionConfigTest {
```
MikeDombo, line 22: extend with GGExtension

MikeDombo (approved):

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,88 @@
...
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.util.Coerce;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class SessionConfig {
+    private static final Logger LOGGER = LogManager.getLogger(SessionConfig.class);
+    // arbitrarily set default
+    protected static final int DEFAULT_SESSION_CAPACITY = 1000;
+    public static final String SESSION_CAPACITY_TOPIC = "sessionCapacity";
```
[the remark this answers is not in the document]
vaibhavmurkute: I agree. Addressed this in : https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/90. Thanks!

src/main/java/com/aws/greengrass/device/session/SessionConfig.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Helper method that handles integer configuration parameters with a range check.
+     *
+     * @param configParameter   name of the configuration parameter
+     * @param defaultValue  default value
+     * @param low   low end of the range check
+     * @param high  high end of the range check
+     * @return returns the parsed value of the parameter.
+     *      If the value passed to the parameter is not a valid integer,
+     *      or within the specified range, the default value for the parameter will be returned.
```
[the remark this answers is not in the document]
vaibhavmurkute: addressed in: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/90

PR #91, opened 2022-06-14
refactor: move IPC APIs to ClientDevicesAuthServiceAPI

jbutler:
**Issue #, if available:**

**Description of changes:**
Move IPC API functionality into ClientDevicesAuthServiceApi class so that it can also be used directly by internal clients.

This change also removes external dependence on the DeviceAuthClient, which will allow us to remove some of that dead code in a subsequent PR.

This change is needed in simplify Moquette changes needed to handle session eviction. Specifically, this enables Moquette to create sessions in one shot, rather than creating a session based on client certificate and then attaching thing information.

Unit / integ tests. Manual tests with updated Moquette.

Corresponding Moquette change: https://github.com/aws-greengrass/aws-greengrass-moquette-mqtt/pull/102

### DIFF
--- a/src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
@@ -17,6 +17,7 @@ import com.aws.greengrass.dependency.ImplementsService;
 import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.device.configuration.GroupConfiguration;
 import com.aws.greengrass.device.configuration.GroupManager;
+import com.aws.greengrass.device.exception.AuthenticationException;
 import com.aws.greengrass.device.exception.CloudServiceInteractionException;
 import com.aws.greengrass.device.iot.IotAuthClient;
 import com.aws.greengrass.device.session.MqttSessionFactory;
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -231,6 +223,29 @@ protected void shutdown() throws InterruptedException {
         certificateManager.stopMonitors();
     }
 
+    @Override
+    public void postInject() {
+        super.postInject();
+        try {
```
jbutler, line 229: TODO: We can consider migrating certificate APIs in a follow up PR, which will require a MQTT bridge CDA dependency bump as well

src/main/java/com/aws/greengrass/device/ClientDevicesAuthService.java
```diff
@@ -231,6 +223,29 @@ protected void shutdown() throws InterruptedException {
...
+            authorizationHandler.registerComponent(this.getName(),
+                    new HashSet<>(Collections.singletonList(SUBSCRIBE_TO_CERTIFICATE_UPDATES)));
```
MikeDombo, line 231: although this is currently useless, go ahead and register all the other operations too, not just subscribe to cert updates.

MikeDombo (approved):

MikeDombo (approved):

naikvidish (approved):

vaibhavmurkute (approved):

PR #93, opened 2022-06-15
feat: cache iot certificate IDs

vaibhavmurkute:
**Description of changes:** 
- Caches IoT Certificate IDs for certificatePem to avoid making same cloud requests for every MQTT connect.

**Why is this change necessary:** To avoid avoidable cloud requests

**How was this change tested:** Unit tests (`mvn clean package`)

### DIFF
--- a/src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.device;
+
+import com.aws.greengrass.device.exception.AuthenticationException;
+import com.aws.greengrass.device.exception.AuthorizationException;
+import com.aws.greengrass.device.iot.IotAuthClient;
+import com.aws.greengrass.device.session.SessionManager;
+
+import java.util.Map;
+import java.util.Optional;
+import javax.inject.Inject;
+
+public class ClientDevicesAuthServiceApi {
+    private final IotAuthClient iotAuthClient;
+    private final SessionManager sessionManager;
+    private final DeviceAuthClient deviceAuthClient;
+
+    /**
+     * Constructor.
+     *
+     * @param iotAuthClient    iot auth client
+     * @param sessionManager   session manager
+     * @param deviceAuthClient device auth client
+     */
+    @Inject
+    public ClientDevicesAuthServiceApi(IotAuthClient iotAuthClient,
+                                       SessionManager sessionManager,
+                                       DeviceAuthClient deviceAuthClient) {
+        this.iotAuthClient = iotAuthClient;
+        this.sessionManager = sessionManager;
+        this.deviceAuthClient = deviceAuthClient;
+    }
+
+    /**
+     * Verify client device identity.
+     * @param certificatePem PEM encoded client certificate.
+     * @return True if the provided client certificate is trusted.
+     */
+    public boolean verifyClientDeviceIdentity(String certificatePem) {
+        // Allow internal clients to verify their identities
+        if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
+            return true;
+        } else {
+            Optional<String> certificateId = iotAuthClient.getActiveCertificateId(certificatePem);
+            return certificateId.isPresent();
+        }
+    }
+
+    /**
+     * Get client auth token.
+     * @param credentialType    Type of client credentials
+     * @param deviceCredentials Client credential map
+     * @return client auth token to be used for future authorization requests.
+     * @throws AuthenticationException if unable to authenticate client credentials
+     */
+    public String getClientDeviceAuthToken(String credentialType, Map<String, String> deviceCredentials)
+        throws AuthenticationException {
+        return sessionManager.createSession(credentialType, deviceCredentials);
+    }
+
+    /**
+     * Close client auth session.
+     *
+     * <P>Note that closing auth sessions is strictly optional</P>
+     * @param authToken Auth token corresponding to the session to be closed.
+     */
+    public void closeClientDeviceAuthSession(String authToken) {
+        sessionManager.closeSession(authToken);
+    }
+
+    /**
+     * Authorize client action.
+     * @param authorizationRequest Authorization request, including auth token, operation, and resource
+     * @return true if the client action is allowed
+     * @throws AuthorizationException if the client action is not allowed
+     */
+    public boolean authorizeClientDeviceAction(AuthorizationRequest authorizationRequest)
+            throws AuthorizationException {
+        return deviceAuthClient.canDevicePerform(authorizationRequest);
+    }
+}
--- a/src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
@@ -7,7 +7,7 @@ package com.aws.greengrass.device;
 
 import com.aws.greengrass.device.exception.AuthenticationException;
 import com.aws.greengrass.device.exception.AuthorizationException;
-import com.aws.greengrass.device.iot.IotAuthClient;
+import com.aws.greengrass.device.iot.CertificateRegistry;
 import com.aws.greengrass.device.session.SessionManager;
 
 import java.util.Map;
@@ -15,22 +15,22 @@ import java.util.Optional;
 import javax.inject.Inject;
 
 public class ClientDevicesAuthServiceApi {
-    private final IotAuthClient iotAuthClient;
+    private final CertificateRegistry certificateRegistry;
     private final SessionManager sessionManager;
     private final DeviceAuthClient deviceAuthClient;
 
     /**
      * Constructor.
      *
-     * @param iotAuthClient    iot auth client
-     * @param sessionManager   session manager
-     * @param deviceAuthClient device auth client
+     * @param certificateRegistry iot auth client
+     * @param sessionManager      session manager
+     * @param deviceAuthClient    device auth client
      */
     @Inject
-    public ClientDevicesAuthServiceApi(IotAuthClient iotAuthClient,
+    public ClientDevicesAuthServiceApi(CertificateRegistry certificateRegistry,
                                        SessionManager sessionManager,
                                        DeviceAuthClient deviceAuthClient) {
-        this.iotAuthClient = iotAuthClient;
+        this.certificateRegistry = certificateRegistry;
         this.sessionManager = sessionManager;
         this.deviceAuthClient = deviceAuthClient;
     }
@@ -45,7 +45,7 @@ public class ClientDevicesAuthServiceApi {
         if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
             return true;
         } else {
-            Optional<String> certificateId = iotAuthClient.getActiveCertificateId(certificatePem);
+            Optional<String> certificateId = certificateRegistry.getIotCertificateIdForPem(certificatePem);
             return certificateId.isPresent();
         }
     }
--- a/src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
@@ -11,7 +11,6 @@ import com.aws.greengrass.device.iot.CertificateRegistry;
 import com.aws.greengrass.device.session.SessionManager;
 
 import java.util.Map;
-import java.util.Optional;
 import javax.inject.Inject;
 
 public class ClientDevicesAuthServiceApi {
@@ -45,8 +44,7 @@ public class ClientDevicesAuthServiceApi {
         if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
             return true;
         } else {
-            Optional<String> certificateId = certificateRegistry.getIotCertificateIdForPem(certificatePem);
-            return certificateId.isPresent();
+            return certificateRegistry.isCertificateValid(certificatePem);
         }
     }
 
--- a/src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
+++ b/src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
@@ -17,16 +17,19 @@ import java.util.Map;
 import java.util.Optional;
 import javax.inject.Inject;
 
-import static com.aws.greengrass.device.ClientDevicesAuthService.DEFAULT_MAX_ACTIVE_AUTH_TOKENS;
-
 
 public class CertificateRegistry {
     private static final Logger logger = LogManager.getLogger(CertificateRegistry.class);
-    public static final int REGISTRY_CACHE_SIZE = DEFAULT_MAX_ACTIVE_AUTH_TOKENS;
+    public static final int REGISTRY_CACHE_SIZE = 50;
     // holds mapping of certificateHash (SHA-256 hash of certificatePem) to IoT Certificate Id;
     // size-bound by default cache size, evicts oldest written entry if the max size is reached
     private static final Map<String, String> certificateHashToIdMap = Collections.synchronizedMap(
-            new LinkedHashMap<>(REGISTRY_CACHE_SIZE, 0.75f, false));
+            new LinkedHashMap<String, String>(REGISTRY_CACHE_SIZE, 0.75f, false) {
+                @Override
+                protected boolean removeEldestEntry(Map.Entry eldest) {
+                    return size() > REGISTRY_CACHE_SIZE;
+                }
+            });
 
     private final IotAuthClient iotAuthClient;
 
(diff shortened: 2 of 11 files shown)


src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -125,5 +135,75 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+                                + "policy grants the greengrass:VerifyClientDeviceIdentity permission.");
+                throw new CloudServiceInteractionException("Failed to verify client device identity", e);
+            }
+        }
+
+        private Optional<String> getCertificateIdAvailableLocally(String certificatePem) {
+            try {
+                String certHash = sha256Hash(certificatePem);
+                if (certificateHashToIdMap.containsKey(certHash)) {
+                    return Optional.of(certificateHashToIdMap.get(certHash));
```
MikeDombo, line 176: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.



**Problem**  
 You are using a `ConcurrentHashMap`, but your usage of `containsKey()` and `get()` may not be thread-safe at lines: **175 and 176**. In between the check and the `get()` another thread can remove the key and the `get()` will return `null`. The remove that can remove the key is at line: **141**.

**Fix**  
 Consider calling `get()`, checking instead of your current check if the returned object is `null`, and then using that object only, without calling `get()` again.

**More info**  
 [View an example on GitHub](https://github.com/apache/hadoop/blob/f16cf877e565084c66bc63605659b157c4394dc8/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/s3guard/S3Guard.java#L302-L304) (external link).

MikeDombo, line 176: optional.ofNullable(map.get

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -125,5 +135,75 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+                    && certificateIdToThingMap.get(certificateId).stream().anyMatch(thing -> thing.equals(thingName));
+        }
+
+        private synchronized void attachThingToCertificateLocally(String thingName, String certificateId) {
+            List<String> attachedThings = certificateIdToThingMap.getOrDefault(certificateId, new ArrayList<>());
+            attachedThings.add(thingName);
+            certificateIdToThingMap.put(certificateId, attachedThings);
+        }
+
+        private String sha256Hash(String text) throws NoSuchAlgorithmException {
```
MikeDombo, line 203: `Digest.calculate` from Nucleus

src/test/java/com/aws/greengrass/device/iot/IotAuthClientTest.java
```diff
@@ -35,9 +36,7 @@
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.lenient;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
MikeDombo, line 39: no * imports

jbutler:
This doesn't strike me as the right place to add caching (though we can possible consider it as a _short_ term stop gap). Not only will it make this class overly complex, but it also doesn't really cleanly fit with our future roadmap items (e.g. CA based auth, offline auth, etc).

The IPC layer probably isn't the right place, either. Any broker needs to be able to take advantage of whatever we build.

Have you considered adding a layer between `MqttSessionFactory` and `IotAuthClient`? We can mirror the cloud services here -- have a `CertificateRegistry` and `ThingRegistry` perhaps?

Another note:
The _**only**_ thing we want right now is a certificate PEM (or pem hash) -> IoT cert ID mapping to prevent there from being two calls to `VerifyClientDeviceIdentity` for every MQTT connect. Anything more will require security review / pen test and is explicitly out of scope for now.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -19,22 +19,36 @@
...
 
+    void clearLocalAuthCache();
+
     class Default implements IotAuthClient {
         private static final Logger logger = LogManager.getLogger(Default.class);
         private static final RetryUtils.RetryConfig SERVICE_EXCEPTION_RETRY_CONFIG =
                 RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofMillis(100)).maxAttempt(3)
                         .retryableExceptions(Arrays.asList(ThrottlingException.class, InternalServerException.class))
                         .build();
+        // TODO: clear local auth-cache as necessary
```
jbutler, line 48: Memory considerations aside, I can't think of a reason we'd ever need to clear PEM -> Cert ID mappings. IDs are immutable and the chances of two distinct certs hashing to the same ID are infinitesimally small.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -19,22 +19,36 @@
...
+        // certificateHash (SHA-256 hash of certificatePem) is stored instead of actual certificatePem
+        private static final Map<String, String> certificateHashToIdMap = new ConcurrentHashMap<>();
+        private static final Map<String, List<String>> certificateIdToThingMap = new ConcurrentHashMap<>();
```
jbutler, line 51: Remove. We're not doing this yet

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -125,5 +135,75 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+                return Optional.empty();
+            }
+            return Optional.empty();
+        }
+
+        private void attachCertificateIdToPem(String certificateId, String certificatePem) {
+            try {
+                certificateHashToIdMap.put(sha256Hash(certificatePem), certificateId);
+            } catch (NoSuchAlgorithmException e) {
+                logger.atTrace().cause(e).log("Could not store CertificatePem to CertificateId mapping");
```
jbutler, line 188: This is one of those "impossible" things that a customer would want to know about. Failure to save cert ID information means GG can't work offline. This should be at least warn. Maybe error

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -125,5 +135,75 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+            }
+        }
+
+        private synchronized boolean isThingAttachedToCertificateLocally(String thingName, String certificateId) {
```
jbutler, line 192: Same as above. Remove for now

jbutler (changes requested):

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: We consider a certificate "active" when IoT Core returns corresponding Certificate ID for its pem. When user manually _deactivates_ a certificate from the console, does IoT Core removes its CertID from its registry? Should this cached CertPem <-> CertId mapping be still valid then?

jbutler: CertID is just a hash computed from the certificate itself. Metadata about the certificate can change without the ID becoming invalid so keeping it is harmless.

It comes back to the point of this cache. It is NOT to prevent cloud calls if a device disconnects and reconnects. We are doing that later.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
```diff
@@ -55,29 +69,17 @@ public Optional<String> getActiveCertificateId(String certificatePem) {
...
-                return Optional.empty();
-            } catch (Exception e) {
-                logger.atError().cause(e).kv("certificatePem", certificatePem)
-                        .log("Failed to verify client device identity with cloud. Check that the core device's IoT "
-                                + "policy grants the greengrass:VerifyClientDeviceIdentity permission.");
-                throw new CloudServiceInteractionException("Failed to verify client device identity", e);
+            // cache active IoT Certificate Ids locally to avoid multiple cloud requests;
+            // mapping of certificateHash <-> IoT Certificate Id is cached;
+            // certificateHash is SHA-256 hash of certificatePem (to avoid storing actual certificatePem)
+            String certId =  getCertificateIdAvailableLocally(certificatePem)
```
jcosentino11, line 75: nit: since we return optional , can just keep optional here instead of converting back to optional at the end

```
Optional<String> certId = ...
certId.ifPresent(id -> attachCertificateIdToPem(id, certificatePem))
return certId;

```

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: addressed

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: PR updated, thanks.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: removed

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: log level set to `WARN`

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param certificateId IoT Certificate ID
+     * @param certificatePem Certificate PEM
+     */
+    private void registerCertificateIdForPem(String certificateId, String certificatePem) {
+        try {
+            certificateHashToIdMap.put(Digest.calculate(certificatePem), certificateId);
+        } catch (NoSuchAlgorithmException e) {
+            logger.atWarn().cause(e).log("Could not store CertificatePem to CertificateId mapping");
+        }
+    }
```
MikeDombo, line 107: move try/catch to a different method to handle the logging.

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+    private Optional<String> fetchActiveCertificateId(String certificatePem) {
+        return iotAuthClient.getActiveCertificateId(certificatePem);
+    }
+
+    private Optional<String> getAssociatedCertificateId(String certificatePem) {
+        try {
+            String certId = certificateHashToIdMap.get(Digest.calculate(certificatePem));
+            if (certId != null) {
+                return Optional.of(certId);
+            }
```
MikeDombo, line 88: this is what `Optional.ofNullable` is for. Use it

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
vaibhavmurkute: Do you mean extract hash-calculation to a separate method with this log? This log seems appropriate for this method. Why move it?

MikeDombo: Correct, move the log. SHA-256 is required by the Java spec, it cannot be missing. The log can be far more generic as this should never ever happen.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: Sorry couldn't find a way to chain `Optional` returns here. We can use `Optional<T>.or(() -> ...)` but that was added after JDK-9.

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
vaibhavmurkute: updated

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
```diff
@@ -80,16 +80,16 @@ private Optional<String> fetchActiveCertificateId(String certificatePem) {
...
-            if (certId != null) {
-                return Optional.of(certId);
-            }
-        } catch (NoSuchAlgorithmException e) {
-            return Optional.empty();
-        }
-        return Optional.empty();
+        String certId =  Optional.ofNullable(getCertificateHash(certificatePem))
+                .map(certificateHashToIdMap::get).orElse(null);
+        return Optional.ofNullable(certId);
```
MikeDombo, line 92: why do you need `orElse` here? Can't you just return from the `.map` call directly?

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+import java.util.Optional;
+import javax.inject.Inject;
+
+import static com.aws.greengrass.device.ClientDevicesAuthService.DEFAULT_MAX_ACTIVE_AUTH_TOKENS;
+
+
+public class CertificateRegistry {
+    private static final Logger logger = LogManager.getLogger(CertificateRegistry.class);
+    public static final int REGISTRY_CACHE_SIZE = DEFAULT_MAX_ACTIVE_AUTH_TOKENS;
```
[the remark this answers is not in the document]
jbutler: I actually disagree on this one. The cache only needs to hold on to the cert ID for a few milliseconds between TLS handshake and MQTT connect. Of course, we'll need config later. But the more important piece there will probably be related to disk usage and less about the size of an in memory cache.

For now we can keep this small (maybe 50?) and not expose config.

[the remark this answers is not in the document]
MikeDombo: I was thinking about that as well, if we only want to cut down on that one call, then having a far smaller cache is totally fine.

src/test/java/com/aws/greengrass/device/iot/CertificateRegistryTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @AfterEach
+    void afterEach() {
+        registry.clear();
+    }
+
+    @Test
+    void GIVEN_certificatePem_and_cloudProperResponse_WHEN_getIotCertificateIdForPem_THEN_certificateIdReturned() {
+        String mockCertPem = "certificatePem";
```
jbutler, line 50: Nit: can pull these out as constants since they're used in all of these tests

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+    // holds mapping of certificateHash (SHA-256 hash of certificatePem) to IoT Certificate Id;
+    // size-bound by default cache size, evicts oldest written entry if the max size is reached
+    private static final Map<String, String> certificateHashToIdMap = Collections.synchronizedMap(
+            new LinkedHashMap<>(REGISTRY_CACHE_SIZE, 0.75f, false));
```
MikeDombo, line 29: this creates a map with an initial capacity, but not a fixed capacity, right? How are we stopping this from growing forever?

MikeDombo: https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html#removeEldestEntry-java.util.Map.Entry-

Default is `false`, so it will grow forever.

MikeDombo: With a specific max size, we can just use a load factor of 1. It can allocate that size and never need to change it.

vaibhavmurkute: right, missed to add the size check for eviction. Updated now. Thanks.

src/test/java/com/aws/greengrass/device/iot/CertificateRegistryTest.java
vaibhavmurkute: updated

jbutler (approved):

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
vaibhavmurkute: right, updated return from `.map()` itself.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/device/iot/CertificateRegistry.java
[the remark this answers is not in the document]
vaibhavmurkute: updated cache size to a fixed smaller value (50) for now.

src/main/java/com/aws/greengrass/device/iot/IotAuthClient.java
vaibhavmurkute: makes sense now :) 
Updated.

src/main/java/com/aws/greengrass/device/ClientDevicesAuthServiceApi.java
```diff
@@ -45,7 +45,7 @@ public boolean verifyClientDeviceIdentity(String certificatePem) {
         if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
             return true;
         } else {
-            Optional<String> certificateId = iotAuthClient.getActiveCertificateId(certificatePem);
+            Optional<String> certificateId = certificateRegistry.getIotCertificateIdForPem(certificatePem);
```
jbutler, line 48: We may still want to call the auth client directly here. What does `isThingAttachedToCertificate` return if a deactivated certificate is attached to the Thing?

Test:
* Connect client device to local broker
* Deactivate client certificate
* Ensure we can not re-connect again

jbutler (changes requested):
Needs a follow up, DNM

MikeDombo (approved):

jbutler (approved):

PR #94, opened 2022-06-16
fix: cert gen fixes

jbutler:
**Issue #, if available:**

**Description of changes:**
Properly constructs the X500 Principal which is used as the Subject in generated certificates.

This change also removes some code that is no longer needed.

Manual testing to ensure OpenSSL can parse the subject in broker server certs

### DIFF
--- a/src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateRequestGenerator.java
+++ b/src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateRequestGenerator.java
@@ -52,9 +52,9 @@ public final class CertificateRequestGenerator {
      * @throws IOException fails to add SAN Extension to CSR builder
      */
     public static String createCSR(KeyPair keyPair,
-                                                  String thingName,
-                                                  List<InetAddress> ipAddresses,
-                                                  List<String> dnsNames) throws OperatorCreationException, IOException {
+                                   String thingName,
+                                   List<InetAddress> ipAddresses,
+                                   List<String> dnsNames) throws OperatorCreationException, IOException {
 
         // Create PKCS10 certificate request
         X500Principal x500Principal = getx500PrincipalForThing(thingName);
@@ -117,7 +117,7 @@ public final class CertificateRequestGenerator {
      * @return RDN sequence.
      */
     private static X500Principal getx500PrincipalForThing(String commonName) {
-        String distinguishedName = String.format("CN=%sC=%sST=%sL=%sO=%sOU=%s",
+        String distinguishedName = String.format("CN=%s,C=%s,ST=%s,L=%s,O=%s,OU=%s",
                 commonName,
                 CSR_COUNTRY,
                 CSR_PROVINCE,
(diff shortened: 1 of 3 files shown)


src/test/java/com/aws/greengrass/certificatemanager/certificate/CertificateHelperTest.java
```diff
@@ -36,7 +36,7 @@ public class CertificateHelperTest {
     private static final String EXPECTED_ISSUER_PRINCIPAL
             = "CN=Greengrass Core CA,L=Seattle,ST=Washington,OU=Amazon Web Services,O=Amazon.com Inc.,C=US";
     private static final String EXPECTED_SUBJECT_PRINCIPAL
-            = "CN=testCNC\\=USST\\=WashingtonL\\=SeattleO\\=Amazon.com Inc.OU\\=Amazon Web Services";
+            = "CN=testCN,C=US,ST=Washington,L=Seattle,O=Amazon.com Inc.,OU=Amazon Web Services";
```
MikeDombo, line 39: how could we have validated this before, using Java? Why are we only finding this now?

src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateRequestGenerator.java
```diff
@@ -52,21 +46,28 @@ private CertificateRequestGenerator() {
...
+     * @throws IOException fails to add SAN Extension to CSR builder
+     */
+    public static String createCSR(KeyPair keyPair, String thingName) throws OperatorCreationException, IOException {
         // Create PKCS10 certificate request
         X500Principal x500Principal = getx500PrincipalForThing(thingName);
         PKCS10CertificationRequestBuilder p10Builder = new JcaPKCS10CertificationRequestBuilder(
                 x500Principal, keyPair.getPublic());
 
-        // Add SAN cert extensions to request
-        ExtensionsGenerator extensions = getSANExtensions(ipAddresses, dnsNames);
```
MikeDombo, line 65: we don't need any SAN?

MikeDombo (approved):

src/test/java/com/aws/greengrass/certificatemanager/certificate/CertificateHelperTest.java
jbutler: Good question. I don't think anything really looks at this.

Theoretically we could add a test that parses the Subject and pulls out the individual fields. Without this change we just get a string with no semantic meaning.

src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateRequestGenerator.java
jbutler: The certificate generator overwrites this information anyway. This is leftover from early days before we had the connectivity information stuff ready.

https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/fd5003a0aa74af7f4abead417595d58302ce5f24/src/main/java/com/aws/greengrass/certificatemanager/certificate/CertificateHelper.java#L244

PR #95, opened 2022-06-17
feat: cert subscription api

jbutler:
**Issue #, if available:**

**Description of changes:**


src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -117,10 +119,62 @@ public List<String> getCACertificates() throws KeyStoreException, IOException, C
...
+            GetCertificateRequestOptions.CertificateType certificateType =
+                    getCertificateRequest.getCertificateRequestOptions().getCertificateType();
+            X509Certificate[] caCertificates = getX509CACertificates();
+            KeyPair keyPair = CertificateStore.newRSAKeyPair(2048);
+
+            if (certificateType.equals(GetCertificateRequestOptions.CertificateType.SERVER)) {
+                KeyPair finalKeyPair = keyPair;
+                Consumer<X509Certificate> consumer = (t) -> {
+                    GetCertificateUpdate getCertificateUpdate =
+                            new GetCertificateUpdate(finalKeyPair, t, caCertificates);
```
MikeDombo, line 154: What if the CAs change? Will we be able to take the necessary action?

src/main/java/com/aws/greengrass/device/api/GetCertificateUpdate.java
```diff
@@ -0,0 +1,19 @@
...
+package com.aws.greengrass.device.api;
+
+
+import lombok.Value;
+
+import java.security.KeyPair;
+import java.security.cert.X509Certificate;
+
+@Value
+public class GetCertificateUpdate {
```
MikeDombo, line 15: Certificate update or perhaps certificate update event?

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
jbutler: Good catch, that should be in the closure. This is kind of an interim change just to get us off the old APIs, so it will be simplified once we don't have the old consumer type

src/main/java/com/aws/greengrass/certificatemanager/CertificateManager.java
```diff
@@ -132,61 +186,52 @@ public String getCaPassPhrase() {
...
+
+        // Add certificate generator to monitors first in order to avoid missing events
+        // that happen while the initial certificate is being generated.
+        certExpiryMonitor.addToMonitor(certificateGenerator);
+        cisShadowMonitor.addToMonitor(certificateGenerator);
+
+        certificateGenerator.generateCertificate(connectivityInfoProvider::getCachedHostAddresses,
+                "initialization of server cert subscription");
+
+        serverCertSubscriptions.compute(serviceName, (k, v) -> {
```
MikeDombo, line 226: We didn't have this restriction before, right? I'd be a little worried to keep this. I'm not sure why a component would have more than one subscriber desired, but maybe.

jbutler: I thought about this as well. I don't intend to keep it this way, but this was the easiest way to achieve this without introducing a second map for tracking subscriptions. In the meantime I've added a TODO

src/main/java/com/aws/greengrass/device/api/GetCertificateUpdate.java
jbutler: Done

MikeDombo (approved):

jcosentino11 (approved):

PR #101, opened 2022-07-13
refactor: update and get certs based on usage and not within the subscribe

saranyailla:
**Issue #, if available:**

**Description of changes:**
1. Request reinstall when the configuration related to the CA changes. This will reinstall the components that depend on client device auth for CA. 
2. Upload CA out of the config subscription thread to avoid blocking as it includes re-tryable cloud calls. 
3. Update and get CA based on their usage by the caller and move it out of the config subscription logic. 

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -16,8 +16,10 @@ import com.aws.greengrass.clientdevices.auth.certificate.CertificateStore;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificatesConfig;
 import com.aws.greengrass.clientdevices.auth.certificate.ClientCertificateGenerator;
 import com.aws.greengrass.clientdevices.auth.certificate.ServerCertificateGenerator;
+import com.aws.greengrass.clientdevices.auth.exception.CertificateAuthorityNotFoundException;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
 import com.aws.greengrass.clientdevices.auth.iot.ConnectivityInfoProvider;
+import com.aws.greengrass.config.Topics;
 import lombok.NonNull;
 
 import java.io.IOException;
@@ -25,9 +27,11 @@ import java.security.KeyPair;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
+import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.X509Certificate;
 import java.time.Clock;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -70,14 +74,8 @@ public class CertificateManager {
         this.certificatesConfig = certificatesConfig;
     }
 
-    /**
-     * Initialize the certificate manager.
-     * @param caPassphrase  CA Passphrase
-     * @param caType        CA type
-     * @throws KeyStoreException if unable to load the CA key store
-     */
-    public void update(String caPassphrase, CertificateStore.CAType caType) throws KeyStoreException {
-        certificateStore.update(caPassphrase, caType);
+    public void setCertificateStoreConfig(Topics config, Topics runtimeConfig) {
+        certificateStore.setCertificateStoreConfig(config, runtimeConfig);
     }
 
     /**
@@ -103,17 +101,20 @@ public class CertificateManager {
      * @throws KeyStoreException if unable to retrieve the certificate
      * @throws IOException if unable to write certificate
      * @throws CertificateEncodingException if unable to get certificate encoding
+     * @throws CertificateAuthorityNotFoundException if unable to retrieve the CA
      */
-    public List<String> getCACertificates() throws KeyStoreException, IOException, CertificateEncodingException {
-        return Collections.singletonList(CertificateHelper.toPem(certificateStore.getCACertificate()));
+    public List<String> getCACertificates() throws KeyStoreException, IOException,
+            CertificateEncodingException, CertificateAuthorityNotFoundException {
+        List<String> certificatePems = new ArrayList<>();
+        Certificate[] certs = certificateStore.getCaCertificateChain();
+        for (Certificate cert : certs) {
+            certificatePems.add(CertificateHelper.toPem((X509Certificate) cert));
+        }
+        return certificatePems;
     }
 
-    private X509Certificate[] getX509CACertificates() throws KeyStoreException {
-        return new X509Certificate[]{certificateStore.getCACertificate()};
-    }
-
-    public String getCaPassPhrase() {
-        return certificateStore.getCaPassphrase();
+    private X509Certificate[] getX509CACertificates() throws KeyStoreException, CertificateAuthorityNotFoundException {
+        return certificateStore.getCaCertificateChain();
     }
 
     /**
@@ -129,6 +130,7 @@ public class CertificateManager {
      *
      * @param getCertificateRequest get certificate request
      * @throws CertificateGenerationException if unable to generate certificate
+     * @throws CertificateAuthorityNotFoundException if unable to retrieve the CA
      */
     public void subscribeToCertificateUpdates(GetCertificateRequest getCertificateRequest)
             throws CertificateGenerationException {
@@ -157,7 +159,7 @@ public class CertificateManager {
                 };
                 subscribeToClientCertificateUpdatesNoCSR(getCertificateRequest, keyPair.getPublic(), consumer);
             }
-        } catch (NoSuchAlgorithmException | KeyStoreException e) {
+        } catch (NoSuchAlgorithmException | KeyStoreException | CertificateAuthorityNotFoundException e) {
             throw new CertificateGenerationException(e);
         }
     }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -27,7 +27,6 @@ import java.security.KeyPair;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
-import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.X509Certificate;
 import java.time.Clock;
@@ -106,9 +105,9 @@ public class CertificateManager {
     public List<String> getCACertificates() throws KeyStoreException, IOException,
             CertificateEncodingException, CertificateAuthorityNotFoundException {
         List<String> certificatePems = new ArrayList<>();
-        Certificate[] certs = certificateStore.getCaCertificateChain();
-        for (Certificate cert : certs) {
-            certificatePems.add(CertificateHelper.toPem((X509Certificate) cert));
+        X509Certificate[] certs = certificateStore.getCaCertificateChain();
+        for (X509Certificate cert : certs) {
+            certificatePems.add(CertificateHelper.toPem(cert));
         }
         return certificatePems;
     }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
@@ -7,10 +7,10 @@ package com.aws.greengrass.clientdevices.auth;
 
 import com.aws.greengrass.authorization.AuthorizationHandler;
 import com.aws.greengrass.clientdevices.auth.api.ClientDevicesAuthServiceApi;
-import com.aws.greengrass.clientdevices.auth.certificate.CertificateStore;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificatesConfig;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupConfiguration;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
+import com.aws.greengrass.clientdevices.auth.exception.CertificateAuthorityNotFoundException;
 import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
 import com.aws.greengrass.clientdevices.auth.session.MqttSessionFactory;
 import com.aws.greengrass.clientdevices.auth.session.SessionConfig;
@@ -30,11 +30,10 @@ import com.aws.greengrass.lifecyclemanager.PluginService;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.GreengrassServiceClientFactory;
 import com.aws.greengrass.util.RetryUtils;
-import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService;
-import software.amazon.awssdk.services.greengrassv2data.model.AccessDeniedException;
+import software.amazon.awssdk.services.greengrassv2.model.AccessDeniedException;
 import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
 import software.amazon.awssdk.services.greengrassv2data.model.PutCertificateAuthoritiesRequest;
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
@@ -63,6 +62,7 @@ import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.VER
 public class ClientDevicesAuthService extends PluginService {
     public static final String CLIENT_DEVICES_AUTH_SERVICE_NAME = "aws.greengrass.clientdevices.Auth";
     public static final String DEVICE_GROUPS_TOPICS = "deviceGroups";
+    public static final String CERTIFICATE_AUTHORITY_TOPIC = "certificateAuthority";
     public static final String CA_TYPE_TOPIC = "ca_type";
     public static final String CA_PASSPHRASE = "ca_passphrase";
     public static final String CERTIFICATES_KEY = "certificates";
@@ -83,7 +83,6 @@ public class ClientDevicesAuthService extends PluginService {
     private final CertificateManager certificateManager;
 
     private final GreengrassServiceClientFactory clientFactory;
-
     private final ClientDevicesAuthServiceApi clientDevicesAuthServiceApi;
     private final DeviceConfiguration deviceConfiguration;
     private final AuthorizationHandler authorizationHandler;
@@ -137,6 +136,7 @@ public class ClientDevicesAuthService extends PluginService {
         SessionCreator.registerSessionFactory("mqtt", mqttSessionFactory);
         certificateManager.updateCertificatesConfiguration(new CertificatesConfig(this.getConfig()));
         sessionManager.setSessionConfig(new SessionConfig(this.getConfig()));
+        certificateManager.setCertificateStoreConfig(this.getConfig(), this.getRuntimeConfig());
     }
 
     private int getValidCloudCallQueueSize(Topics topics) {
@@ -161,7 +161,8 @@ public class ClientDevicesAuthService extends PluginService {
      * |    |---- deviceGroups:
      * |         |---- definitions : {}
      * |         |---- policies : {}
-     * |    |---- ca_type: [...]
+     * |    |---- certificateAuthority:
+     * |        |---- ca_type: "..."
      * |    |---- certificates: {}
      * |---- runtime
      * |    |---- ca_passphrase: "..."
@@ -171,7 +172,19 @@ public class ClientDevicesAuthService extends PluginService {
     @Override
     protected void install() throws InterruptedException {
         super.install();
+
+        // Before subscribing to the config updates, look up for CA type topic. This will create a node with default
+        // value as null if the topic doesn't exist already.
+        getContext().runOnPublishQueueAndWait(() -> {
+            this.config.lookup(CONFIGURATION_CONFIG_KEY, CERTIFICATE_AUTHORITY_TOPIC, CA_TYPE_TOPIC);
+        });
+        getContext().waitForPublishQueueToClear();
+
+        // Subscribe to config updates
         configChangeHandler();
+
+        // Update CA based on the config.
+        updateCAType();
     }
 
     private void configChangeHandler() {
@@ -181,7 +194,6 @@ public class ClientDevicesAuthService extends PluginService {
             }
             logger.atDebug().kv("why", whatHappened).kv("node", node).log();
             Topics deviceGroupTopics = this.config.lookupTopics(CONFIGURATION_CONFIG_KEY, DEVICE_GROUPS_TOPICS);
-            Topic caTypeTopic = this.config.lookup(CONFIGURATION_CONFIG_KEY, CA_TYPE_TOPIC);
 
             // Attempt to update the thread pool size as needed
             try {
@@ -205,14 +217,12 @@ public class ClientDevicesAuthService extends PluginService {
 
             if (whatHappened == WhatHappened.initialized || node == null) {
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
-                updateCAType(caTypeTopic);
             } else if (node.childOf(DEVICE_GROUPS_TOPICS)) {
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
-            } else if (node.childOf(CA_TYPE_TOPIC)) {
-                if (caTypeTopic.getOnce() == null) {
-                    return;
-                }
-                updateCAType(caTypeTopic);
+            } else if (node.childOf(CERTIFICATE_AUTHORITY_TOPIC)) {
+                logger.atInfo("service-config-change").kv("event", whatHappened).kv("node", node)
+                        .log("Requesting re-installation of client device auth component");
+                requestReinstall();
             }
         });
     }
@@ -234,12 +244,10 @@ public class ClientDevicesAuthService extends PluginService {
         super.postInject();
         try {
             authorizationHandler.registerComponent(this.getName(),
-                    new HashSet<>(Arrays.asList(new String[]{
-                            SUBSCRIBE_TO_CERTIFICATE_UPDATES,
+                    new HashSet<>(Arrays.asList(SUBSCRIBE_TO_CERTIFICATE_UPDATES,
                             VERIFY_CLIENT_DEVICE_IDENTITY,
                             GET_CLIENT_DEVICE_AUTH_TOKEN,
-                            AUTHORIZE_CLIENT_DEVICE_ACTION
-                    })));
+                            AUTHORIZE_CLIENT_DEVICE_ACTION)));
         } catch (com.aws.greengrass.authorization.exceptions.AuthorizationException e) {
             logger.atError("initialize-cda-service-authorization-error", e)
                     .log("Failed to initialize the client device auth service with the Authorization module.");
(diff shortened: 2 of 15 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +215,51 @@ protected void install() throws InterruptedException {
...
+                String keyAlias = securityService.getKeyAlias(privateKeyUri, certificateUri);
+                PrivateKey privateKey = ((X509KeyManager) keyManagers[0]).getPrivateKey(keyAlias);
+                X509Certificate[] certificateChain =
+                        ((X509KeyManager) keyManagers[0]).getCertificateChain(keyAlias);
+                certificateManager.setCA(privateKey, certificateChain);
+                List<String> caCerts = certificateManager.getCACertificates();
+                uploadCoreDeviceCAs(caCerts);
+                updateCACertificateConfig(caCerts);
+            } catch (ServiceUnavailableException | KeyLoadingException | KeyStoreException | IOException | CertificateEncodingException e) {
+                e.printStackTrace();
```
MikeDombo, line 258: handle properly

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +215,51 @@ protected void install() throws InterruptedException {
...
                 }
-                updateCAType(caTypeTopic);
+                updateCertificateAuthority();
             }
         });
 
     }
 
+    private void updateCertificateAuthority() {
+        String caKeyPath = Coerce.toString(config.lookup(CERTIFICATE_AUTHORITY_TOPIC, CA_PRIVATE_KEY_PATH).dflt(""));
```
MikeDombo, line 232: I don't think that using `dflt` here is appropriate. You can use `findOrDefault`, or just handle the null later on

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +215,51 @@ protected void install() throws InterruptedException {
...
+                Coerce.toString(config.lookup(CERTIFICATE_AUTHORITY_TOPIC, CA_CERTIFICATE_FILE_PATH).dflt(""));
+        Topic caType = config.lookup(CERTIFICATE_AUTHORITY_TOPIC, CA_TYPE_TOPIC);
+
+        // If the key and cert path of CA are empty, generate a CA
+        if (Utils.isEmpty(caKeyPath) || Utils.isEmpty(caCertPath)) {
+            updateCAType(caType);
+        } else {
+            // Get CA from URI and upload
+            URI privateKeyUri = SecurityService.uriFromPossibleFileURIString(caKeyPath);
+            URI certificateUri = SecurityService.uriFromPossibleFileURIString(caCertPath);
```
MikeDombo, line 243: since these are new, we can be strict here. We can ensure that these are absolutely URIs and not just file paths. We couldn't do that in the Nucleus for compatibility, but that's not a concern here. Recommend trying to parse as a URI and then handling failure appropriately.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +215,51 @@ protected void install() throws InterruptedException {
...
+            try {
+                KeyManager[] keyManagers =
+                        this.getContext().get(SecurityService.class).getKeyManagers(privateKeyUri, certificateUri);
```
MikeDombo, line 246: inject security service, rather than looking it up in the context?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -105,6 +107,13 @@ public void update(String passphrase, CAType caType) throws KeyStoreException {
             createAndStoreDefaultKeyStore(caType);
             logger.atDebug().log("successfully created new CA keystore");
         }
+
+        try {
+            caCertChain = keyStore.getCertificateChain(CA_KEY_ALIAS);
+            caPrivateKey = (PrivateKey) keyStore.getKey(CA_KEY_ALIAS, getPassphrase());
```
MikeDombo, line 113: don't blindly cast, check first

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -62,6 +70,9 @@
 public class ClientDevicesAuthService extends PluginService {
     public static final String CLIENT_DEVICES_AUTH_SERVICE_NAME = "aws.greengrass.clientdevices.Auth";
     public static final String DEVICE_GROUPS_TOPICS = "deviceGroups";
+    public static final String CERTIFICATE_AUTHORITY_TOPIC = "certificateAuthority";
+    public static final String CA_PRIVATE_KEY_PATH = "privateKeyPath";
+    public static final String CA_CERTIFICATE_FILE_PATH = "certificateFilePath";
```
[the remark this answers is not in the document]
saranyailla: I took this from system configuration we've for device keys and cert.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +215,51 @@ protected void install() throws InterruptedException {
...
+                        this.getContext().get(SecurityService.class).getKeyManagers(privateKeyUri, certificateUri);
+                // Get alias from the key URI.
+                SecurityService securityService = this.getContext().get(SecurityService.class);
+                String keyAlias = securityService.getKeyAlias(privateKeyUri, certificateUri);
+                PrivateKey privateKey = ((X509KeyManager) keyManagers[0]).getPrivateKey(keyAlias);
+                X509Certificate[] certificateChain =
+                        ((X509KeyManager) keyManagers[0]).getCertificateChain(keyAlias);
+                certificateManager.setCA(privateKey, certificateChain);
+                List<String> caCerts = certificateManager.getCACertificates();
+                uploadCoreDeviceCAs(caCerts);
```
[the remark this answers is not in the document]
saranyailla: This call is made inside of `updateCAType(topic)` https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/e00b8f127685b3940a7339bd51f7308196bfa833/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java#L337-L338

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +215,51 @@ protected void install() throws InterruptedException {
...
+        if (Utils.isEmpty(caKeyPath) || Utils.isEmpty(caCertPath)) {
+            updateCAType(caType);
+        } else {
+            // Get CA from URI and upload
+            URI privateKeyUri = SecurityService.uriFromPossibleFileURIString(caKeyPath);
+            URI certificateUri = SecurityService.uriFromPossibleFileURIString(caCertPath);
+            try {
+                KeyManager[] keyManagers =
+                        this.getContext().get(SecurityService.class).getKeyManagers(privateKeyUri, certificateUri);
+                // Get alias from the key URI.
```
MikeDombo, line 247: we shouldn't need the alias for any reason. We should be able to get this information, at worst, by iterating through the possible aliases from `getServerAliases` or `getClientAliases`.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
saranyailla: updated

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +220,54 @@ protected void install() throws InterruptedException {
...
+            // Get CA from URI and upload
+            try {
+                URI privateKeyUri = new URI(caKeyPath);
+                URI certificateUri = new URI(caCertPath);
+                KeyPair keyPair = securityService.getKeyPair(privateKeyUri, certificateUri);
+                List<X509Certificate> certificateChain = securityService.getCertificateChain(privateKeyUri,
+                        certificateUri);
+                certificateManager.setCA(keyPair.getPrivate(), (Certificate[]) certificateChain.toArray());
+            } catch (URISyntaxException e) {
+                throw new ServiceError("Exception occurred while parsing the certificate authority "
```
MikeDombo, line 255: always use `serviceErrored` and include the full context of the error. You're throwing away the actual causes

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +220,54 @@ protected void install() throws InterruptedException {
...
+                        + "configuration");
+            } catch (ServiceUnavailableException | KeyLoadingException | IOException | CertificateException e) {
```
MikeDombo, line 257: ServiceUnavailableException must be retried, you cannot just bail out. It is expected to be thrown for a while, before the pkcs11 provider is loaded.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +220,54 @@ protected void install() throws InterruptedException {
...
+                throw new ServiceError("Exception occurred while getting the certificate authority");
+            }
+            try {
+                List<String> caCerts = certificateManager.getCACertificates();
+                uploadCoreDeviceCAs(caCerts);
+                updateCACertificateConfig(caCerts);
+            } catch (KeyStoreException | IOException | CertificateEncodingException e) {
+                logger.atError().cause(e).log("Failed to get the CA certificates");
+                throw new ServiceError("Exception occurred while getting the CA certificates");
```
MikeDombo, line 266: `serviceErrored`

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -285,12 +340,8 @@ private void updateCAType(Topic topic) {
                 String caType = caTypeList.get(0);
                 certificateManager.update(getPassphrase(), CertificateStore.CAType.valueOf(caType));
             }
-
-            List<String> caCerts = certificateManager.getCACertificates();
```
MikeDombo, line 289: why aren't we doing this during startup anymore?

saranyailla: Updated.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,19 +226,67 @@ protected void install() throws InterruptedException {
...
+                        () -> securityService.getCertificateChain(privateKeyUri, certificateUri),
+                        "get-certificate-chain", logger);
+                certificateManager.setCA(keyPair.getPrivate(), (Certificate[]) certificateChain.toArray());
+            } catch (URISyntaxException e) {
+                logger.atError().kv(CA_PRIVATE_KEY_URI, "")
+                        .kv(CA_CERTIFICATE_URI, "")
+                        .setCause(e)
+                        .log("Exception occurred while parsing the certificate authority configuration");
+                serviceErrored(e);
+            } catch (InterruptedException | KeyLoadingException | IOException | CertificateException e) {
```
MikeDombo, line 270: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -198,21 +219,81 @@ protected void install() throws InterruptedException {
...
+            uploadCoreDeviceCAs(caCerts);
+            updateCACertificateConfig(caCerts);
+        } catch (KeyStoreException | IOException | CertificateEncodingException e) {
+            logger.atError().cause(e).log("Failed to get the CA certificates");
+            serviceErrored(e);
+        }
+    }
+
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private void updateCertificateManager(URI privateKeyUri, URI certificateUri) {
```
MikeDombo, line 278: what thread is running this?

If this is running in the main context thread, this will not work. The service unavailable exception is the only exception which should be retried.

The service will only become available when it registers itself, and to do that it will be constructed by the context thread. So make sure we are not in the context thread, otherwise this is guaranteed to fail.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -105,30 +111,53 @@ public void update(String passphrase, CAType caType) throws KeyStoreException {
...
+     * Get CA private key.
+     * This getter method checks if the private key is null before returning it to avoid null pointer exceptions.
+     *
+     * @return caCertificateChain
+     * @throws ServiceError service error
+     **/
+    public PrivateKey getCaPrivateKey() {
+        if (caPrivateKey == null) {
+            throw new ServiceError("Unable to find the CA private key");
+        }
```
MikeDombo, line 159: why are you throwing? We don't like throwing.

src/test/resources/com/aws/greengrass/clientdevices/auth/ca_cert.pem.crt
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
MikeDombo, line 1: don't commit keys and certs. Even though these don't go to anything, appsec won't be happy. 

Generate the keys and certs during the test.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -105,30 +111,53 @@ public void update(String passphrase, CAType caType) throws KeyStoreException {
...
+     * This getter method checks if the certificate chain is null before returning it to avoid null pointer exceptions.
+     *
+     * @return caCertificateChain
+     * @throws ServiceError service error
+     **/
+    public Certificate[] getCaCertificateChain() {
+        if (caCertificateChain == null) {
+            throw new ServiceError("Unable to find the CA certificate chain");
+        }
+        return caCertificateChain.clone();
```
MikeDombo, line 146: why cloning?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
MikeDombo: If you want this to be nullable, then use `Optional<PrivateKey>`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
saranyailla: To avoid MethodReturnsInternalArray PMD violation. This one references the same array though.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -198,21 +227,89 @@ protected void install() throws InterruptedException {
...
+                uploadCoreDeviceCAs(caCerts);
+                updateCACertificateConfig(caCerts);
+            } catch (KeyStoreException | IOException | CertificateEncodingException e) {
+                logger.atError().cause(e).log("Failed to update the CA certificates");
+                serviceErrored(e);
+            }
+        });
+
+        try {
+            fut.get(10, TimeUnit.SECONDS);
```
MikeDombo, line 286: you're blocking the main context thread. Do not block this thread, ever

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -15,8 +15,10 @@
 import com.aws.greengrass.util.platforms.Platform;
 import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.Setter;
 import org.bouncycastle.cert.CertIOException;
 import org.bouncycastle.operator.OperatorCreationException;
+import software.amazon.awssdk.aws.greengrass.model.ServiceError;
```
MikeDombo, line 21: do not use `ServiceError`, this is an IPC error

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -285,12 +382,8 @@ private void updateCAType(Topic topic) {
                 String caType = caTypeList.get(0);
                 certificateManager.update(getPassphrase(), CertificateStore.CAType.valueOf(caType));
             }
-
-            List<String> caCerts = certificateManager.getCACertificates();
-            uploadCoreDeviceCAs(caCerts);
-            updateCACertificateConfig(caCerts);
             updateCaPassphraseConfig(certificateManager.getCaPassPhrase());
-        } catch (KeyStoreException | IOException | CertificateEncodingException | IllegalArgumentException
+        } catch (KeyStoreException | IllegalArgumentException
```
jbutler, line 386: There's an existing bug here - if uploading the CA fails, we shouldn't enter a broken state. Should think about how to handle retries

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -279,7 +303,7 @@ private void saveKeyStore() throws IOException, CertificateException,
 
         // TODO: Clean this up
         // Temporarily store public CA since CA information is not yet available in cloud
-        X509Certificate caCert = getCACertificate();
+        Certificate caCert = keyStore.getCertificate(CA_KEY_ALIAS);
```
jbutler, line 306: Nit: while you're here, remove the old comment and write the full cert chain

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateGenerator.java
```diff
@@ -60,12 +60,17 @@ public synchronized void generateCertificate(Supplier<List<String>> connectivity
             return;
         }
 
+        if (certificateStore.getCaPrivateKey() == null) {
```
jbutler, line 63: When would this be null?

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -80,6 +84,11 @@ public void update(String caPassphrase, CertificateStore.CAType caType) throws K
         certificateStore.update(caPassphrase, caType);
     }
 
+    public void setCA(PrivateKey privateKey, X509Certificate... certificates) {
+        certificateStore.setCaCertificateChain(certificates);
+        certificateStore.setCaPrivateKey(privateKey);
```
jbutler, line 89: This doesn't seem like the correct interface. You shouldn't be able to update these independently or else you can end up in an inconsistent state

pom.xml
```diff
@@ -73,13 +73,13 @@
...
             <artifactId>nucleus</artifactId>
-            <version>2.6.0-SNAPSHOT</version>
+            <version>2.7.0-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.6.0-SNAPSHOT</version>
+            <version>2.7.0-SNAPSHOT</version>
```
jbutler, line 82: This is breaking backward compatibility. Need to update the PR title (`feat!: ...`)

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -105,30 +110,49 @@ public void update(String passphrase, CAType caType) throws KeyStoreException {
...
+    public Certificate[] getCaCertificateChain() {
+        if (caCertificateChain == null) {
+            return new Certificate[0];
+        }
+        return caCertificateChain;
+    }
+
+    /**
+     * Get CA private key.
+     * This getter method checks if the private key is null before returning it to avoid null pointer exceptions.
```
jbutler, line 150: This statement concerns me. This should never happen as long as the component is in the `RUNNING` state

saranyailla: That comment has to be removed.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateGenerator.java
saranyailla: As you commented out earlier, this should not be null if the component is in `RUNNING` state. But because we're using `serviceErrored(e)` to put the component in BROKEN state, it takes some amount of time for the Nucleus to identify that this component is BROKEN  and retry. But, in this time, CDA continues to execute next steps and sometimes resulting in null pointer exceptions.


I've seen this happening here while trying to get certificates after the service errored https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/ff646044c87875f465dba5cccd0b8f5b54132ad1/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java#L272-L276 but I added it for both cert  and key.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
saranyailla: We're retrying the cloud request here - https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/ff646044c87875f465dba5cccd0b8f5b54132ad1/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java#L423
Did you mean the same? Also, those exceptions are not from `uploadCoreDeviceCAs`.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
saranyailla: Updated.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -198,21 +203,47 @@ protected void install() throws InterruptedException {
...
             if (whatHappened == WhatHappened.initialized || node == null) {
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
-                updateCAType(caTypeTopic);
+                updateCertificateAuthority();
             } else if (node.childOf(DEVICE_GROUPS_TOPICS)) {
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
-            } else if (node.childOf(CA_TYPE_TOPIC)) {
-                if (caTypeTopic.getOnce() == null) {
+            } else if (node.childOf(CERTIFICATE_AUTHORITY_TOPIC)) {
+                // If the CA type is null because of the change to its configuration to null, do not update the CA.
```
jbutler, line 219: I'd argue this is wrong. A customer should be able to know the type of CA by looking at the configuration, without needing to be aware of previous state. Make this deterministic.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -198,21 +203,47 @@ protected void install() throws InterruptedException {
...
+        // Upload the generated or provided CA certificates to the GG cloud and update config
+        try {
+            List<String> caCerts = certificateManager.getCACertificates();
+            if (caCerts.isEmpty()) {
+                serviceErrored(new Exception("Unable to get the CA certificates"));
+            }
+            uploadCoreDeviceCAs(caCerts);
+            updateCACertificateConfig(caCerts);
+        } catch (KeyStoreException | IOException | CertificateEncodingException e) {
+            serviceErrored(e);
```
jbutler, line 243: It's not clear from this code what exceptions you're catching and what you're not. This should be refactored to improve clarity. For example, remove `uploadCoreDeviceCAs` from the try block since you're not catching the exception it throws. 
Clean up the logic such that we retry but continue. The customer may not be relying on cloud discovery at all, so IMO we should continue running.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/ServerCertificateGenerator.java
```diff
@@ -60,6 +60,11 @@ public synchronized void generateCertificate(Supplier<List<String>> connectivity
             return;
         }
 
+        if (certificateStore.getCaPrivateKey() == null) {
+            logger.atWarn().log("Not generating new server certificate as CA private key is not found");
+            return;
+        }
```
jbutler, line 66: This makes me super uncomfortable. We **_CANNOT_** silently fail like this. An error needs to be bubbled up so some action can be taken. Also consider increasing log level to error.
Can fix this as part of a separate PR.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -170,13 +172,28 @@ private int getValidCloudCallQueueSize(Topics topics) {
...
+        getContext().waitForPublishQueueToClear();
+
+        // Subscribe to config updates
+        configChangeHandler();
+
+        // Update CA based on the config.
+        updateCAType();
+    }
+
+    private void configChangeHandler() {
```
MikeDombo, line 190: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 13. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test. We recommend that you simplify this method or break it into multiple methods.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -103,17 +101,20 @@ public void stopMonitors() {
...
-    }
-
-    private X509Certificate[] getX509CACertificates() throws KeyStoreException {
-        return new X509Certificate[]{certificateStore.getCACertificate()};
+    public List<String> getCACertificates() throws KeyStoreException, IOException,
+            CertificateEncodingException, CertificateAuthorityNotFoundException {
+        List<String> certificatePems = new ArrayList<>();
+        Certificate[] certs = certificateStore.getCaCertificateChain();
+        for (Certificate cert : certs) {
+            certificatePems.add(CertificateHelper.toPem((X509Certificate) cert));
```
MikeDombo, line 111: blindly casting? Check instanceof

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -200,17 +217,14 @@ protected void install() throws InterruptedException {
...
                 updateDeviceGroups(whatHappened, deviceGroupTopics);
-            } else if (node.childOf(CA_TYPE_TOPIC)) {
-                if (caTypeTopic.getOnce() == null) {
-                    return;
-                }
-                updateCAType(caTypeTopic);
+            } else if (node.childOf(CERTIFICATE_AUTHORITY_TOPIC)) {
+                logger.atInfo("service-config-change").kv("event", whatHappened).kv("node", node)
+                        .log("Requesting re-installation of client device auth component");
+                requestReinstall();
```
MikeDombo, line 225: reinstall will force any dependencies of CDA to stop. Make sure this is what we intend.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -270,28 +282,15 @@ private void updateDeviceGroups(WhatHappened whatHappened, Topics deviceGroupsTo
...
-            }
-
             List<String> caCerts = certificateManager.getCACertificates();
             uploadCoreDeviceCAs(caCerts);
             updateCACertificateConfig(caCerts);
-            updateCaPassphraseConfig(certificateManager.getCaPassPhrase());
-        } catch (KeyStoreException | IOException | CertificateEncodingException | IllegalArgumentException
-                | CloudServiceInteractionException e) {
+        } catch (CloudServiceInteractionException e) {
+            logger.atError().cause(e).log("Unable to upload core CA certs to the cloud. ");
```
MikeDombo, line 291: no period

saranyailla: Addressed in https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/105

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
saranyailla: I think that's what we want too because the dependency component is re-installed as well.

MikeDombo: Dependencies will not be reinstalled.

And why is that what we want? We do not want the MQTT broker to be taken down unnecessarily.

saranyailla: We reinstall CDA only when CA related config changes. So, any component that has hard dependency on CDA should be restarted to use the new CA.

MikeDombo: Why should they _restart_ though? The CA is sent in a subscription based API so that dependencies don't need to restart.

saranyailla: Oh right, that works too. Will have to update the dependency components to subscribe to these changes then if they're not already.

PR #102, opened 2022-08-08
feat: enable offline auth with caching

vaibhavmurkute:
**Description of changes:** Enables offline auth with caching. AuthN/AuthZ workflows consult cache when cloud is unreachable.

Notes:
- TTL for cache entries and the periodic cache refresh frequency are set to one day. These parameters will be made configurable.
- Cache-refresh here only removes stale entries, but it is envisioned to sync cache results with cloud also. Cache interface will also support an API to persist cache locally on the filesystem to preserve cache across device power cycles.
- Default cache size is changed from 50 to 2500 entries for now.

Outline:
```
                            interface                           +----------------------------+
                 +-----------------------+                    |                            |
                 |  RefreshableRegistry  |                    | RefreshableRegistryEntry   |
                 |                       |                    |                            |
           +-----+  - refresh()          | -----+             |                            |
           |     |                       |      |             |                            |
           |     |                       |      |             |   - isValid()              |
           |     +-----------------------+      |             |   - getValidTill()         |
           |                                    |             |   - setValidTill()         |
           |                                    |             |                            |
           |              implements            |             |                            |
           v                                    v             +----------------------------+
+-----------------------+           +-------------------+
|                       |           |                   |
|  CertificateRegistry  |           |   ThingRegistry   |
|                       |           |                   |
|                       |           |                   |
+-----------------------+           +-------------------+





                                +------------------------------------------------+
                                |   RegistryRefreshMonitor                       |
                                |                                                |
                                |   - Set<RefreshableRegistry> monitoredRegistry;|
                                |                                                |
                                |   - addToMonitor(RefreshableRegistry)          |
                                |   - startMonitor()                             |
                                |   - stopMonitor()                              |
                                |                                                |
                                +------------------------------------------------+
```

TO-DOs (in a separate PR) :
- periodically sync cache results with cloud
- persist cache on the filesystem
- make cache config parameters customer-configurable.

**Why is this change necessary:** a part of BYOCA.

**How was this change tested:** unit tests, `mvn verify`

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.clientdevices.auth.certificate.CertificatesConfig;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupConfiguration;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
 import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
+import com.aws.greengrass.clientdevices.auth.iot.registry.RegistryRefreshMonitor;
 import com.aws.greengrass.clientdevices.auth.session.MqttSessionFactory;
 import com.aws.greengrass.clientdevices.auth.session.SessionConfig;
 import com.aws.greengrass.clientdevices.auth.session.SessionCreator;
@@ -80,6 +81,7 @@ public class ClientDevicesAuthService extends PluginService {
     private final GroupManager groupManager;
 
     private final CertificateManager certificateManager;
+    private final RegistryRefreshMonitor registryRefreshMonitor;
 
     private final GreengrassServiceClientFactory clientFactory;
 
@@ -107,6 +109,7 @@ public class ClientDevicesAuthService extends PluginService {
      * @param greengrassCoreIPCService    core IPC service
      * @param mqttSessionFactory          session factory to handling mqtt credentials
      * @param sessionManager              session manager
+     * @param registryMonitor             registry-refresh monitor
      * @param clientDevicesAuthServiceApi client devices service api handle
      */
     @SuppressWarnings("PMD.ExcessiveParameterList")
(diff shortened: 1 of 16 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -80,6 +81,7 @@ public class ClientDevicesAuthService extends PluginService {
     private final GroupManager groupManager;
 
     private final CertificateManager certificateManager;
+    private final RegistryRefreshMonitor registryRefreshMonitor;
```
jbutler, line 84: There's a lot of functionality in this PR. Can you split it up? Here's what I'm envisioning:

1. refactor: add ThingRegistry -- just get the basic infrastructure in place
2. feat: add caching to enable offline auth -- continue to make cloud calls, but remember previous results if the cloud is unavailable
3. feat: add caching TTL -- cache eviction and configuration
4. feat: add cache warmer
5. feat: disk backed caching

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/RefreshableRegistry.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.iot.registry;
+
+public interface RefreshableRegistry {
```
jbutler, line 8: Consider using composition here rather than implementing an interface. For example, imagine constructing the registries with a cache refresher object. The cache itself should know when things need to be refreshed. This interface seems a bit rigid, in the sense that the registry is opaque. The thing responsible for driving the refresh can't really do anything intelligent.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/RefreshableRegistryEntry.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.iot.registry;
+
+import java.time.Instant;
+
+public interface RefreshableRegistryEntry {
```
jbutler, line 10: Similar comment here. You can likely get this behavior without forcing the registries to implement some entry interface. Imagine the cache itself tracking this information separate from the object itself.

PR #104, opened 2022-08-17
feat: enable offline certificate registry access

vaibhavmurkute:
**Description of changes:** CertificateRegistry returns locally registered results when cloud is unreachable.

**Why is this change necessary:** required for enabling offline auth

**How was this change tested:** unit tests, integ tests

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
@@ -241,7 +241,7 @@ class VerifyClientDeviceIdentityTest {
     }
 
     @Test
-    void GIVEN_broker_WHEN_verify_client_identity_with_exception_THEN_error_is_thrown(ExtensionContext context)
+    void GIVEN_broker_WHEN_verify_client_identity_with_exception_THEN_return_cached_result(ExtensionContext context)
             throws Exception {
         startNucleusWithConfig("cda.yaml");
         ignoreExceptionOfType(context, NullPointerException.class);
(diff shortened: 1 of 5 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
```diff
@@ -218,7 +216,7 @@ void GIVEN_broker_WHEN_verify_client_identity_with_validation_exception_THEN_err
     }
 
     @Test
-    void GIVEN_broker_WHEN_verify_client_identity_with_internal_server_exception_THEN_error_is_thrown(
+    void GIVEN_broker_WHEN_verify_client_identity_with_internal_server_exception_THEN_verify_with_cached_result(
```
jbutler, line 219: Hmm, this is a behavior change that we need think about carefully. Let's chat tomorrow about it

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
```diff
@@ -234,16 +232,17 @@ void GIVEN_broker_WHEN_verify_client_identity_with_internal_server_exception_THE
...
 
-            Exception err = Assertions.assertThrows(Exception.class, () -> {
-                verifyClientIdentity(ipcClient, request, null);
-            });
-            assertEquals(err.getCause().getClass(), ServiceError.class);
+            Pair<CompletableFuture<Void>, Consumer<VerifyClientDeviceIdentityResponse>> cb =
+                    asyncAssertOnConsumer((m) -> assertFalse(m.isIsValidClientDevice()));
+
+            verifyClientIdentity(ipcClient, request, cb.getRight());
+            cb.getLeft().get(10, TimeUnit.SECONDS);
```
jbutler, line 239: I think this is more complicated than it needs to be. There's no need for anything to be async. `verifyClientIdentity` already blocks.

```
Consumer<VerifyClientDeviceIdentityResponse> cb = r -> assertFalse(r.isIsValidClientDevice());
verifyClientIdentity(ipcClient, request, cb)
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/VerifyClientDeviceIdentityTest.java
vaibhavmurkute: I changed the offline-auth behavior to [throw exception when offline with uncached auth result](https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/104/commits/e7323670266c8895834dddd905fef08b90a482bc), for both CertificateRegistry and ThingRegistry. Is this okay?

jbutler (approved):

MikeDombo (approved):

PR #105, opened 2022-08-17
fix: do not throw an error when cloud upload fails

saranyailla:
**Issue #, if available:**

**Description of changes:**
When uploading the core CA to cloud fails, CDA retries for certain type of exceptions like `ThrottlingException` and `InternalServerException` where as for all other exceptions, the service goes into error state. 

But with this change, for all other non-retryable exceptions, we log the exception but do not put the service in error state. We retry for `AccessDeniedException` as well.  

When we support offline authentication, client devices do not get the core device CAs from the cloud. So, even when the core device fails to upload the core CAs to cloud, client device auth component should be running as the client devices do not rely on them in all the cases. 

Updated test. 

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
@@ -295,7 +295,7 @@ public class ClientDevicesAuthService extends PluginService {
             updateCACertificateConfig(caCerts);
             updateCaPassphraseConfig(certificateManager.getCaPassPhrase());
         } catch (CloudServiceInteractionException e) {
-            logger.atError().cause(e).log("Unable to upload core CA certs to the cloud. ");
+            logger.atError().log(e.getMessage());
         } catch (KeyStoreException | IOException | CertificateEncodingException | IllegalArgumentException
                 e) {
             serviceErrored(e);
@@ -346,7 +346,7 @@ public class ClientDevicesAuthService extends PluginService {
                     .log("Failed to put core CA certificates to cloud. Check that the core device's IoT policy grants"
                             + " the greengrass:PutCertificateAuthorities permission.");
             throw new CloudServiceInteractionException(
-                    String.format("Failed to put core %s CA certificates to cloud", thingName), e);
+                    String.format("Failed to upload core %s CA certificates to cloud", thingName), e);
         }
     }
 
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
@@ -34,6 +34,7 @@ import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService;
+import software.amazon.awssdk.services.greengrassv2data.model.AccessDeniedException;
 import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
 import software.amazon.awssdk.services.greengrassv2data.model.PutCertificateAuthoritiesRequest;
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -170,6 +171,10 @@ private int getValidCloudCallQueueSize(Topics topics) {
     @Override
     protected void install() throws InterruptedException {
         super.install();
+        configChangeHandler();
+    }
+
+    private void configChangeHandler() {
```
MikeDombo, line 177: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 14. By comparison, 99% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test. We recommend that you simplify this method or break it into multiple methods.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -343,4 +349,10 @@ private void uploadCoreDeviceCAs(List<String> certificatePemList) {
                     String.format("Failed to put core %s CA certificates to cloud", thingName), e);
         }
     }
+
+    private static List<Class> uploadCARetryExceptions() {
+        return Arrays.asList(ThrottlingException.class,
+                InternalServerException.class,
+                AccessDeniedException.class);
```
MikeDombo, line 356: Careful about retrying access denied. Double check what we're logging for failure. I don't want to emit a log every second when the permissions aren't changing.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -290,8 +294,10 @@ private void updateCAType(Topic topic) {
             uploadCoreDeviceCAs(caCerts);
             updateCACertificateConfig(caCerts);
             updateCaPassphraseConfig(certificateManager.getCaPassPhrase());
+        } catch (CloudServiceInteractionException e) {
+            logger.atError().cause(e).log("Unable to upload core CA certs to the cloud. ");
         } catch (KeyStoreException | IOException | CertificateEncodingException | IllegalArgumentException
-                | CloudServiceInteractionException e) {
+                e) {
```
MikeDombo, line 300: who is calling this method? Is it safe to block the calling thread indefinitely?

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
saranyailla: At info level, we are logging
`put-core-ca-certificate. task failed and will be retried.` only once for the first attempt. But at debug level, we log forevery attempt.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
jbutler: Does this log the failure twice? It's already logged below as part of `uploadCoreDeviceCAs`

jbutler (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -290,8 +294,10 @@ private void updateCAType(Topic topic) {
             uploadCoreDeviceCAs(caCerts);
             updateCACertificateConfig(caCerts);
             updateCaPassphraseConfig(certificateManager.getCaPassPhrase());
+        } catch (CloudServiceInteractionException e) {
+            logger.atError().log(e.getMessage());
```
MikeDombo, line 298: why aren't we logging the actual cause? Why only the message? This can hide valuable info.

MikeDombo: Nevermind.

Maybe add a comment though that this is logged by uploadCoreDeviceCAs, though I'm not so sure that's the best idea. Maybe that method shouldn't log, and should only throw and let this log it.

MikeDombo (approved):

PR #106, opened 2022-08-18
fix: only rotate certs when cis connectivity info changes

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Simplified version of https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/76. This change modifies `CISShadowMonitor` so that it rotates certificates when it _detects a change in the CIS connectivity info_, rather than a change in shadow document version.  

This fixes an issue where certificates are rotated unnecessarily when a recconect occurs (when we process a CIS shadow, we publish reported state to it, bumping the version. if a recconect occurs in the future, certs will be rotated even if connectivity info hasn't changed, because `CISShadowMonitor` only cared that the shadow version changed)

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
@@ -36,6 +36,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -64,7 +65,14 @@ public class CISShadowMonitor {
 
     private MqttClientConnection connection;
     private IotShadowClient iotShadowClient;
-    private int lastVersion = 0;
+    /**
+     * Lock to ensure shadow changes are processed serially.
+     */
+    private final Object shadowProcessingLock = new Object();
+    /**
+     * Host addresses obtained from CIS, used to de-duplicate processing requests.
+     */
+    private List<String> hostAddresses;
     private Future<?> subscribeTaskFuture;
     private final List<CertificateGenerator> monitoredCertificateGenerators = new CopyOnWriteArrayList<>();
     private final ExecutorService executorService;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -217,56 +225,74 @@ private void processCISShadow(ShadowDeltaUpdatedEvent event) {
...
-                Thread.currentThread().interrupt();
-                return;
-            } catch (Exception e) {
-                LOGGER.atError().kv(VERSION, version).cause(e)
-                        .log("Failed to get connectivity info from cloud."
-                                + " Check that the core device's IoT policy "
-                                + "grants the greengrass:GetConnectivityInfo permission");
-                return;
-            }
+            synchronized (shadowProcessingLock) {
```
jcosentino11, line 233: this isn't the technically best approach, but previous attempt, https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/76, ended up being complicated, so this may be the pragmatic approach.  open to suggestions of course

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -217,56 +225,74 @@ private void processCISShadow(ShadowDeltaUpdatedEvent event) {
...
 
-            try {
-                for (CertificateGenerator cg : monitoredCertificateGenerators) {
-                    cg.generateCertificate(connectivityInfoProvider::getCachedHostAddresses,
-                            "connectivity info was updated");
+                List<String> newHostAddresses;
+                try {
+                    newHostAddresses = fetchHostAddressesFromCIS();
+                } catch (InterruptedException e) {
+                    LOGGER.atDebug().kv(VERSION, version).cause(e)
```
MikeDombo, line 240: Don't log the cause of interrupt. We don't really care.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -217,56 +225,74 @@ private void processCISShadow(ShadowDeltaUpdatedEvent event) {
...
+                            .log("Retry workflow for getting connectivity info interrupted");
+                    Thread.currentThread().interrupt();
+                    return;
+                } catch (Exception e) {
```
MikeDombo, line 244: No retries?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -217,56 +225,74 @@ private void processCISShadow(ShadowDeltaUpdatedEvent event) {
...
-            } catch (CertificateGenerationException e) {
-                LOGGER.atError().kv(VERSION, version).cause(e).log("Failed to generate new certificates");
-                return;
-            }
 
-            try {
-                updateCISShadowReportedState(version, desiredState);
-            } finally {
-                lastVersion = version;
+                List<String> previousHostAddresses = hostAddresses;
```
MikeDombo, line 252: Should we use a set instead? We don't want duplicates and we don't really care about the order.

PR #107, opened 2022-08-18
fix: lazy load connectivity info

jbutler:
Initial calls to getCachedConnectivityInfo() may return an empty list if
it occurs before an initial call to getConnectivityInfo(). This change
will ensure that getConnectivityInfo() has been called when returning
cached results.

**Issue #, if available:**

**Description of changes:**


src/main/java/com/aws/greengrass/clientdevices/auth/iot/ConnectivityInfoProvider.java
```diff
@@ -68,7 +71,9 @@ public List<ConnectivityInfo> getConnectivityInfo() {
         try {
             GetConnectivityInfoResponse getConnectivityInfoResponse = clientFactory.getGreengrassV2DataClient()
                     .getConnectivityInfo(getConnectivityInfoRequest);
-            if (getConnectivityInfoResponse.hasConnectivityInfo()) {
+            if (getConnectivityInfoResponse == null) {
+                LOGGER.atWarn().log("Failed to retrieve connectivity info");
```
MikeDombo, line 75: add a bit more info like "response was null"?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/ConnectivityInfoProvider.java
```diff
@@ -52,6 +52,9 @@ public ConnectivityInfoProvider(DeviceConfiguration deviceConfiguration,
      * @return list of cached connectivity info items
      */
     public List<String> getCachedHostAddresses() {
```
MikeDombo, line 54: is this a safe change to make, since now this call can block while it waits for a cloud response, right?

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ConnectivityInfoProviderTest.java
```diff
@@ -115,7 +115,7 @@ void GIVEN_cached_connectivity_info_WHEN_get_cached_connectivity_info_THEN_conne
         doReturn(getConnectivityInfoResponse).when(greengrassV2DataClient)
                 .getConnectivityInfo(any(GetConnectivityInfoRequest.class));
 
-        connectivityInfoProvider.getConnectivityInfo();
+        //connectivityInfoProvider.getConnectivityInfo();
```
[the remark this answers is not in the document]
jbutler: Sorry, meant to delete

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/ConnectivityInfoProvider.java
jbutler: Good call... should test with EMQx to make sure we're not completely breaking things (if it breaks, that's an EMQx bug).

MikeDombo: What if there is no IPD, does this work correctly when there is no CIS info at all?

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/ConnectivityInfoProvider.java
jbutler: It will just get an empty response

jbutler: This shouldn't block cert gen. That process is all async anyway. But we should test it (if it's not, that's still a bug)

MikeDombo (commented):
Still want this change?

PR #108, opened 2022-08-18
fix: rotate certs when change in cis shadow version field

jcosentino11:
Issue #, if available:

Description of changes:

This change modifies CISShadowMonitor so that it rotates certificates when it detects a change in the CIS shadow `version` field, rather than a change in shadow _document_ version.

Why is this change necessary:

This fixes an issue where certificates are rotated unnecessarily when a recconect occurs (when we process a CIS shadow, we publish reported state to it, bumping the shadow document version. if a recconect occurs in the future, certs will be rotated even if connectivity info hasn't changed, because CISShadowMonitor only cared that the shadow version changed)

How was this change tested:
Tested by blocking MQTT traffic (to force a disconnect), then unblocking it to force a shadow fetch. Ensured that the shadow was not reprocessed.

```
iptables -A OUTPUT -p tcp --destination-port 8883 -s 172.31.6.196 -j DROP
...
iptables -F
```

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
@@ -36,6 +36,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -64,7 +65,7 @@ public class CISShadowMonitor {
 
     private MqttClientConnection connection;
     private IotShadowClient iotShadowClient;
-    private int lastVersion = 0;
+    private String lastVersion;
     private Future<?> subscribeTaskFuture;
     private final List<CertificateGenerator> monitoredCertificateGenerators = new CopyOnWriteArrayList<>();
     private final ExecutorService executorService;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -209,16 +210,23 @@ private void subscribeToShadowTopics() throws InterruptedException {
     }
 
     private void processCISShadow(GetShadowResponse response) {
-        processCISShadow(response.version, response.state.desired);
+        String cisVersion = (String) response.state.desired.get("version");
```
MikeDombo, line 213: blind casting is scary. Use `Coerce.toString` at least

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -273,11 +281,10 @@ private synchronized void processCISShadow(int version, Map<String, Object> desi
      * @param version CIS shadow version
      * @param desired CIS shadow <b>desired</b> state
      */
-    private void updateCISShadowReportedState(int version, Map<String, Object> desired) {
+    private void updateCISShadowReportedState(String version, Map<String, Object> desired) {
         LOGGER.atInfo().kv(VERSION, version).log("Reporting CIS version");
         UpdateShadowRequest updateShadowRequest = new UpdateShadowRequest();
         updateShadowRequest.thingName = shadowName;
-        updateShadowRequest.version = version;
```
MikeDombo, line 280: didn't we fix a bug where this request was failing because we didn't include a version? And now we're putting that bug back

MikeDombo: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/72

Perhaps version isn't needed.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
MikeDombo: Is this change backwards/forwards compatible? If the customer updates to this version of CDA and then rolls back, will this change break anything?

jcosentino11: yep, completely missed that

MikeDombo: Un-resolved. Can you answer the question about compatibility too?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
jcosentino11: that was cause of missing `state` field, missing version should be fine, @jbutler just did a successful manual test

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
```diff
@@ -273,11 +281,10 @@ private synchronized void processCISShadow(int version, Map<String, Object> desi
      * @param version CIS shadow version
      * @param desired CIS shadow <b>desired</b> state
      */
-    private void updateCISShadowReportedState(int version, Map<String, Object> desired) {
+    private void updateCISShadowReportedState(String version, Map<String, Object> desired) {
```
jbutler, line 284: This works, but is confusing. Let's just update this to:

```
private void updateCISShadowReportedState(Map<String, Object> reported)
```

Then when calling this, pass the desired state and leave a comment saying that reported should now match desired

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CISShadowMonitor.java
jcosentino11: sorry didn't see the compatibility comment at first--- no this won't affect compatibility,  the shadow contents itself aren't being changed

MikeDombo (approved):

jbutler (approved):

PR #110, opened 2022-08-25
refactor: update CDA configuration

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Model  the CA config of CDA config and also include BYO CA keys. 
- Pass the CAConfiguration object to cert manager

- Needed to support BYOCA feature. 
- Refactoring code

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -44,6 +44,7 @@ public class CertificateManager {
     private final Clock clock;
     private final Map<GetCertificateRequest, CertificateGenerator> certSubscriptions = new ConcurrentHashMap<>();
     private CertificatesConfig certificatesConfig;
+    @SuppressWarnings("PMD.UnusedPrivateField")
     private CAConfiguration caConfiguration;
 
 
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -16,9 +16,9 @@ import com.aws.greengrass.clientdevices.auth.certificate.CertificateStore;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificatesConfig;
 import com.aws.greengrass.clientdevices.auth.certificate.ClientCertificateGenerator;
 import com.aws.greengrass.clientdevices.auth.certificate.ServerCertificateGenerator;
+import com.aws.greengrass.clientdevices.auth.configuration.CAConfiguration;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
 import com.aws.greengrass.clientdevices.auth.iot.ConnectivityInfoProvider;
-import com.aws.greengrass.config.Topics;
 import lombok.NonNull;
 
 import java.io.IOException;
@@ -44,6 +44,7 @@ public class CertificateManager {
     private final Clock clock;
     private final Map<GetCertificateRequest, CertificateGenerator> certSubscriptions = new ConcurrentHashMap<>();
     private CertificatesConfig certificatesConfig;
+    private CAConfiguration caConfiguration;
 
 
     /**
@@ -72,8 +73,8 @@ public class CertificateManager {
         this.certificatesConfig = certificatesConfig;
     }
 
-    public void setCertificateStoreConfig(Topics cdaComponentConfiguration, Topics cdaRuntimeConfiguration) {
-        certificateStore.setCertificateStoreConfig(cdaComponentConfiguration, cdaRuntimeConfiguration);
+    public void setCAConfiguration(CAConfiguration caConfiguration) {
+        this.caConfiguration = caConfiguration;
     }
 
     /**
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -234,4 +234,8 @@ public class CertificateManager {
         certExpiryMonitor.removeFromMonitor(gen);
         cisShadowMonitor.removeFromMonitor(gen);
     }
+
+    public void  updateCAConfiguration() {
+        caConfiguration.updateCAConfiguration();
+    }
 }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -18,6 +18,7 @@ import com.aws.greengrass.clientdevices.auth.certificate.ClientCertificateGenera
 import com.aws.greengrass.clientdevices.auth.certificate.ServerCertificateGenerator;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
 import com.aws.greengrass.clientdevices.auth.iot.ConnectivityInfoProvider;
+import com.aws.greengrass.config.Topics;
 import lombok.NonNull;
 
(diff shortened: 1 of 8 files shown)


src/test/resources/com/aws/greengrass/clientdevices/auth/config_with_ec_ca.yaml
```diff
@@ -2,10 +2,10 @@
 services:
   aws.greengrass.clientdevices.Auth:
     configuration:
-      ca_type: ["ECDSA_P256"]
+      certificateAuthority:
+        ca_type: ["ECDSA_P256"]
   main:
     lifecycle:
-      install:
-        all: echo All installed
```
MikeDombo, line 9: why is this changed? this isn't a valid recipe anymore.

MikeDombo: Not that it needs to be a valid recipe, you can just remove the echo thing.

saranyailla: yeah, tests using invalid recipes have unnecessary component broken logs.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -137,6 +140,7 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager, Certif
         SessionCreator.registerSessionFactory("mqtt", mqttSessionFactory);
         certificateManager.updateCertificatesConfiguration(new CertificatesConfig(this.getConfig()));
         sessionManager.setSessionConfig(new SessionConfig(this.getConfig()));
+        certificateManager.setCertificateStoreConfig(this.getConfig(), this.getRuntimeConfig());
```
jbutler, line 143: Take a look at how we deserialize GroupDefinition - ideally we can just follow a similar pattern here.

A couple benefits (I can think of):
* Don't need to plumb `Topics` all the way through. No need to deal with casting. We can add appropriate getters
* Config is immutable. Behavior is easier to reason about
* No need to add a bunch of static strings to this class and import them in other places

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -83,6 +83,7 @@ public CertificateStore(Path workPath) {
         this.workPath = workPath;
     }
 
+
```
MikeDombo, line 86: nit: remove

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+public class CAConfiguration {
+    private final Topics certificateAuthorityTopics;
+    private final Topics cdaRuntimeTopics;
+
+    /**
+     * Construct CA Configuration object from the CDA component configuration.
+     *
+     * @param cdaConfigTopics CDA service configuration topics
+     */
+    @Inject
```
MikeDombo, line 32: I don't think you can actually inject this.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -63,6 +64,9 @@
 public class ClientDevicesAuthService extends PluginService {
     public static final String CLIENT_DEVICES_AUTH_SERVICE_NAME = "aws.greengrass.clientdevices.Auth";
     public static final String DEVICE_GROUPS_TOPICS = "deviceGroups";
+    public static final String CERTIFICATE_AUTHORITY_TOPIC = "certificateAuthority";
```
jbutler, line 67: Move sub-keys of `certificateAuthority` into CAConfiguration

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+import static com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService.CA_CERTIFICATE_URI;
+import static com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService.CA_PASSPHRASE;
+import static com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService.CA_PRIVATE_KEY_URI;
+import static com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService.CA_TYPE_TOPIC;
+import static com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService.CERTIFICATE_AUTHORITY_TOPIC;
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.RUNTIME_STORE_NAMESPACE_TOPIC;
+
+public class CAConfiguration {
+    private final Topics certificateAuthorityTopics;
```
jbutler, line 23: Now take this one step further and make this immutable. No need to store topics. We want `CertificateManager` and `CertificateStore` to always be operating on the same config. You can't achieve that if this class is dynamically reading from config

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
jbutler: I'm also going to nitpick on this. Please remove

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -70,6 +74,10 @@ public void updateCertificatesConfiguration(CertificatesConfig certificatesConfi
         this.certificatesConfig = certificatesConfig;
     }
 
+    public void setCAConfiguration(CAConfiguration caConfiguration) {
```
jbutler, line 77: Right now you have `setCAConfiguration` and `updateCAConfiguration` -- remove one of them

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    private String caCertificateUri;
+    private List<String> caTypeList;
+    private String caPassphrase;
+
+    /**
+     * Updates the CA configuration with the latest CDA config.
+     *
+     * @param cdaConfigTopics CDA service configuration topics
+     */
+    public synchronized void setCAConfiguration(Topics cdaConfigTopics) {
```
jbutler, line 34: Actually, this class still isn't immutable. Think about the cert rotation workflow - it's still possible for that to trigger during a config update which could result in us getting into a non-deterministic state.

I actually want you to re-instantiate this config class whenever there is a config change. Everything should be final

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -43,6 +44,9 @@ public class CertificateManager {
     private final Clock clock;
     private final Map<GetCertificateRequest, CertificateGenerator> certSubscriptions = new ConcurrentHashMap<>();
     private CertificatesConfig certificatesConfig;
+    @SuppressWarnings("PMD.UnusedPrivateField")
+    private CAConfiguration caConfiguration;
```
MikeDombo, line 48: future PR: read/write access to this needs to be synchronized or else you may be reading from multiple configuration instances.

MikeDombo (approved):

PR #111, opened 2022-08-29
feat: registry cache ttl

vaibhavmurkute:
**Description of changes:** This change adds TTL to the registry entries and adds a refresh mechanism that removes stale (invalid) registry entries, scheduled to run at the configured frequency. For now, this TTL and the refresh frequency is set to 1 day. These parameters will be made customer-configurable in subsequent PR.

**Why is this change necessary:** To manage registry cache: invalidate registry entries and force-consult cloud after TTL.

**How was this change tested:** Unit tests, integ tests (`mvn verify`)

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateEntry.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateEntry.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.iot.registry;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+import java.time.Instant;
+
+@Getter
+@AllArgsConstructor
+public class CertificateEntry {
+    private Instant validTill;
+    private String certificateHash;
+    private String iotCertificateId;
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!(obj instanceof CertificateEntry)) {
+            return false;
+        }
+        return this.getIotCertificateId()
+                .equals(((CertificateEntry) obj).iotCertificateId);
+    }
+
+    @Override
+    public int hashCode() {
+        return this.getIotCertificateId().hashCode();
+    }
+}
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
@@ -11,8 +11,11 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Digest;
 import com.aws.greengrass.util.Utils;
+import lombok.AccessLevel;
+import lombok.Getter;
 
 import java.security.NoSuchAlgorithmException;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -22,10 +25,11 @@ import javax.inject.Inject;
 
 public class CertificateRegistry {
     private static final Logger logger = LogManager.getLogger(CertificateRegistry.class);
-    // holds mapping of certificateHash (SHA-256 hash of certificatePem) to IoT Certificate Id;
+    // holds mapping of certificateHash (SHA-256 hash of certificatePem) to its registry entry;
     // size-bound by default cache size, evicts oldest written entry if the max size is reached
-    private final Map<String, String> registry = Collections.synchronizedMap(
-            new LinkedHashMap<String, String>(RegistryConfig.REGISTRY_CACHE_SIZE, 0.75f, false) {
+    @Getter(AccessLevel.PROTECTED)
+    private final Map<String, CertificateEntry> registry = Collections.synchronizedMap(
+            new LinkedHashMap<String, CertificateEntry>(RegistryConfig.REGISTRY_CACHE_SIZE, 0.75f, false) {
                 @Override
                 protected boolean removeEldestEntry(Map.Entry eldest) {
                     return size() > RegistryConfig.REGISTRY_CACHE_SIZE;
@@ -37,11 +41,13 @@ public class CertificateRegistry {
     /**
      * Constructor.
      *
-     * @param iotAuthClient IoT Auth Client
+     * @param iotAuthClient    IoT Auth Client
+     * @param refreshScheduler Registry refresh scheduler
      */
     @Inject
-    public CertificateRegistry(IotAuthClient iotAuthClient) {
+    public CertificateRegistry(IotAuthClient iotAuthClient, RegistryRefreshScheduler refreshScheduler) {
         this.iotAuthClient = iotAuthClient;
+        refreshScheduler.schedule(this::refreshRegistry);
     }
 
     /**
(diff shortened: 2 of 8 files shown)


src/test/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistryTest.java
```diff
@@ -115,4 +120,29 @@ void GIVEN_offline_initialization_WHEN_isCertificateValid_THEN_throw_exception()
...
         assertThrows(CloudServiceInteractionException.class, () -> registry.getIotCertificateIdForPem(mockCertPem));
     }
+
+    @Test
+    void GIVEN_certRegistry_WHEN_refreshRegistry_THEN_stale_entries_removed() {
+        String certHash1 = "certHash1";
+        String certHash2 = "certHash2";
+        String certHash3 = "certHash3";
+
+        CertificateEntry validEntry = new CertificateEntry(Instant.now().plusSeconds(10L),
```
jcosentino11, line 130: should use same base instant for all entries---the `Instant.now()` two lines below could be different than the one here.  same goes for other test class

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -62,16 +68,46 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
     }
 
     private void registerCertificateForThing(Thing thing, Certificate certificate) {
-        registry.put(thing.getThingName(), certificate.getIotCertificateId());
+        Set<CertificateEntry> certEntrySet = registry.computeIfAbsent(thing.getThingName(), s -> new HashSet<>());
+        // replace previous association with a new entry with extended TTL
+        clearThingCertificateAssociation(thing, certificate);
+        certEntrySet.add(getNewCertificateEntry(certificate));
+        registry.put(thing.getThingName(), certEntrySet);
```
jcosentino11, line 75: not needed? already taken care of from `computeIfAbsent` above

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -62,16 +68,46 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+        return certEntry.getValidTill().isAfter(Instant.now());
+    }
+
+    /**
+     * Removes stale (invalid) registry entries.
+     */
+    void refreshRegistry() {
+        registry.values().forEach(certificateEntries ->
+                certificateEntries.removeIf(certEntry -> !isValidCertificateEntry(certEntry)));
+        registry.values().removeIf(Set::isEmpty);
```
jcosentino11, line 111: doesn't look like this does anything? did you mean to remove the entry instead?

vaibhavmurkute: The removal on #L109 might leave an entry with an empty-set value in the registry map: `(String, Set<CertificateEntry>)`. So we remove such unnecessary empty-set entries.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
vaibhavmurkute: right, not needed

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
jcosentino11: Ah, i didn't realize removal here takes care of removing the full entry. nevermind :)

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateEntry.java
```diff
@@ -0,0 +1,33 @@
...
+    private String certificateHash;
+    private String iotCertificateId;
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!(obj instanceof CertificateEntry)) {
+            return false;
+        }
+        return this.getIotCertificateId()
+                .equals(((CertificateEntry) obj).iotCertificateId);
```
jbutler, line 26: This is too lenient. What if there is no cert id?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -22,10 +25,11 @@
 
 public class CertificateRegistry {
     private static final Logger logger = LogManager.getLogger(CertificateRegistry.class);
-    // holds mapping of certificateHash (SHA-256 hash of certificatePem) to IoT Certificate Id;
+    // holds mapping of certificateHash (SHA-256 hash of certificatePem) to its registry entry;
     // size-bound by default cache size, evicts oldest written entry if the max size is reached
-    private final Map<String, String> registry = Collections.synchronizedMap(
-            new LinkedHashMap<String, String>(RegistryConfig.REGISTRY_CACHE_SIZE, 0.75f, false) {
+    @Getter(AccessLevel.PROTECTED)
```
jbutler, line 30: Protected getters always make me raise an eyebrow. Avoid doing this, ESPECIALLY just for unit tests. If you can't initialize your object under test using the public API, then your test is too tightly coupled to the implementation and you should refactor.
Your tests for this class should not break when we replace this map with something else.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -157,4 +166,21 @@ private String getCertificateHash(String certificatePem) {
...
+    }
+
+    private boolean isValidCertificateEntry(CertificateEntry certEntry) {
+        return certEntry.getValidTill().isAfter(Instant.now());
+    }
+
+    /**
+     * Removes stale (invalid) registry entries.
+     */
+    void refreshRegistry() {
```
jbutler, line 183: May want to rename this. It's not refreshing anything, it's just purging.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/RegistryConfig.java
```diff
@@ -6,5 +6,7 @@
 package com.aws.greengrass.clientdevices.auth.iot.registry;
 
 public class RegistryConfig {
-    public static final int REGISTRY_CACHE_SIZE = 50;
+    protected static final int REGISTRY_CACHE_SIZE = 50;
+    protected static final int REGISTRY_CACHE_ENTRY_TTL_SECONDS = 24 * 60 * 60; // 1 day
+    protected static final int REGISTRY_REFRESH_FREQUENCY_SECONDS = 24 * 60 * 60; // 1 day
```
jbutler, line 11: Do you think this is the config we want to expose? What if there is no network when the refresh cycle runs? How will you handle intermittent cloud failures?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/RegistryRefreshScheduler.java
```diff
@@ -0,0 +1,64 @@
...
+    public RegistryRefreshScheduler(ScheduledExecutorService ses) {
+        this.ses = ses;
+    }
+
+    /**
+     * Schedules the runnable to be run at configured interval.
+     *
+     * @param runnable Runnable registry refresh
+     */
+    public void schedule(Runnable runnable) {
```
jbutler, line 36: How are certs removed?

PR #116, opened 2022-09-06
feat: support configurable CA

Nelsonochoam:
**Description of changes:**
This PR adds support for configuring custom certificates from the provided `certificateAuthority` configuration.

This introduces the ability for customers to provide their own CA instead of having the CDA generate it.

There are 2 new tests that ensure the appropriate certificate chain is set once it is provided.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,13 +73,13 @@
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.7.0-SNAPSHOT</version>
+            <version>2.8.0-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.7.0-SNAPSHOT</version>
+            <version>2.8.0-SNAPSHOT</version>
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -73,13 +73,13 @@
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.6.0-SNAPSHOT</version>
+            <version>2.7.0-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.6.0-SNAPSHOT</version>
+            <version>2.7.0-SNAPSHOT</version>
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -18,6 +18,9 @@ import com.aws.greengrass.clientdevices.auth.certificate.ClientCertificateGenera
 import com.aws.greengrass.clientdevices.auth.certificate.ServerCertificateGenerator;
 import com.aws.greengrass.clientdevices.auth.configuration.CAConfiguration;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
+import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
+import com.aws.greengrass.clientdevices.auth.exception.InvalidCertificateAuthorityException;
+import com.aws.greengrass.clientdevices.auth.exception.InvalidConfigurationException;
 import com.aws.greengrass.clientdevices.auth.iot.ConnectivityInfoProvider;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
@@ -259,15 +262,14 @@ public class CertificateManager {
     /**
      * Configures the KeyStore to use a certificates provided from the CA configuration.
      *
-     * @throws KeyStoreException if the stored key is not a private key
-     * @throws Exception if failing to get the keyPair or certificateChain
-     * @throws IllegalArgumentException if configuration is invalid for CA
+     * @throws InvalidConfigurationException if the CA configuration doesn't have both a privateKeyUri and
+     *                                      certificateUri
      */
-    @SuppressWarnings("PMD.SignatureDeclareThrowsException")
-    public void configureCustomCA() throws KeyStoreException, Exception {
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public void configureCustomCA() throws InvalidConfigurationException {
         if (!caConfiguration.isUsingCustomCA()) {
-            String errMsg = "Invalid configuration: certificateUri and privateKeyUri are required.";
-            throw new IllegalArgumentException(errMsg);
+            throw new InvalidConfigurationException(
+                    "Invalid configuration: certificateUri and privateKeyUri are required.");
         }
 
         URI privateKeyUri = caConfiguration.getPrivateKeyUri().get();
@@ -278,27 +280,30 @@ public class CertificateManager {
                 .retryableExceptions(Collections.singletonList(ServiceUnavailableException.class)).build();
 
 
-        KeyPair keyPair = RetryUtils.runWithRetry(retryConfig,
-                () -> securityService.getKeyPair(privateKeyUri, certificateUri),
-                "get-key-pair", logger);
+        try {
+            KeyPair keyPair = RetryUtils.runWithRetry(retryConfig,
+                    () -> securityService.getKeyPair(privateKeyUri, certificateUri),
+                    "get-key-pair", logger);
 
-        X509Certificate[] certificateChain = RetryUtils.runWithRetry(retryConfig,
-                () -> securityService.getCertificateChain(privateKeyUri, certificateUri),
-                "get-certificate-chain", logger);
+            X509Certificate[] certificateChain = RetryUtils.runWithRetry(retryConfig,
+                    () -> securityService.getCertificateChain(privateKeyUri, certificateUri),
+                    "get-certificate-chain", logger);
 
-
-        certificateStore.setCaCertificateChain(certificateChain);
-        certificateStore.setCaPrivateKey(keyPair.getPrivate());
+            certificateStore.setCaCertificateChain(certificateChain);
+            certificateStore.setCaPrivateKey(keyPair.getPrivate());
+        } catch (Exception e) {
+            throw new InvalidCertificateAuthorityException(String.format("Failed to configure CA: There was an error "
+                    + "reading the provided private key %s or certificate chain %s", privateKeyUri, certificateUri), e);
+        }
     }
 
     /**
      * Uploads the stored certificates to the cloud.
      *
-     * @param thingName  Core device name
-     *
-     * @throws CertificateEncodingException  If unable to get certificate encoding
-     * @throws KeyStoreException if unable to retrieve the certificate
-     * @throws IOException If unable to read certificate
+     * @param thingName Core device name
+     * @throws CertificateEncodingException      If unable to get certificate encoding
+     * @throws KeyStoreException                if unable to retrieve the certificate
+     * @throws IOException                      If unable to read certificate
      */
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     public void uploadCoreDeviceCAs(String thingName) throws CertificateEncodingException, KeyStoreException,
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -73,13 +73,13 @@
...
             <artifactId>nucleus</artifactId>
-            <version>2.6.0-SNAPSHOT</version>
+            <version>2.7.0-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.6.0-SNAPSHOT</version>
+            <version>2.7.0-SNAPSHOT</version>
```
Nelsonochoam, line 82: Depends on a new method on the security service of the Nucleus [see here](https://github.com/aws-greengrass/aws-greengrass-nucleus/commit/c4cfe2bf8dace146556d2ddabd4dd90b474004f7#diff-d808e9ba1be2c30f79fe423fcc84d462f988c5696bd699894ea720d993c333faR179). Only available after snapshot 2.7.0.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -253,18 +259,39 @@ private void removeCGFromMonitors(CertificateGenerator gen) {
...
 
+    /**
+     * Configures the KeyStore to use a certificates provided from the CA configuration.
+     *
+     * @throws KeyStoreException if the stored key is not a private key
+     * @throws Exception if failing to get the keyPair or certificateChain
+     * @throws IllegalArgumentException if configuration is invalid for CA
+     */
+    @SuppressWarnings("PMD.SignatureDeclareThrowsException")
+    public void configureCustomCA() throws KeyStoreException, Exception {
```
Nelsonochoam, line 270: The reason for suppressing warnings is we use the `RetryUtils` which under the hood just throw plain `Exceptions` we would have to update it if we want to avoid doing this.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -177,6 +173,17 @@ private void subscribeToConfigChanges() {
         config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe(this::configChangeHandler);
     }
 
+    private void onConfigurationChanged() {
```
Nelsonochoam, line 176: Just moved this down here

src/test/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfigurationTest.java
```diff
@@ -44,56 +46,55 @@ void afterEach() throws IOException {
     }
 
     @Test
-    public void GIVEN_cdaDefaultConfiguration_WHEN_getCATypeList_THEN_returnsEmptyList() {
+    public void GIVEN_cdaDefaultConfiguration_WHEN_getCATypeList_THEN_returnsEmptyList() throws URISyntaxException {
```
Nelsonochoam, line 49: All these changes are just to declare that it might throw an exception if you provide and invalid URI

src/test/java/com/aws/greengrass/clientdevices/auth/CertificateManagerTest.java
```diff
@@ -69,25 +87,88 @@ public class CertificateManagerTest {
...
         certificateManager.update("");
     }
 
+    @AfterEach
+    void afterEach() {
+        reset(securityServiceMock);
+    }
+
+    @Test
+    void GIVEN_customCAConfiguration_WHEN_configureCustomCA_THEN_returnsCustomCA() throws Exception {
```
Nelsonochoam, line 110: Tests custom certificate chain can be set.

src/test/java/com/aws/greengrass/clientdevices/auth/CertificateManagerTest.java
```diff
@@ -69,25 +87,88 @@ public class CertificateManagerTest {
...
+        certificateManager.configureCustomCA();
+
+        // Then
+        List<String> caPemStrings = certificateManager.getCACertificates();
+        String caPem = caPemStrings.get(0);
+        assertEquals(caPem, CertificateHelper.toPem(caCertificate));
+    }
+
+    @Test
+    void Given_defaultCAConfiguration_THEN_returnsAutoGeneratedCA() throws Exception {
```
Nelsonochoam, line 145: Just ensures that when the old config produces a default ca and not the one we are providing

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
```diff
@@ -104,17 +117,25 @@ private static CertificateStore.CAType getCaTypeFromConfiguration(Topics configu
...
+
+        return Optional.of(new URI(maybePrivateKeyUri));
     }
 
-    private static String getCaCertificateUriFromConfiguration(Topics certAuthorityTopic) {
-        // NOTE: Parse the key to ensure it is a valid URI
-        //  before returning it
-        Topic privateKeyUriTopic = certAuthorityTopic.find(CA_CERTIFICATE_URI);
-        return Coerce.toString(privateKeyUriTopic);
+    private static Optional<URI> getCaCertificateUriFromConfiguration(Topics certAuthorityTopic) throws
```
Nelsonochoam, line 131: Changed this method to parse the configuration string into a valid URI, it returns optional since it could be well... optional...

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -183,8 +208,8 @@ private void createAndStoreDefaultKeyStore(CAType caType) throws KeyStoreExcepti
         }
         // generate new passphrase for new CA certificate
         passphrase = generateRandomPassphrase().toCharArray();
-        Certificate[] certificateChain = { caCertificate };
-        ks.setKeyEntry("CA", kp.getPrivate(), getPassphrase(), certificateChain);
+        caCertificateChain = new Certificate[]{ caCertificate };
+        ks.setKeyEntry(CA_KEY_ALIAS, kp.getPrivate(), getPassphrase(), caCertificateChain);
```
Nelsonochoam, line 212: Avoid hard coding the alias and store the generated certificate chain

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -275,28 +282,35 @@ private void updateDeviceGroups(WhatHappened whatHappened, Topics deviceGroupsTo
         }
     }
 
-    private void updateCAType() {
-        // NOTE: This entire method will be moved out of here and will become part of a workflow.
-        String thingName = Coerce.toString(deviceConfiguration.getThingName());
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
```
Nelsonochoam, line 285: The reason we are suppressing here is because we are using the retries that throw a generic exception if it fails.

pom.xml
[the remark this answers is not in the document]
MikeDombo: 2.7 has already been released. I don't think that the feature was added in 2.7, but 2.8.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -253,18 +256,39 @@ private void removeCGFromMonitors(CertificateGenerator gen) {
...
+        if (!caConfiguration.isUsingCustomCA()) {
+            String errMsg = "Invalid configuration: certificateUri and privateKeyUri are required.";
```
MikeDombo, line 269: create a static final string instead of an inline string. Or just put it in the exception constructor without an intermediate variable.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -253,18 +256,39 @@ private void removeCGFromMonitors(CertificateGenerator gen) {
...
+            throw new IllegalArgumentException(errMsg);
```
MikeDombo, line 270: Use a checked exception.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
MikeDombo: This method should hide that complexity. The only exception types thrown by retryutil are the exceptions which can be thrown by the callable which it is retrying. You can simply extract out the true exception type and rethrow the real exception.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -280,21 +304,32 @@ private RetryUtils.RetryConfig getRetryConfig() {
...
                     () -> clientFactory.getGreengrassV2DataClient().putCertificateAuthorities(request),
                     "put-core-ca-certificate", logger);
         } catch (InterruptedException e) {
             logger.atInfo().log("Put core CA certificates got interrupted");
             // interrupt the current thread so that higher-level interrupt handlers can take care of it
             Thread.currentThread().interrupt();
         } catch (Exception e) {
-            throw new CloudServiceInteractionException("Failed to put core CA certificates to cloud. Check that the "
-                    + "core device's IoT policy grants the greengrass:PutCertificateAuthorities permission.", e);
+            String errMsg = "Failed to put core CA certificates to cloud. Check that the "
```
MikeDombo, line 328: why is this logging here now instead of throwing as before? The caller no doesn't know if it succeeds.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -177,6 +170,17 @@ private void subscribeToConfigChanges() {
...
     }
 
+    private void onConfigurationChanged() {
+        runtimeConfiguration = new RuntimeConfiguration(getRuntimeConfig());
+
+        try {
+            caConfiguration = CAConfiguration.from(getConfig());
+            certificateManager.updateCAConfiguration(caConfiguration);
+        } catch (URISyntaxException e) {
+            serviceErrored(e);
```
MikeDombo, line 180: we shouldn't be using serviceErrored unless you want to break the component.

Nelsonochoam: Got it and if there is and URISyntax we should break the component, given we that is an invalid configuration

MikeDombo: No, I don't think you do want that. This component is currently written in a way which is impossible to get out of the broken state. If we do want to break, then we must have a way to get out of that bad state. To get out of broken, you must `requestReinstall()`.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
Nelsonochoam: Yeah this was more statistical I didn't like how it looked putting it directly in the constructor but I can change that

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
jbutler: Breaking when customer provides garbage config is the appropriate thing to do here.

What's the recommended way to "un-break" yourself? Does a deployment not trigger a restart?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -128,9 +136,25 @@ public PrivateKey getCAPrivateKey() throws KeyStoreException {
      * @throws KeyStoreException if unable to retrieve the certificate
      */
     public X509Certificate getCACertificate() throws KeyStoreException {
-        return (X509Certificate) keyStore.getCertificate(CA_KEY_ALIAS);
+        return (X509Certificate) getCaCertificateChain()[0];
```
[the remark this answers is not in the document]
Nelsonochoam: There is a method that returns the entire chain (it is just a getter though)

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
MikeDombo: No, as we've previous discussed, plugins are not automatically restarted. The plugin can choose to reinstall using `requestReinstall()`. It should do this when configuration changes which will un-break it.

Breaking is generally not the right thing to do because it will take down all components which depend on CDA due to CDA shutting down.

jbutler: Dependent components can't function without valid CA configuration. I'm not worried about that.

That's fine. We can request reinstall if config changes while broken. In the meantime, this is an intermediate PR. CDA already transitions to broken and these config knobs aren't exposed yet. I'd recommend fixing in a follow up PR.

jbutler (approved):

saranyailla (approved):

PR #118, opened 2022-09-07
Refactor application logic into use cases

Nelsonochoam:
**Description of changes:**
This changes extract application logic away from the ClientDeviceAuthService by moving it to a use case (please note that the use case refactored in this PR will still be improved, the certificateManager won't be a thing in the future). 

Some additional changes:

* Introduced a class called CDAConfiguration that represents the entire application configuration and composes several configuration object

**Why is this change necessary:**
Increased complexity makes it hard for us to reason about software. Thinking about use cases (workflows) make it easier for us to develop the application and increase visibility into what is the functionality we support, making it easier to maintain and tailor separate scenarios.

**How was this change tested:**
Ensure tests are still passing and new tests for use cases.

PR #117, opened 2022-09-07
refactor: initial connectivity domain structure

jbutler:
**Issue #, if available:**

**Description of changes:**
Initial stab at creating a `connectivity` domain. This doesn't actually do anything yet, but sets the stage for future changes.

**Why is this change necessary:**
Needed to iterative move toward a use case driven flow, which will eventually allow us to de-couple CIS updates and the cert rotation workflows.

**How was this change tested:**
Unit tests

**Any additional information or context required to review the change:**

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #119, opened 2022-09-08
refactor: introduce usecases

Nelsonochoam:
**Description of changes:**
This PR introduces use cases. They provide a means of separating application logic and tailor an application workflow but interacting with the domain of the application.

Currently there is low cohesion in some of the classes we have given they don't have well-defined responsibilities and they end up being a mix of application logic, with domain logic and even infra logic. This changes start defining those layers that will help with code discoverability and also help to untangle some classes from being responsible of things they shouldn't be.

Wrote unit tests

For more information about how this would look when used see [this PR](https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/118/files#diff-5ed149cd1cf64a63ac65ce04f8e317e447aedac8aebebc261ff91bc3cca9b076R202)


src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class UseCases {
+    private final Context context;
+
+    public interface UseCase<R, D> {
+        @SuppressWarnings("PMD.SignatureDeclareThrowsException")
+        R execute(D dto) throws Exception;
+    }
+
+    @Inject
```
MikeDombo, line 22: this can't be injected. You don't want it to inject any random `Topics`

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.inject.Inject;
+
+
+public class UseCases {
+    private final Context context;
+
+    public interface UseCase<R, D> {
+        @SuppressWarnings("PMD.SignatureDeclareThrowsException")
+        R execute(D dto) throws Exception;
+    }
```
MikeDombo, line 20: We have `CrashableFunction` in the Nucleus which is the same thing (more or less)

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
Nelsonochoam: True

jbutler (dismissed):
Other usage examples in #117 and #118

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
Nelsonochoam: Just saw it. The only issue with that interface is that forces the Exception generic type to be defined and if we want to use it for a function that does not throw any exception it wouldn't work (please correct me if I am wrong). For example

```
class Foo implements CrashableFuntion<Void, Void, Exception> {
     @Override
     public Void apply(Void a) {
         System.out.println("Hello);
    }
}
```

would complain saying we need to throw an exception or pass something as the 3rd generic param

MikeDombo: You should be able to just use a lambda `(Void a) -> { sout; return null }`.

jbutler (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,57 @@
...
+    private static UseCases instance;
+
+    // Delegates to CrashableFunction but provides a domain rich alias
+    public interface UseCase<R, D, E extends Exception> extends CrashableFunction<D, R, E> {}
+
+    private UseCases(Context context) {
+        this.context = context;
+    }
+
+    public static void init(Topics topics) {
```
jbutler, line 24: Add the call to `init` in the service entry?

Nelsonochoam: Yeah, I intended to do that on the next PR but I can do it in this one

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -128,6 +129,9 @@ public ClientDevicesAuthService(Topics topics, GroupManager groupManager,
         SessionCreator.registerSessionFactory("mqtt", mqttSessionFactory);
         certificateManager.updateCertificatesConfiguration(new CertificatesConfig(getConfig()));
         sessionManager.setSessionConfig(new SessionConfig(getConfig()));
+
+        // Start use cases, so they could be invoked anywhere on the app.
+        UseCases.init(getContext());
```
Nelsonochoam, line 134: This just starts the useCases with the context so we can invoke them anywhere in the app

src/main/java/com/aws/greengrass/clientdevices/auth/api/usecases/UseCaseDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+package com.aws.greengrass.clientdevices.auth.api.usecases;
+
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+class UseCaseDecorator implements InvocationHandler {
```
Nelsonochoam, line 15: Why this? - I thought it might be a good idea to have some sort of logging about useCase runtime to verify if introducing useCases is impacting application run time in a negative way. Additionally this will give us a bit more insight into what the application is doing and potentially in the future this are metrics we could use and verify if we are meeting certain SLAs

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.aws.greengrass.clientdevices.auth.api;
+
+import com.aws.greengrass.dependency.Context;
+import com.aws.greengrass.util.CrashableFunction;
+
+
+public class UseCases {
+    private final Context context;
+    private static UseCases instance;
```
MikeDombo, line 14: I will caution you against statics/singleton pattern. It makes testing and dependency injection more of a hassle (if not impossible).

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    /**
+     * Allows to provide a specific dependency we want to be injected on the use case being tested.
+     *
+     * @param clazz Class of the instance you want to inject1
+     * @param ob Concrete instance of the class
+     */
+    public static UseCases provide(Class clazz, Object ob) {
```
MikeDombo, line 39: can be generic `<T> (Class<T> clazz, T obj)`

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,56 @@
...
+     * Wrapper around context that will generate a new instance of a Use Case with
+     * the latest dependencies on the context.
+     *
+     * @param clazz Use Case class to be built
+     * @param <C>   Use Case concrete class
+     */
+    public static <C extends UseCase> C get(Class<C> clazz) {
+        checkCanRun();
+        return instance.context.newInstance(clazz);
+    }
```
MikeDombo, line 55: if this class is just wrapping the Context, why is this useful?

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Wrapper around context that will generate a new instance of a Use Case with
+     * the latest dependencies on the context.
+     *
+     * @param clazz Use Case class to be built
+     * @param <C>   Use Case concrete class
+     */
+    public static <C extends UseCase> C get(Class<C> clazz) {
+        checkCanRun();
+        return instance.context.newInstance(clazz);
```
MikeDombo, line 54: Don't use newInstance. You should use `get` so that it will retrieve a pre-existing value (if any). It will create a new instance if needed.

jbutler: We want this to create a new instance. Our workflows grab immutable configuration wrappers. Without this, use cases won't be able to get new configuration objects when they're updated

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
jbutler: Ironically this actually makes it much easier to use DI throughout the rest of the code

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
Nelsonochoam: It basically hides how we get useCases we just happen to be wrapping the context for now. It also builds the useCases which is something other classes don't necessarily need to be exposed to

PR #120, opened 2022-09-09
refactor: split upload CA into separate use case

jbutler:
**Issue #, if available:**

**Description of changes:**
Splits CA upload into a separate use case so that it can be independently retried. For now this use case is called from within the original configure CA use case, but eventually it may be called independently (for example, if the network is down when we start).

Additionally, this also uses the new Nucleus API for retrieving the data client. This new API throws a checked exception for invalid DeviceConfiguration rather than an NPE, which will allow us to do something more intelligent later.

Unit tests.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -22,6 +22,7 @@ import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationExce
 import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
 import com.aws.greengrass.clientdevices.auth.exception.InvalidCertificateAuthorityException;
 import com.aws.greengrass.clientdevices.auth.exception.InvalidConfigurationException;
+import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.security.SecurityService;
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -320,12 +322,16 @@ public void uploadCoreDeviceCAs(String thingName) throws CertificateEncodingExce
...
         try {
             RetryUtils.runWithRetry(retryConfig,
-                    () -> clientFactory.getGreengrassV2DataClient().putCertificateAuthorities(request),
+                    () -> clientFactory.fetchGreengrassV2DataClient().putCertificateAuthorities(request),
                     "put-core-ca-certificate", logger);
         } catch (InterruptedException e) {
             logger.atInfo().log("Put core CA certificates got interrupted");
             // interrupt the current thread so that higher-level interrupt handlers can take care of it
             Thread.currentThread().interrupt();
+        } catch (DeviceConfigurationException e) {
```
MikeDombo, line 331: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: An exception is caught and re-thrown within the catch clause without any other operations being performed on the exception, making the catch clause unnecessary.

**Fix**: Perform additional operations on the exception that has been caught. You can wrap a standard exception into a custom exception to add additional information or special handling. When you do that, make sure to set the original exception as the cause. Otherwise, you will lose the stack trace and message of the original exception, which will make it difficult to analyze the event that caused the exception.  
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

Nelsonochoam (approved):

jcosentino11 (approved):

PR #121, opened 2022-09-11
feat: introduce basic domain events

jbutler:
**Issue #, if available:**

**Description of changes:**
Introduces a basic domain event emitter which allows for domain events to be emitted, and for listeners to subscribe to events. 

This change also introduces a CACertificateChainChanged event which is used to decouple CA configuration from CA registration.

This is needed to de-couple use cases which otherwise would not need to know about each other.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.api;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+
+public class DomainEvents {
+    private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventMap = new ConcurrentHashMap<>();
+
+    @FunctionalInterface
+    public interface DomainEventListener<T> {
+        void handle(T event);
+    }
+
+    /**
+     * Register event listener.
+     * @param listener Event listener callback
+     * @param clazz    Type of domain event
+     * @param <T>      Type of domain event
+     */
+    public <T> void registerListener(DomainEventListener<T> listener, Class<T> clazz) {
+        CopyOnWriteArrayList<DomainEventListener> listenerList =
+                eventMap.computeIfAbsent(clazz, (k) -> new CopyOnWriteArrayList<>());
+        listenerList.addIfAbsent(listener);
+    }
+
+    /**
+     * Emit an event.
+     * @param domainEvent Domain event
+     * @param <T>         Type of domain event
+     */
+    public <T> void emit(T domainEvent) {
+        List<DomainEventListener> listenerList = eventMap.get(domainEvent.getClass());
+        if (listenerList != null) {
+            for (DomainEventListener<T> listener : listenerList) {
+                listener.handle(domainEvent);
+            }
+        }
+    }
+}
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
@@ -11,6 +11,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 
+@SuppressWarnings({"rawtypes","unchecked"})
 public class DomainEvents {
     private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventListeners = new ConcurrentHashMap<>();
 
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
@@ -12,7 +12,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 
 
 public class DomainEvents {
-    private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventMap = new ConcurrentHashMap<>();
+    private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventListeners = new ConcurrentHashMap<>();
 
     @FunctionalInterface
     public interface DomainEventListener<T> {
@@ -26,9 +26,9 @@ public class DomainEvents {
      * @param <T>      Type of domain event
      */
     public <T> void registerListener(DomainEventListener<T> listener, Class<T> clazz) {
-        CopyOnWriteArrayList<DomainEventListener> listenerList =
-                eventMap.computeIfAbsent(clazz, (k) -> new CopyOnWriteArrayList<>());
-        listenerList.addIfAbsent(listener);
+        CopyOnWriteArrayList<DomainEventListener> listeners =
+                eventListeners.computeIfAbsent(clazz, (k) -> new CopyOnWriteArrayList<>());
+        listeners.addIfAbsent(listener);
     }
 
     /**
@@ -37,9 +37,9 @@ public class DomainEvents {
      * @param <T>         Type of domain event
      */
     public <T> void emit(T domainEvent) {
-        List<DomainEventListener> listenerList = eventMap.get(domainEvent.getClass());
-        if (listenerList != null) {
-            for (DomainEventListener<T> listener : listenerList) {
+        List<DomainEventListener> listeners = eventListeners.get(domainEvent.getClass());
+        if (listeners != null) {
+            for (DomainEventListener<T> listener : listeners) {
                 listener.handle(domainEvent);
             }
         }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
@@ -25,7 +25,7 @@ public class DomainEvents {
      * @param clazz    Type of domain event
      * @param <T>      Type of domain event
      */
-    public <T> void registerListener(DomainEventListener<T> listener, Class<T> clazz) {
+    public <T extends DomainEvent> void registerListener(DomainEventListener<T> listener, Class<T> clazz) {
         CopyOnWriteArrayList<DomainEventListener> listeners =
                 eventListeners.computeIfAbsent(clazz, (k) -> new CopyOnWriteArrayList<>());
         listeners.addIfAbsent(listener);
@@ -36,7 +36,7 @@ public class DomainEvents {
      * @param domainEvent Domain event
      * @param <T>         Type of domain event
      */
-    public <T> void emit(T domainEvent) {
+    public <T extends DomainEvent> void emit(T domainEvent) {
         List<DomainEventListener> listeners = eventListeners.get(domainEvent.getClass());
         if (listeners != null) {
             for (DomainEventListener<T> listener : listeners) {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
@@ -16,7 +16,7 @@ public class DomainEvents {
     private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventListeners = new ConcurrentHashMap<>();
 
     @FunctionalInterface
-    public interface DomainEventListener<T> {
+    public interface DomainEventListener<T extends DomainEvent> {
         void handle(T event);
     }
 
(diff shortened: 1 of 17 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
```diff
@@ -0,0 +1,51 @@
...
+     * Trigger certificate authority registration use case.
+     *
+     * @param event Certificate authority change event
+     */
+    public void triggerCertificateAuthorityRegistrationUseCase(ConfiguredCertificateAuthorityEvent event) {
+        RegisterCertificateAuthorityUseCase useCase = UseCases.get(RegisterCertificateAuthorityUseCase.class);
+        try {
+            useCase.apply(null);
+        } catch (UseCaseException e) {
+            // TODO: Move retry logic from the domain to here
```
jbutler, line 48: Need to sort out logging

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
```diff
@@ -0,0 +1,47 @@
...
+     * Emit an event.
+     * @param domainEvent Domain event
+     * @param <T>         Type of domain event
+     */
+    public <T> void emit(T domainEvent) {
+        List<DomainEventListener> listenerList = eventMap.get(domainEvent.getClass());
+        if (listenerList != null) {
+            for (DomainEventListener<T> listener : listenerList) {
+                listener.handle(domainEvent);
+            }
```
jbutler, line 44: Note: in the future we may want to add in a work queue here instead of dispatching events synchronously

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
+import com.aws.greengrass.clientdevices.auth.api.UseCases;
+import com.aws.greengrass.clientdevices.auth.certificate.events.ConfiguredCertificateAuthorityEvent;
+import com.aws.greengrass.clientdevices.auth.certificate.usecases.RegisterCertificateAuthorityUseCase;
+import com.aws.greengrass.clientdevices.auth.exception.UseCaseException;
+
+import javax.inject.Inject;
+
+public class ConfiguredCertificateAuthorityEventListener {
```
Nelsonochoam, line 16: Could we use the `DomainEventListener` interface?. Also I am really happy you added a Listener class which can serve as a controller to redirect to the proper use case based on the event contents

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
```diff
@@ -0,0 +1,51 @@
+/*
```
Nelsonochoam, line 1: Could we place the listeners on a package inside the domain they belong to?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/events/ConfiguredCertificateAuthorityEvent.java
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.certificate.events;
+
+import lombok.Getter;
+
+import java.security.cert.Certificate;
+
+public class ConfiguredCertificateAuthorityEvent {
```
Nelsonochoam, line 12: Could we have a `DomainEvent` interface?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -80,13 +83,15 @@ public enum CAType {
...
 
     @Inject
-    public CertificateStore(Kernel kernel) throws IOException {
-        this.workPath = kernel.getNucleusPaths().workPath(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME);
+    public CertificateStore(Kernel kernel, DomainEvents eventEmitter) throws IOException {
+        this(kernel.getNucleusPaths().workPath(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME),
+                eventEmitter);
     }
 
     // For unit tests
```
Nelsonochoam, line 91: This is kind of bad, just having an API for unit tests

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -139,6 +144,15 @@ public X509Certificate getCACertificate() throws KeyStoreException {
...
     }
 
+    /**
+     * Sets the CA certificate chain.
+     *
+     * @param caCertificateChain Array of CA certificates
+     */
+    public void setCaCertificateChain(Certificate... caCertificateChain) {
+        this.caCertificateChain = caCertificateChain;
+        eventEmitter.emit(new ConfiguredCertificateAuthorityEvent(caCertificateChain));
```
Nelsonochoam, line 154: This is beautiful

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
```diff
@@ -0,0 +1,51 @@
...
+                this::triggerCertificateAuthorityRegistrationUseCase,
+                ConfiguredCertificateAuthorityEvent.class);
+    }
+
+    /**
+     * Trigger certificate authority registration use case.
+     *
+     * @param event Certificate authority change event
+     */
+    public void triggerCertificateAuthorityRegistrationUseCase(ConfiguredCertificateAuthorityEvent event) {
```
Nelsonochoam, line 43: `handle`?

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
```diff
@@ -0,0 +1,47 @@
...
+package com.aws.greengrass.clientdevices.auth.api;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+
+public class DomainEvents {
+    private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventMap = new ConcurrentHashMap<>();
```
Nelsonochoam, line 15:
```suggestion
    private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> subscribers = new ConcurrentHashMap<>();
```

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
Nelsonochoam: Dispatching synchronously has its benefits. I think we should have an emitAsync maybe besides the `emit`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
jbutler: Agree, and I've left similar comments in the past. Though in this case it's a pain to set up the kernel object to return the right work path :-/

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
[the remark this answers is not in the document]
jbutler: Definitely agree. We can probably get fancy in the future and make these listener classes pretty small

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
jbutler: Happy to make structural changes. Just kinda threw this in a listeners package. But this is still certificate related. Are you suggesting we break CA registration into a separate domain?

Nelsonochoam: This was my bad. My brain is fried

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
Nelsonochoam: The event name is not very expressive of what we are doing. Maybe `CACertificateChainChanged`.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/ConfiguredCertificateAuthorityEventListener.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
+import com.aws.greengrass.clientdevices.auth.api.UseCases;
+import com.aws.greengrass.clientdevices.auth.certificate.events.ConfiguredCertificateAuthorityEvent;
+import com.aws.greengrass.clientdevices.auth.certificate.usecases.RegisterCertificateAuthorityUseCase;
+import com.aws.greengrass.clientdevices.auth.exception.UseCaseException;
+
+import javax.inject.Inject;
+
+public class ConfiguredCertificateAuthorityEventListener
```
Nelsonochoam, line 16: Similar to the comment above maybe this should be `
```
public class CACertificateChainChangedListener implements DomainEventListener<CACertificateChainChanged> {
 ....
}
```

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+
+@SuppressWarnings({"rawtypes","unchecked"})
+public class DomainEvents {
+    private final Map<Class, CopyOnWriteArrayList<DomainEventListener>> eventListeners = new ConcurrentHashMap<>();
+
+    @FunctionalInterface
+    public interface DomainEventListener<T> {
```
Nelsonochoam, line 19:
```suggestion
    public interface DomainEventListener<T extends DomainEvent> {
```

src/test/java/com/aws/greengrass/clientdevices/auth/api/DomainEventsTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    void GIVEN_noListener_WHEN_eventEmitted_THEN_itsFine() {
+        domainEvents.emit(new TestEvent());
+        domainEvents.emit(new TestEvent2());
+    }
+
+    @Test
+    void GIVEN_twoListenersForSameEvent_WHEN_eventEmitted_THEN_bothListenersFire() {
+        AtomicInteger listener1Count = new AtomicInteger(0);
```
Nelsonochoam, line 40: Do we need to use an `AtomicInteger` for tests? I get it if we were using multiple threads on the test

jbutler: This isn't to avoid a race. You can only reference effectively final variables in a Lambda function, so this can't be a regular `int`.

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
jbutler: Good catch. Missed this one

Nelsonochoam (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param domainEvent Domain event
+     * @param <T>         Type of domain event
+     */
+    public <T extends DomainEvent> void emit(T domainEvent) {
+        List<DomainEventListener> listeners = eventListeners.get(domainEvent.getClass());
+        if (listeners != null) {
+            for (DomainEventListener<T> listener : listeners) {
+                listener.handle(domainEvent);
+            }
+        }
```
vaibhavmurkute, line 46: Method name `emit()` sounds bit misleading here as it inherently handles the emitted event. May be rename to `trigger()`?

jbutler: How it's handled is an implementation detail. The caller does not know about what (if anything) happens. It is simply emitting an event

vaibhavmurkute (dismissed):

jbutler:
Overriding based on previous approvals. #122 resulted in a merge conflict which was trivially resolved

PR #122, opened 2022-09-12
feat: call configure ca every time the certificateAuthority config changes

Nelsonochoam:
**Issue #, if available:**

**Description of changes:**
Ensure we are calling the use case to configure the certificate authority every time there is a change on the CDA configuration for the `certificateAuthority`.

Additional changes:
* Made the UseCases injectable instead of a singleton due to the fact it make them easier to test by allowing us to mock provided values of use cases in case we want to do simple unit tests.

We need to make sure we are updating the CAs when any value under the certificateAuthority configuration changes

Wrote a unit test.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
@@ -6,26 +6,26 @@
 package com.aws.greengrass.clientdevices.auth.api;
 
 import com.aws.greengrass.dependency.Context;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.CrashableFunction;
 
 
 public class UseCases {
-    private final Context context;
-    private static UseCases instance;
+    private Context context;
+    public static final Logger logger = LogManager.getLogger(UseCases.class);
 
     // Delegates to CrashableFunction but provides a domain rich alias
     public interface UseCase<R, D, E extends Exception> extends CrashableFunction<D, R, E> {}
 
-    private UseCases(Context context) {
+    public void init(Context context) {
         this.context = context;
     }
 
-    public static void init(Context context) {
-       instance = new UseCases(context);
-    }
 
-    private static void checkCanRun() {
-        if (instance == null) {
+
+    private void checkCanRun() {
+        if (context == null) {
             throw new RuntimeException("No UseCases instance found, make sure you initialize them first");
         }
     }
@@ -37,10 +37,10 @@ public class UseCases {
      * @param ob Concrete instance of the class
      * @param <T> instance type
      */
-    public static <T> UseCases provide(Class<T> clazz, T ob) {
+    public <T> UseCases provide(Class<T> clazz, T ob) {
         checkCanRun();
-        instance.context.put(clazz, ob);
-        return instance;
+        context.put(clazz, ob);
+        return this;
     }
 
     /**
@@ -50,8 +50,8 @@ public class UseCases {
      * @param clazz Use Case class to be built
      * @param <C>   Use Case concrete class
      */
-    public static <C extends UseCase> C get(Class<C> clazz) {
+    public <C extends UseCase> C get(Class<C> clazz) {
         checkCanRun();
-        return instance.context.newInstance(clazz);
+        return context.newInstance(clazz);
     }
 }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
@@ -6,14 +6,11 @@
 package com.aws.greengrass.clientdevices.auth.api;
 
 import com.aws.greengrass.dependency.Context;
-import com.aws.greengrass.logging.api.Logger;
-import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.CrashableFunction;
 
 
 public class UseCases {
     private Context context;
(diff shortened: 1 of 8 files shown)


src/test/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthServiceTest.java
```diff
@@ -343,12 +350,48 @@ void GIVEN_GG_with_cda_WHEN_updated_ca_type_THEN_ca_is_updated()
...
         List<List<String>> certificatesInRequests =
                 putCARequestArgumentCaptor.getAllValues().stream().map(
                         PutCertificateAuthoritiesRequest::coreDeviceCertificates).collect(
                         Collectors.toList());
-        assertThat(certificatesInRequests, contains(initialCACerts, secondCACerts, thirdCACerts));
+        assertThat(certificatesInRequests, contains(initialCACerts, thirdCACerts));
+    }
+
+    @Test
+    void GIVEN_certificateAuthorityConfiguration_WHEN_itChanges_THEN_CAisConfigured() throws InterruptedException,
```
Nelsonochoam, line 362: This is the new test that checks that we are calling the use case when the certificateAuthority configuration changes.

MikeDombo:

src/test/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthServiceTest.java
```diff
@@ -343,12 +350,48 @@ void GIVEN_GG_with_cda_WHEN_updated_ca_type_THEN_ca_is_updated()
         assertThat(initialCA, not(thirdCA));
         assertThat(getCaPassphrase(), not(initialCaPassPhrase));
 
-        verify(client, times(3)).putCertificateAuthorities(putCARequestArgumentCaptor.capture());
+        verify(client, times(2)).putCertificateAuthorities(putCARequestArgumentCaptor.capture());
```
Nelsonochoam, line 353: This is not a mistake, before this changes we were unnecessarily configuring CAs 1 time. When no `ca_type` is provided we default to use RSA and if we then change the configuration  to RSA, the config is effectively the same. The CA types is still RSA so we don't need to reconfigure them again.

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -6,26 +6,26 @@
...
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.CrashableFunction;
 
 
 public class UseCases {
-    private final Context context;
-    private static UseCases instance;
+    private Context context;
+    public static final Logger logger = LogManager.getLogger(UseCases.class);
```
MikeDombo, line 16: public seems wrong

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/ConfigureCertificateAuthorityUseCase.java
```diff
@@ -40,8 +44,10 @@ public Void apply(Void unused) throws UseCaseException {
         //  the ClientDeviceAuthService first.
         try {
             if (cdaConfiguration.isUsingCustomCA()) {
+                UseCases.logger.info("Configuration custom certificate authority");
```
MikeDombo, line 47: why doesn't this class have its own logger?

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CDAConfiguration.java
```diff
@@ -98,4 +98,20 @@ public Optional<URI> getPrivateKeyUri() {
...
+     * configuration.
+     *
+     * @param config  CDAConfiguration
+     */
+    public boolean certificateAuthorityChanged(CDAConfiguration config) {
+        if (config == null) {
+            return true;
+        }
+
+        return !config.getCertificateUri().equals(getCertificateUri())
```
MikeDombo, line 113: You can use `Objects.equals` to handle nulls better

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
Nelsonochoam: I didn't wan't to create a logger per usecase, but it doesn't seem that bad to do it.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CDAConfiguration.java
Nelsonochoam: Will do

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CDAConfiguration.java
```diff
@@ -98,4 +99,20 @@ public Optional<URI> getPrivateKeyUri() {
...
         return ca.getCertificateUri();
     }
+
+    /**
+     * Verifies if the configuration for the certificateAuthority has changed, given a previous
+     * configuration.
+     *
+     * @param config  CDAConfiguration
+     */
+    public boolean certificateAuthorityChanged(CDAConfiguration config) {
```
jbutler, line 109: `equals` ?

Nelsonochoam: I like the `certificateAuthorityChanged` equals might be good if we are comparing the whole config.

jbutler (approved):

PR #125, opened 2022-09-13
fix: present full cert chain

jbutler:
This change adds support for returning the entire CA certificate chain to clients who subscribe to certificate updates. This change is required in order to support Greengrass using an intermediate CA to issue server certificates.

**Issue #, if available:**

**Description of changes:**

Unit tests. Manually testing with MQTT broker to ensure that the full certificate chain is returned as part of the SubscribeToCertificateUpdates operations.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
@@ -17,7 +17,6 @@ import com.aws.greengrass.util.FileSystemPermission;
 import com.aws.greengrass.util.platforms.Platform;
 import lombok.AccessLevel;
 import lombok.Getter;
-import lombok.Setter;
 import org.bouncycastle.cert.CertIOException;
 import org.bouncycastle.operator.OperatorCreationException;
 
@@ -43,6 +42,7 @@ import java.security.interfaces.ECPrivateKey;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.spec.ECGenParameterSpec;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.Date;
 import javax.inject.Inject;
 
@@ -74,8 +74,7 @@ public class CertificateStore {
     private final DomainEvents eventEmitter;
 
     @Getter
-    @Setter
-    private Certificate[] caCertificateChain;
+    private X509Certificate[] caCertificateChain;
     private PrivateKey caPrivateKey;
 
     public enum CAType {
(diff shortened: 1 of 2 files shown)


Nelsonochoam (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -142,15 +141,24 @@ public PrivateKey getCAPrivateKey() {
      * @throws KeyStoreException if unable to retrieve the certificate
      */
     public X509Certificate getCACertificate() throws KeyStoreException {
```
jcosentino11, line 143:
```suggestion
    public X509Certificate getCACertificate() {
```
same w/  javadoc

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -142,15 +141,24 @@ public PrivateKey getCAPrivateKey() {
...
-        return (X509Certificate) getCaCertificateChain()[0];
+        return getCaCertificateChain()[0];
```
jcosentino11, line 144: do we want to handle if `getCaCertificateChain()` is null?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -142,15 +141,24 @@ public PrivateKey getCAPrivateKey() {
...
+    }
+
+    private void setCaCertificateChain(Certificate... caCertificateChain) throws KeyStoreException {
```
jcosentino11, line 147: do we want to handle case if chain is null?  probably not happening in practice but may be good to have for completeness

Nelsonochoam: It would be nice if the type system could tell it we don't accept null values, would that @NotNull annotation work?

jbutler: It does, but it just auto-generates the code that throws a NPE. I think in this case we prefer a checked exception. Will add

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
jbutler: I actually went the other direction. Downstream consumers are catching KeyStoreException and then passing straight through to methods that are annotated with @NotNull.
Maybe there's something we can do in the future to prevent this from ever getting into a bad state. But for now, this is a more surgical change.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -142,15 +141,35 @@ public PrivateKey getCAPrivateKey() {
...
+        if (caCertificateChain == null) {
+            throw new KeyStoreException("No certificate chain provided");
+        }
+        for (Certificate cert : caCertificateChain) {
+            if (!(cert instanceof X509Certificate)) {
+                throw new KeyStoreException("Unsupported certificate type");
+            }
+        }
+        setCaCertificateChain(Arrays.stream(caCertificateChain).toArray(X509Certificate[]::new));
```
MikeDombo, line 160: caCertificateChain is already an array, `...` is an array. Why not just cast it?

jbutler: Class cast exception. The objects are x509 certs but the array is not.

jcosentino11 (approved):

Nelsonochoam (approved):

PR #126, opened 2022-09-14
refactor: fine grained use cases based on events

Nelsonochoam:
**Description of changes:**
This is just part of the changes we have introduced to decouple logic. This changes remove the calls of use cases directly from the ClientDeviceAuthService and created a listener for domain configuration changes.

Listeners acts as a control layer redirecting events to the appropriate use cases.

Continue decoupling application logic


MikeDombo:

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -195,22 +187,18 @@ private void configChangeHandler(WhatHappened whatHappened, Node node) {
...
-                updateDeviceGroups(whatHappened, deviceGroupTopics);
-            } else if (cdaConfiguration.hasCAConfigurationChanged(prevCdaConfiguration)) {
-                useCases.get(ConfigureCertificateAuthorityUseCase.class).apply(null);
-            }
-        } catch (UseCaseException e) {
-            serviceErrored(e);
+        if (whatHappened == WhatHappened.initialized || node == null) {
+            updateDeviceGroups(whatHappened, deviceGroupTopics);
+        } else if (node.childOf(DEVICE_GROUPS_TOPICS)) {
+            updateDeviceGroups(whatHappened, deviceGroupTopics);
```
jbutler, line 193: nit: merge?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/listeners/CAConfigurationChangedListener.java
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param event Certificate authority configuration change event
+     */
+    @Override
+    public void handle(CAConfigurationChanged event)  {
+        CAConfiguration caConfiguration = event.getConfiguration();
+
+        try {
+            if (caConfiguration.isUsingCustomCA()) {
+                useCases.get(ConfigureCustomCertificateAuthority.class).apply(null);
```
jbutler, line 52: Hmmm... I know we did all that fancy stuff in UseCases such that they would always get the most up to date config, but now I'm wondering if we should be passing it in.
As written, there is a potential race here (e.g. if config switches from custom to managed after this event is generated). It should sort itself out, but we'd need to be ready for CA config to fail

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/ConfigureManagedCertificateAuthority.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public Void apply(Void unused) throws UseCaseException {
+        logger.info("Configuring Greengrass managed certificate authority.");
+
+        try {
+            // NOTE: Update doesn't really reflect what we are doing we and the store should just store objects
+            // not have the logic of how to create them
+            certificateManager.generateCA(configuration.getCaPassphrase(), configuration.getCaType());
+            configuration.updateCAPassphrase(certificateManager.getCaPassPhrase());
+            configuration.updateCACertificates(certificateManager.getCACertificates());
```
jbutler, line 50: This is race-y. It shouldn't be possible to interleave CA configuration and runtime config updates. Can we synchronize this block? (I believe you can sync on the `certificateManager` object)

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
jbutler: That's fine

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/ConfigureManagedCertificateAuthority.java
jbutler: Discussed offline. Synchronizing here actually doesn't solve the issue. We need to ensure that configuration updates are processed in the correct order. For now, we'll rely on single threaded execution and document this restriction as one that needs to be addressed before we can support a multi-threaded work queue

src/test/java/com/aws/greengrass/clientdevices/auth/api/UseCasesTest.java
```diff
@@ -102,33 +99,12 @@ void GIVEN_aUseCaseWithDependencies_WHEN_ran_THEN_itExecutesWithNoExceptions() {
...
-        assertThrows(InvalidConfigurationException.class, () -> { useCase.apply(null); });
+        assertThrows(UseCaseException.class, () -> { useCase.apply(null); });
     }
 
     @Test
     void GIVEN_aUseCaseWithParameters_WHEN_ran_itAcceptsTheParamsAndReturnsThem() {
         UseCaseWithParameters useCase = useCases.get(UseCaseWithParameters.class);
         assertEquals(useCase.apply("hello"), "hello");
     }
-
```
Nelsonochoam, line 113: Removed this since we are not providing the configuration like this anymore. We determined it would do more harm than good if/when we introduced threading to process use cases.

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -26,27 +27,13 @@ private void checkCanRun() {
     }
 
     /**
-     * Allows to provide a specific dependency we want to be injected on the use case being tested.
```
Nelsonochoam, line 29: Removing since we are not using this anymore. We determined that doing this might lead to issues when running use cases using multiple threads

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -195,22 +187,16 @@ private void configChangeHandler(WhatHappened whatHappened, Node node) {
...
-                updateDeviceGroups(whatHappened, deviceGroupTopics);
-                useCases.get(ConfigureCertificateAuthorityUseCase.class).apply(null);
-            } else if (node.childOf(DEVICE_GROUPS_TOPICS)) {
-                updateDeviceGroups(whatHappened, deviceGroupTopics);
-            } else if (cdaConfiguration.hasCAConfigurationChanged(prevCdaConfiguration)) {
-                useCases.get(ConfigureCertificateAuthorityUseCase.class).apply(null);
-            }
-        } catch (UseCaseException e) {
-            serviceErrored(e);
+        if (whatHappened == WhatHappened.initialized || node == null || node.childOf(DEVICE_GROUPS_TOPICS)) {
```
Nelsonochoam, line 190: More stuff got removed from the service 🔥

jbutler (approved):

MikeDombo (approved):

PR #127, opened 2022-09-16
feat: return results from use cases

Nelsonochoam:
**Description of changes:**
This code adds a Result class meant to be used to represent the state of an operation after completion. The main idea is that use cases will return some result state to the upper layer calling them and the upper layer can decide what to do with the result.

A result doesn't necessarily has to be generated from us catching an exception. If we have an operation or a use case that validates something and the validation failed we could return an error or warning indicating that the operation failed. It gives us a better way of representing result without having interpret things like

* If a method returns null what does that mean? Was it success, failure?
* Using try catch and re raising exceptions for everything

It also adheres better to the functional paradigm. The result class is actually a Functor and we could technically expand it to have methods like `map` which would only execute an operation if there is a valid result. This leads to code that is more declarative and also remove the needs of additional `ifelse` logic.

**Why is this change needed?**
Before use cases were returning null values when they ran properly or doing a weird try catch to catch an exception and then re-throwing it as another exception type `UseCaseException`.

This was just a refactor, make sure existing tests passed


src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.api;
+
+
+public final class Result<T> {
```
Nelsonochoam, line 9: I don't know if java have a construct like this but the benefit of having this value object is that now we can better represent invalid errors states and decide to react to them later. A few advantages of using it are:

1) No more scenarios likeDo I throw an error and let the client figure out how to handle it? Do I return null?
2) Failures can be represented in terms of the domain

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -137,27 +142,36 @@ private int getValidCloudCallQueueSize(Topics topics) {
...
     }
 
-
     @Override
     protected void install() throws InterruptedException {
         super.install();
         registerEventListeners();
-        subscribeToConfigChanges();
+        onConfigurationChanged();
+        registerSubscribers();
```
jbutler, line 150: Does this work? There won't be subscribers for the config change event

src/main/java/com/aws/greengrass/clientdevices/auth/api/DomainEvents.java
```diff
@@ -5,19 +5,34 @@
...
+    }
 
     @FunctionalInterface
     public interface DomainEventListener<T extends DomainEvent> {
-        void handle(T event);
+        Result<?> handle(T event);
+    }
+
+    public void onError(Consumer<Result<? extends Exception>> consumer) {
+        this.errorHandler = consumer;
```
jbutler, line 35: There are probably better ways to signal back to the service that an error has occurred. This requires the failure to have been triggered by an event, which I don't think will always be the case. Why not just introduce a critical failure event that the service can subscribe to?

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
Nelsonochoam: Yes, I know it seems fishy but that is how was working before. Before starting the listeners we need to call this so that it boots up the configuration with some initial values. If you check what we had before the ` onConfigurationChanged()` was before inside of `subscribeToConfigChanges` which got renamed to ` registerSubscribers` and was being called before the call to ` config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe(this::configChangeHandler);`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/RegisterCertificateAuthorityUseCase.java
```diff
@@ -21,7 +21,7 @@
 import javax.inject.Inject;
 
 public class RegisterCertificateAuthorityUseCase
-        implements UseCases.UseCase<Void, Void> {
+        implements UseCases.UseCase<Exception, Void> {
```
jbutler, line 24: Interesting take. Maybe we can wrap these so exceptions never get thrown from use cases (but without requiring every use case to catch Exception). Something for later

jbutler (approved):

jbutler (approved):

jbutler (approved):

vaibhavmurkute (approved):

PR #130, opened 2022-09-20
refactor: simplify CDA service entry

jbutler:
**Issue #, if available:**

**Description of changes:**
Stop injecting everything into CDA and just grab it from the context when it's needed.

More of this will eventually go away once we introduce more UseCases. Infrastructure init will also move, along with configuration.

It's not. Just cleans things up a bit.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
@@ -51,33 +51,23 @@ import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.GET
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.SUBSCRIBE_TO_CERTIFICATE_UPDATES;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.VERIFY_CLIENT_DEVICE_IDENTITY;
 
-@SuppressWarnings("PMD.DataClass")
 @ImplementsService(name = ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME)
 public class ClientDevicesAuthService extends PluginService {
     public static final String CLIENT_DEVICES_AUTH_SERVICE_NAME = "aws.greengrass.clientdevices.Auth";
+
+    // TODO: Move configuration related constants to appropriate configuration class
     public static final String DEVICE_GROUPS_TOPICS = "deviceGroups";
     public static final String PERFORMANCE_TOPIC = "performance";
     public static final String MAX_ACTIVE_AUTH_TOKENS_TOPIC = "maxActiveAuthTokens";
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -51,81 +52,50 @@
...
-
-    private final ClientDevicesAuthServiceApi clientDevicesAuthServiceApi;
-    private final AuthorizationHandler authorizationHandler;
-    private final GreengrassCoreIPCService greengrassCoreIPCService;
     // Limit the queue size before we start rejecting requests
     private static final int DEFAULT_CLOUD_CALL_QUEUE_SIZE = 100;
     private static final int DEFAULT_THREAD_POOL_SIZE = 1;
     public static final int DEFAULT_MAX_ACTIVE_AUTH_TOKENS = 2500;
+
+    private final Context context;
```
MikeDombo, line 70: you already have a context instance, you don't need this. It is in your parent class

MikeDombo: `getContext()` will give it to you.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -51,81 +52,50 @@
...
-        this.useCases.init(getContext());
+        this.context = topics.getContext();
     }
 
+    @Override
+    protected void install() throws InterruptedException {
+        super.install();
+
+        context.get(UseCases.class).init(context);
+        context.get(CertificateManager.class).updateCertificatesConfiguration(new CertificatesConfig(getConfig()));
```
MikeDombo, line 94: getting stuff from the context may fail with a runtime exception if the class (or a class that is needed by that class) cannot be constructed. Maybe this can't be an issue, but just FYI

Nelsonochoam (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
jbutler: Actually I think I'll keep this since it's referenced in quite a few places

MikeDombo: It won't ever be private. `protected`, yeah, but not private.

Just delete your field and then keep using `context` which will access the one in the parent class

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
MikeDombo: Yes, though the results will be a bit different. If it failed in your constructor, then your service would be replaced with `UnloadableService`. Whereas here, it will fail while starting which will end with the service in the broken state.

Nelsonochoam (approved):

PR #131, opened 2022-09-20
feat: upload the ca with the highest trust level from the ca chain

Nelsonochoam:
**Description of changes:**
This CR changes what CA is uploaded to Iot core so that validation of client devices work when providing a certificate chain. Customers having clients using V1 of the IoT device SDK must provide the full certificate chain to CDA in order to be successfully authenticated. V2 is a bit more lenient, it allow clients to be validated with partial chains (meaning not having a core CA present and using and intermediate as a Trust anchor) 

We need this change so that customers bringing their own CA with clients on V1 of the SDK can continue to successfully authenticate their devices using the CA returned from cloud discovery.

Manually tested and validate that authentication would work using openssl by creating a certificate chain using openssl, providing a full chain and ensure it validates it on v1 and v2 and providing a partial chain and ensure it can connect on v2.

I also wrote a UAT that simulates this scenarios and authentication succeeds.

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
@@ -6,27 +6,23 @@
 package com.aws.greengrass.integrationtests.certificateauthority;
 
 import com.aws.greengrass.clientdevices.auth.CertificateManager;
-import com.aws.greengrass.clientdevices.auth.DeviceAuthClient;
+import com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService;
 import com.aws.greengrass.clientdevices.auth.api.CertificateUpdateEvent;
 import com.aws.greengrass.clientdevices.auth.api.ClientDevicesAuthServiceApi;
-import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
 import com.aws.greengrass.clientdevices.auth.api.GetCertificateRequest;
 import com.aws.greengrass.clientdevices.auth.api.GetCertificateRequestOptions;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificateExpiryMonitor;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificateHelper;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificateStore;
-import com.aws.greengrass.clientdevices.auth.certificate.CertificatesConfig;
-import com.aws.greengrass.clientdevices.auth.configuration.CDAConfiguration;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
-import com.aws.greengrass.clientdevices.auth.connectivity.CISShadowMonitor;
-import com.aws.greengrass.clientdevices.auth.connectivity.ConnectivityInformation;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
-import com.aws.greengrass.clientdevices.auth.exception.InvalidConfigurationException;
 import com.aws.greengrass.clientdevices.auth.helpers.CertificateTestHelpers;
-import com.aws.greengrass.clientdevices.auth.iot.registry.CertificateRegistry;
-import com.aws.greengrass.clientdevices.auth.session.SessionManager;
 import com.aws.greengrass.config.Topics;
-import com.aws.greengrass.dependency.Context;
+import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
+import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
+import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
 import com.aws.greengrass.security.SecurityService;
 import com.aws.greengrass.security.exceptions.CertificateChainLoadingException;
 import com.aws.greengrass.security.exceptions.KeyLoadingException;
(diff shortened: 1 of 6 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -36,117 +32,137 @@
...
-    private CertificateManager certificateManager;
-    private CertificateStore certificateStore;
+    Path rootDir;
+    private Kernel kernel;
 
 
     @BeforeEach
-    void beforeEach() {
-        DomainEvents events = new DomainEvents();
-        certificateStore = new CertificateStore(tmpPath, events);
```
Nelsonochoam, line 106: I refactored this tests so they are more integration like instead of calling direct units.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/RegisterCertificateAuthorityUseCase.java
```diff
@@ -48,9 +51,13 @@ public Result apply(Void unused)  {
         String thingName = Coerce.toString(deviceConfiguration.getThingName());
 
         try {
+            X509Certificate[] caChain = certificateManager.getX509CACertificates();
```
Nelsonochoam, line 54: This is the important bit

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/CertificateTestHelpers.java
```diff
@@ -51,46 +69,112 @@ private CertificateTestHelpers() {
         Security.addProvider(new BouncyCastleProvider());
     }
 
+    private enum CertificateTypes {
```
Nelsonochoam, line 72: All this fuzz here is just adding helpers to create correctly generated root cas, intermediate cas, and server certificates. You could verify it generates correct certs using `openssl verify`

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -36,117 +32,137 @@
...
-                mockCertExpiryMonitor, mockShadowMonitor, Clock.systemUTC(), clientFactoryMock, securityServiceMock);
+    @AfterEach
+    void cleanup() {
+        kernel.shutdown();
+    }
 
+    // TODO: Consolidate this test helpers with ClientDevicesAuthServiceTest
+    private void startNucleus() throws InterruptedException {
+        CountDownLatch authServiceRunning = new CountDownLatch(1);
+        Path resourceDirectory = Paths.get(
```
MikeDombo, line 123: You can just use `<class>.getResourceURI`

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -298,15 +305,22 @@ public void configureCustomCA(CDAConfiguration configuration) throws InvalidConf
...
      * @throws DeviceConfigurationException   If unable to retrieve Greengrass V2 Data client
      */
     @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidRethrowingException"})
-    public void uploadCoreDeviceCAs(String thingName) throws CertificateEncodingException, KeyStoreException,
+    public void uploadCoreDeviceCAs(String thingName, List<X509Certificate> certificates)
+            throws CertificateEncodingException, KeyStoreException,
             IOException, DeviceConfigurationException {
-        List<String> certificatePemList = getCACertificates();
+        List<String> certificatePemList = new ArrayList<>();
+        for (X509Certificate cert: certificates) {
```
MikeDombo, line 320: nit: formatting

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/RegisterCertificateAuthorityUseCase.java
```diff
@@ -48,9 +51,13 @@ public Result apply(Void unused)  {
...
+            X509Certificate highestTrustCA = caChain[caChain.length - 1];
```
MikeDombo, line 55: nit: verify length first

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
Nelsonochoam: I tried that but it didn't work, got a NullPointerException, This test is inside the integrationtests folder. Is there something I am missing?. Basically `getClass().getResource(configFileName)` and didn't work

MikeDombo: Sounds like the yaml file is in the wrong place then. `integrationtests` should have its own `resources` directory.

Might also need this: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/c950a691d92f12204d447b021891253d756fd5ec/pom.xml#L328-L341

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -126,18 +127,24 @@ public void stopMonitors() {
...
         return Collections.singletonList(CertificateHelper.toPem(certificateStore.getCACertificate()));
     }
 
-    private X509Certificate[] getX509CACertificates() throws KeyStoreException {
+    /**
+     * Returns the full CA chain.
+     *
+     * @throws KeyStoreException when failing to read certificates from key store.
+     */
+    public X509Certificate[] getX509CACertificates() throws KeyStoreException {
```
jbutler, line 147: Why make changes to CertManager at this point? Can we not just directly call the CertStore?

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -298,15 +305,22 @@ public void configureCustomCA(CDAConfiguration configuration) throws InvalidConf
...
+     * @param certificates a list of X509Certificate to be uploaded to IoTCore
+     *
      * @throws CertificateEncodingException   If unable to get certificate encoding
      * @throws KeyStoreException              If unable to retrieve the certificate
      * @throws IOException                    If unable to read certificate
      * @throws DeviceConfigurationException   If unable to retrieve Greengrass V2 Data client
      */
     @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidRethrowingException"})
-    public void uploadCoreDeviceCAs(String thingName) throws CertificateEncodingException, KeyStoreException,
+    public void uploadCoreDeviceCAs(String thingName, List<X509Certificate> certificates)
```
jbutler, line 316: The cloud only supports a single cert, right? Can we change this method signature to reflect that?

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
Nelsonochoam: I would have to change the useCase and add an additional dependency which means more changes. I was just focusing on the functionality. I can do it though

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
Nelsonochoam: I think we could upload multiple but for CDA we should not need to upload more than one so, I'll patch it

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -36,117 +32,137 @@
...
     }
 
-    /**
-     * Simulates an external party configuring their custom CA.
-     */
-    private X509Certificate[] arrangeCustomCertificateAuthority() throws NoSuchAlgorithmException, CertificateException,
-            OperatorCreationException, CertIOException, URISyntaxException, KeyLoadingException,
-            ServiceUnavailableException, CertificateChainLoadingException, InvalidConfigurationException {
-        // Generate custom certificates
+    private static Pair<X509Certificate[], KeyPair[]> arrangeCredentials() throws NoSuchAlgorithmException,
```
jbutler, line 137: I know you're basing this naming off that article you linked, but `arrangeCredentials` doesn't mean anything to me. IMO it's just a *really* terrible name. Can you make it clear what's actually happening? Rename to `generateCoreAndIntermediateCA` or something

src/integrationtests/resources/com/aws/greengrass/integrationtests/certificateauthority/config.yaml
```diff
@@ -0,0 +1,36 @@
+---
```
Nelsonochoam, line 1: Copy pasta of this resource from test resources to integration test resources

jbutler (approved):

MikeDombo (approved):

PR #132, opened 2022-09-21
feat: cda work queue (DNM)

jbutler:
**Issue #, if available:**

**Description of changes:**
[WIP] Introduces work queue as infrastructure. This can be injected to other classes and used to queue async jobs

This simplifies queuing of work that must happen in another thread context. It also enables scheduling jobs in the future.


src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/GetClientDeviceAuthTokenTest.java
```diff
@@ -156,6 +157,8 @@ void GIVEN_brokerWithValidCredentials_WHEN_GetClientDeviceAuthToken_THEN_returns
         }
     }
 
+    // TODO: Re-enable once infrastructure supports limiting queue size
+    @Disabled
```
jbutler, line 161: Need to sort out how this should work moving forward

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/InfrastructureConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+ * |    |---- performance:
+ * |          |---- cloudRequestQueueSize: "..."
+ * |          |---- maxConcurrentCloudRequests: [...]
+ * </p>
+ */
+@Value
+public final class InfrastructureConfiguration {
+    public static final int DEFAULT_WORK_QUEUE_DEPTH = 100;
+    public static final int DEFAULT_THREAD_POOL_SIZE = 1;
+
```
jbutler, line 28: I don't love this configuration - might make sense to re-organize a bit. But for now, we'll keep it how it is.

Note: I didn't want to change the existing SessionConfig, so this performance key is split between the two config classes. I don't love it, but I also wanted to keep this PR as contained as possible.

src/main/java/com/aws/greengrass/clientdevices/auth/infra/CDAExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public CDAExecutor(ThreadPoolExecutor executor) {
+        this.executor = executor;
+        this.executorQueue = executor.getQueue();
+    }
+
+    public void shutdown() {
+        executor.shutdown();
+    }
+
```
jbutler, line 37: I'll like add methods for scheduling tasks in the future. We'll need to replace the queue with a delay queue for that. That can come later

PR #134, opened 2022-09-21
Ensure latest CA are returned to subscribers when they change

Nelsonochoam:
**Description of changes:**
Fix issue where old CA's are getting returned if the ca chain changes

This is part of a larger fix to ensure that we are rotating the certificates when the CA configuration changes.  It is required so that client components get the latest CA values and can properly authenticate.

Wrote a test

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
@@ -50,7 +50,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Consumer;
+import java.util.function.BiConsumer;
 import javax.inject.Inject;
 
 public class CertificateManager {
@@ -138,17 +138,6 @@ public class CertificateManager {
         return Collections.singletonList(CertificateHelper.toPem(certificateStore.getCACertificate()));
     }
 
-    /**
-     * Returns the full CA chain.
-     *
-     * @throws KeyStoreException when failing to read certificates from key store.
-     *
-     * @deprecated start using the certificateStore.getCaPassphrase directly
-     */
-    private X509Certificate[] getX509CACertificates() throws KeyStoreException {
-        return certificateStore.getCaCertificateChain();
-    }
-
     public String getCaPassPhrase() {
         return certificateStore.getCaPassphrase();
     }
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -174,24 +163,23 @@ public void subscribeToCertificateUpdates(GetCertificateRequest getCertificateRe
...
             // TODO: Should be configurable
             KeyPair keyPair = CertificateStore.newRSAKeyPair(4096);
-            X509Certificate[] caCertificates = getX509CACertificates();
 
             if (certificateType.equals(GetCertificateRequestOptions.CertificateType.SERVER)) {
-                Consumer<X509Certificate> consumer = (t) -> {
+                BiConsumer<X509Certificate, X509Certificate[]> consumer = (serverCert, caCertificates) -> {
                     CertificateUpdateEvent certificateUpdateEvent =
-                            new CertificateUpdateEvent(keyPair, t, caCertificates);
+                            new CertificateUpdateEvent(keyPair, serverCert, caCertificates);
```
jbutler, line 170: I think we still have the same problem here, except with the KeyPair. But I think we can solve that problem later.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -174,24 +163,23 @@ public void subscribeToCertificateUpdates(GetCertificateRequest getCertificateRe
                     getCertificateRequest.getCertificateRequestOptions().getCertificateType();
             // TODO: Should be configurable
             KeyPair keyPair = CertificateStore.newRSAKeyPair(4096);
-            X509Certificate[] caCertificates = getX509CACertificates();
```
Nelsonochoam, line 177: The issue is that we generate the consumer once, when an external component subscribe and then if the `caCertificates` change, we don't return the updated values because we are effectively creating a closure.

src/test/java/com/aws/greengrass/clientdevices/auth/CertificateManagerTest.java
```diff
@@ -86,10 +89,13 @@ public class CertificateManagerTest {
     Path tmpPath;
 
     private CertificateManager certificateManager;
+    private CertificateStore certificateStore;
 
     @BeforeEach
     void beforeEach() throws KeyStoreException {
```
Nelsonochoam, line 95: Even though `certManager` will go away eventually I think having a test for this is good while we still maintain it.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
Nelsonochoam: We are good for the keyPair not changing. From that key pair we only use the public key to generate the cert although it would be nice if we re generated a new one

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -174,24 +163,23 @@ public void subscribeToCertificateUpdates(GetCertificateRequest getCertificateRe
...
+                BiConsumer<X509Certificate, X509Certificate[]> consumer = (serverCert, caCertificates) -> {
                     CertificateUpdateEvent certificateUpdateEvent =
-                            new CertificateUpdateEvent(keyPair, t, caCertificates);
+                            new CertificateUpdateEvent(keyPair, serverCert, caCertificates);
                     getCertificateRequest.getCertificateUpdateConsumer().accept(certificateUpdateEvent);
                 };
                 subscribeToServerCertificateUpdatesNoCSR(getCertificateRequest, keyPair.getPublic(), consumer);
             } else if (certificateType.equals(GetCertificateRequestOptions.CertificateType.CLIENT)) {
-                Consumer<X509Certificate[]> consumer = (t) -> {
+                BiConsumer<X509Certificate, X509Certificate[]> consumer = (clientCert, caCertificates) -> {
```
Nelsonochoam, line 175: Just in case, this change does not change any external APIs. Subscribers pass a `GetCertificateRequest` which accepts a `CertificateUpdateEvent`, that is still the same. See line 178

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateGenerator.java
```diff
@@ -77,9 +77,7 @@ public synchronized void generateCertificate(Supplier<List<String>> connectivity
                     .kv("certExpiry", getExpiryTime())
                     .log("New client certificate generated");
 
-            X509Certificate caCertificate = certificateStore.getCACertificate();
-            X509Certificate[] chain = {certificate, caCertificate};
-            callback.accept(chain);
+            callback.accept(certificate, certificateStore.getCaCertificateChain());
```
jbutler, line 80: I still don't love that things in the certificate store can change midway through this flow. Fixing that will require bigger changes in the certificate store to essentially get a "snapshot".

Let's not try to fix that here, but just keep it in mind. (BTW, I am not expecting customers to change the CA type from managed to provided as doing so requires them to have access to the device to install keys)

jbutler (approved):

PR #135, opened 2022-09-22
feat: rotate certificates when ca changes

Nelsonochoam:
**Description of changes:**
This add a mechanism for us to trigger the subscriber when the ca changes

Without this change, changing the configuration does not rotate the certificates, causing component to fail authentication

Check and run the integration tests

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
@@ -5,7 +5,6 @@
 
 package com.aws.greengrass.integrationtests.certificateauthority;
 
-import com.aws.greengrass.clientdevices.auth.CertificateManager;
 import com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService;
 import com.aws.greengrass.clientdevices.auth.api.CertificateUpdateEvent;
 import com.aws.greengrass.clientdevices.auth.api.ClientDevicesAuthServiceApi;
@@ -50,9 +49,10 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.KeyPair;
-import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertPathValidatorException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CAConfigurationMonitor.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import javax.inject.Inject;
+
+
+// TODO: This class is acting almost like a handler the reason for not using a handler directly is because it registers
+//  CertificateGenerators same as the CertExpiryMonitor and CISMonitor. In order to minimize the surface area of changes
+//  without refactoring how we store subscribers we are taking this approach. Once we refactor how we store handlers
+//  this should become a fully fledged handler.
+public class CAConfigurationMonitor {
```
Nelsonochoam, line 25: In order to avoid doing more refactoring (for now), I am taking this hybrid approach of following what we currently do to register external subscribers that get their certificates rotated (using monitor classes) while at the same time using domain events.

I refrained from creating a listener and a handler due to the fact that we need a way to store the generators and I didn't want how we store the generators for all the monitors yet (it is explained in the TODO)

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -162,41 +162,31 @@ private void givenCDAWithCustomCertificateAuthority(URI privateKeyUri, URI certi
...
+        requestOptions.setCertificateType(type);
+        return new GetCertificateRequest("com.aws.clients.Plugin", requestOptions, cb);
+    }
 
     /**
-     * Simulates the request client components make to create their certificates. Clients pass their certificate
-     * chain pem to get verified.
+     * Simulates the subscription client or server component create to get their certificates.
      */
-    private X509Certificate[] arrangeClientComponentCertificateChain() throws CertificateGenerationException,
```
Nelsonochoam, line 170: This biggerish change you see here is just extracting a few things from here so that we can better simulate an external component subscribing to CDA CA changes through the API

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -263,4 +272,36 @@ void GIVEN_customCAConfigurationWithACAChain_WHEN_registeringCAWithIotCore_THEN_
         assertEquals(lastRequest.coreDeviceCertificates(), expectedPem);
     }
 
-}
\ No newline at end of file
+    @Test
+    void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serverCertificatesAreRotated() throws
```
Nelsonochoam, line 276: Focus here: This is the new test that demonstrates the functionality

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CAConfigurationMonitor.java
```diff
@@ -0,0 +1,93 @@
...
+
+        Result result = Result.ok();
+
+        for (CertificateGenerator generator : monitoredCertificateGenerators) {
+            try {
+                generator.generateCertificate(
+                        connectivityInformation::getCachedHostAddresses,
+                        "Certificate Configuration Changed");
+            } catch (CertificateGenerationException e) {
+                result = Result.warning(e);
```
jbutler, line 68: error?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CAConfigurationMonitor.java
```diff
@@ -0,0 +1,93 @@
...
+    public CAConfigurationMonitor(ConnectivityInformation connectivityInformation, DomainEvents domainEvents) {
+        this.connectivityInformation = connectivityInformation;
+        this.domainEvents = domainEvents;
+    }
+
+    /**
+     * Start ca configuration monitor.
+     */
+    public void startMonitor() {
+        domainEvents.registerListener(this::getHandler, CAConfigurationChanged.class);
```
jbutler, line 49: LOVE how easy it was to hook this in

jbutler (approved):

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -148,44 +147,46 @@ public X509Certificate getCACertificate() throws KeyStoreException {
...
+     *
+     * @param privateKey  leaf CA private key
+     * @param caCertificateChain a CA chain
      *
-     * @param caCertificateChain Array of CA certificates
-     * @throws KeyStoreException if unable to retrieve the certificate chain
+     * @throws KeyStoreException  if privateKey is not instance of PrivateKey or no ca chain provided
      */
-    public void setCaCertificateChain(X509Certificate... caCertificateChain) throws KeyStoreException {
+    public void setCA(Key privateKey, X509Certificate... caCertificateChain) throws KeyStoreException {
```
Nelsonochoam, line 161: I realized there was a potential for us to introduce a very sneaky bug in the scenario where we just call setCertificateChain and do not call setPrivateKey. It might lead to us signing certificates with an old private key. This change just changes the API so we have to set both at once in order to avoid making impossible states possible

jbutler:
Had some force push shenanigans going on there, but `7bcddf4` is getting merged which previously had 2 approvals. Will tackle follow up changes in separate PR

PR #136, opened 2022-09-22
fix: avoid potential impossible states

Nelsonochoam:
**Description of changes:**
Update caChain and private key together to avoid leaving the door open for someone to update the caChain without changing the private key which would lead to generating invalid certificates.

If we don't do this, depending on how we hook the events it can happen that we generate certificates with an old private key.

This is a refactor, ensure all tests are passing.

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
@@ -73,8 +73,6 @@ import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.lenient;
@@ -278,7 +276,8 @@ public class CustomCaConfigurationTest {
     void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serverCertificatesAreRotated() throws
             InterruptedException, CertificateGenerationException, CertificateException, NoSuchAlgorithmException,
             OperatorCreationException, IOException, URISyntaxException, KeyLoadingException,
-            ServiceUnavailableException, CertificateChainLoadingException, ServiceLoadException {
+            ServiceUnavailableException, CertificateChainLoadingException, ServiceLoadException, ExecutionException,
+            TimeoutException {
         Pair<X509Certificate[], KeyPair[]> credentials = givenRootAndIntermediateCA();
         X509Certificate[] chain = credentials.getLeft();
         X509Certificate intermediateCA =  chain[0];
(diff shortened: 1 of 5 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -290,18 +291,19 @@ void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serve
         when(securityServiceMock.getCertificateChain(privateKeyUri, certificateUri)).thenReturn(chain);
 
         AtomicReference<CertificateUpdateEvent> eventRef = new AtomicReference<>();
-        Pair<CompletableFuture<Void>, Consumer<CertificateUpdateEvent>> asyncCall =
```
Nelsonochoam, line 293: Didn't really need to use the `asyncAssertOnConsumer` for the purpose of this test so I just removed it

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -148,44 +148,46 @@ public X509Certificate getCACertificate() throws KeyStoreException {
...
+     *
+     * @param privateKey  leaf CA private key
+     * @param caCertificateChain a CA chain
      *
-     * @param caCertificateChain Array of CA certificates
-     * @throws KeyStoreException if unable to retrieve the certificate chain
+     * @throws KeyStoreException  if privateKey is not instance of PrivateKey or no ca chain provided
      */
-    public void setCaCertificateChain(X509Certificate... caCertificateChain) throws KeyStoreException {
+    public void setCaCertificateChain(Key privateKey, X509Certificate... caCertificateChain) throws KeyStoreException {
```
Nelsonochoam, line 162: Lump together the update of the CA chain and private key. If we don't do this we leave the door open for future bugs where we update we have an old private key stores, we call `setCaCertificateChain` which emits events that might end up creating certificates reading the stale value of the key.

A CA should always be accompanied with its private key and change atomically (all at once) to avoid issues.

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -148,44 +148,46 @@ public X509Certificate getCACertificate() throws KeyStoreException {
...
         }
 
+        this.caCertificateChain = caCertificateChain;
         caPrivateKey = (PrivateKey) privateKey;
+
+        eventEmitter.emit(new CACertificateChainChanged(caCertificateChain));
+    }
+
+
+    private void setCaCertificateChain(Key privateKey, Certificate... caCertificateChain) throws KeyStoreException {
```
jbutler, line 178: This method name doesn't tell me that the private key is being written. `setCertificateAuthority` or `Set CaKeyAndCertificateChain` ?

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -290,18 +291,19 @@ void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serve
...
         subscribeToCertificateUpdates(request);
-        givenCDAWithCustomCertificateAuthority(privateKeyUri, certificateUri);
+        assertNotEquals(CertificateHelper.toPem(chain), CertificateHelper.toPem(eventRef.get().getCaCertificates()));
+        assertFalse(CertificateTestHelpers.wasCertificateIssuedBy(intermediateCA, eventRef.get().getCertificate()));
 
-        // Called 2 times. 1 for initial manages CA and then after the config is changes to use custom CA
-        asyncCall.getLeft().get(2, TimeUnit.SECONDS);
-        CertificateUpdateEvent event = eventRef.get();
-        assertTrue(CertificateTestHelpers.wasCertificateIssuedBy(intermediateCA, event.getCertificate()));
+        // Change to custom CA
```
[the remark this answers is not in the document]
Nelsonochoam: True then we might want to use the async wait. Ill change it back

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
Nelsonochoam: I generally don't like the `And` word in methods it is a smell. It feels like we need an object that represent that intent, forcing things that change together to be together. I don't like the term `Credentials` is there a better term on the TLS, JCA worlds for that?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -148,44 +148,46 @@ public X509Certificate getCACertificate() throws KeyStoreException {
...
-            if (!(cert instanceof X509Certificate)) {
-                throw new KeyStoreException("Unsupported certificate type");
-            }
-        }
-        setCaCertificateChain(Arrays.stream(caCertificateChain).toArray(X509Certificate[]::new));
-    }
 
     /**
-     * Sets the CA certificate chain.
+     * Sets the CA chain and private key that are used to generate certificates. It combines setting both values
```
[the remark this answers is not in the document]
Nelsonochoam: Tabling this for now. I'll make the "Get" operations synchronized in a next PR. I don't want to change the getters on this one

jbutler (approved):

MikeDombo (approved):

PR #137, opened 2022-09-22
feat: move required methods from nucleus to cda

Nelsonochoam:
**Description of changes:**
This is just a copy past from the methods required from  SecurityService on the nucleus to CDA so that we can release CDA without having to force people to update to the latest version of the nucleus if they want to use a custom CA with CDA. In the future we can start using the security service again.

Release CDA without upgrading the nucleus, just a step to de risk a bit this feature.

Added new testing tooling to write the PEMs to file and ensure the existing tests are passing using the ported methods.


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
```diff
@@ -253,4 +269,97 @@ public static X500Name getX500Name(String commonName) {
 
         return nameBuilder.build();
     }
+
```
Nelsonochoam, line 272: All this chunk is copy pasta from Nucleus, I just tweaked it a bit to not have to select a provider, we just wanted the methods here.

src/test/java/com/aws/greengrass/clientdevices/auth/CertificateManagerTest.java
```diff
@@ -160,18 +146,13 @@ void Given_defaultCAConfiguration_THEN_returnsAutoGeneratedCA() throws Exception
         X509Certificate caCertificate = CertificateHelper.createCACertificate(
                 keyPair, Date.from(now), Date.from(now.plusSeconds(10)), "Custom Core CA");
 
-        URI privateKeyUri = new URI("file:///private.key");
-        URI certificateUri = new URI("file:///certificate.pem");
-
-        // TODO: Write the actual certificate to the file system and avoid mocking the security service. Doing
-        //  this is a bad given we are exposing implementation details on the test.
-        when(securityServiceMock.getKeyPair(privateKeyUri, certificateUri)).thenReturn(keyPair);
```
Nelsonochoam, line 168: No more mocking we are writing them and letting the things actually execute the ported methods. This gives us confidence they are working

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/CertificateTestHelpers.java
```diff
@@ -199,4 +207,38 @@ public static boolean wasCertificateIssuedBy(X509Certificate issuerCA, X509Certi
             return false;
         }
     }
+
+    public static void writeToCertificatesToPath(Path filePath, X509Certificate... certificates)
```
Nelsonochoam, line 211: Some tooling to write the PEMs to the file system. This is just used for tests

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
Nelsonochoam, line 272: Ported methods from Nucleus

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -275,22 +269,12 @@ public void configureCustomCA(CDAConfiguration configuration) throws InvalidConf
         URI privateKeyUri = configuration.getPrivateKeyUri().get();
         URI certificateUri = configuration.getCertificateUri().get();
 
-        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
```
Nelsonochoam, line 278: Not sure if removing this was right. In my ming if the certificate is written in the FS and it is wrong then no matter how many times you retry it will still be wrong, so I remove the retry. Please educate me if that is a wrong assumption.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
MikeDombo: Why!?

a) You already have the Nucleus as a dependency, you can use any of its methods
b) There's a good reason why Nucleus abstracts these. These methods will not work when a HSM is used

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
jbutler: Yes, this is wrong. It's not here to protect against the key being wrong, it's here because CDA can come up before the SecurityService itself comes up.

FWIW, the retry policy is still wrong. Waiting 200ms is not even remotely sufficient as a backoff. But we can fix that later.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
jbutler: Don't port everything. Just take the certificate chain bit.

@MikeDombo The goal is to avoid bumping Nucleus dependency. I don't understand your HSM comment. The Nucleus wrapper is just calling the underlying providers, which is what should be happening here

src/test/java/com/aws/greengrass/clientdevices/auth/CertificateManagerTest.java
```diff
@@ -88,8 +85,6 @@ public class CertificateManagerTest {
 
     @Mock
     GreengrassServiceClientFactory clientFactoryMock;
-    @Mock
-    SecurityService securityServiceMock;
```
jbutler, line 92: Still partially needed

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
Nelsonochoam: Well now we are not using the SecurityService... so I guess we should not need to re add it.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
MikeDombo: What is the current Nucleus dependency version? 

What functionality from the newer Nucleus version do you want?

MikeDombo: `Paths.get(privateKeyUri)` does not work when the key isn't a file. What key is going to be loaded by this method, do we know with certainty that it is going to be a file and not a pkcs11 URI?

jbutler: This code isn't correct. It needs to still use the SecurityService.

The new Nucleus API to get the cert chain is just a value. I want to avoid bumping Nucleus dependency for something we can easily include here. That method still uses other SecurityService primitives.

Right now CDA doesn't work with the latest public Nucleus. I want it to

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -8,6 +8,7 @@
 import com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService;
 import com.aws.greengrass.clientdevices.auth.api.CertificateUpdateEvent;
 import com.aws.greengrass.clientdevices.auth.api.ClientDevicesAuthServiceApi;
+import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
```
jbutler, line 11: Can you bump Nucleus back down to 2.7 and make sure this works?

pom.xml
```diff
@@ -73,13 +73,13 @@
         <dependency>
             <groupId>com.aws.greengrass</groupId>
             <artifactId>nucleus</artifactId>
-            <version>2.8.0-SNAPSHOT</version>
+            <version>2.6.0-SNAPSHOT</version>
```
Nelsonochoam, line 76: This was the version it was before all this changes

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
Nelsonochoam: Moved it to 2.6.0 which was what we were using before all the changes, just to be safe and not change what doesn't need to

src/main/java/com/aws/greengrass/clientdevices/auth/exception/CertificateChainLoadingException.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.exception;
```
Nelsonochoam, line 6: Had to copy pasta this from Nucleus as well given it was also a new thing added

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -134,6 +150,44 @@ public PrivateKey getCAPrivateKey() {
...
+     * @param privateKeyUri  private key URI
+     * @param certificateUri certificate URI
+     * @return X509Certificate list that corresponds to the given private key and certificate URIs
+     * @throws KeyLoadingException              if it fails to load key
+     * @throws CertificateChainLoadingException if no certificate chain is found
+     * @throws ServiceUnavailableException     if the security service is not available
+     */
+    public  X509Certificate[] loadCaCertificateChain(URI privateKeyUri, URI certificateUri)
+            throws CertificateChainLoadingException, KeyLoadingException, ServiceUnavailableException {
+        KeyManager[] km = securityService.getKeyManagers(privateKeyUri, certificateUri);
```
MikeDombo, line 167: Yes, good. Now that this is using the security service it will work correctly when the device is configured to use an HSM.

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -344,4 +344,4 @@ public void uploadCoreDeviceCAs(String thingName, X509Certificate certificate)
                     + "core device's IoT policy grants the greengrass:PutCertificateAuthorities permission.", e);
         }
     }
-}
+}
```
MikeDombo, line 347: Please configure your IDE to keep/add a newline at the end of files.

MikeDombo (approved):

PR #138, opened 2022-09-22
feat(wip): add thing registry domain and events

jbutler:
**Issue #, if available:**

**Description of changes:**
Add new Thing Registry interface and emit events when Things are added/updated. This change is somewhat incomplete, as right now timestamps aren't included.

This will be a building block for follow-up PRs which introduce authentication use cases.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
@@ -70,7 +70,7 @@ public final class Thing implements AttributeProvider, Cloneable {
         if (!Pattern.matches(thingNamePattern, thingName)) {
             throw new IllegalArgumentException("Invalid thing name. The thing name must match \"[a-zA-Z0-9\\-_:]+\".");
         }
-        return new Thing(version, thingName, new CopyOnWriteArrayList<>(certificateIds));
+        return new Thing(version, thingName, certificateIds);
     }
 
     @Override
@@ -109,13 +109,13 @@ public final class Thing implements AttributeProvider, Cloneable {
      * @return Certificate IDs
      */
     public List<String> getAttachedCertificateIds() {
-        return new CopyOnWriteArrayList<>(attachedCertificateIds);
+        return new ArrayList<>(attachedCertificateIds);
     }
 
-    private Thing(int version, String thingName, CopyOnWriteArrayList<String> certificateIds) {
+    private Thing(int version, String thingName, List<String> certificateIds) {
         this.version = version;
         this.thingName = thingName;
-        this.attachedCertificateIds = certificateIds;
+        this.attachedCertificateIds = new CopyOnWriteArrayList<>(certificateIds);
     }
 
     @Override
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
@@ -8,48 +8,143 @@ package com.aws.greengrass.clientdevices.auth.iot;
 import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
 import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
 import com.aws.greengrass.clientdevices.auth.session.attribute.WildcardSuffixAttribute;
-import lombok.Value;
+import lombok.Getter;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
 /**
- * This is a value object representing an IoT Thing at specific point in time.
- * It is **NOT** updated when the local Thing Registry is updated, or when
- * changes to this Thing are made in IoT Core.
+ * This is a versioned representation of an IoT Thing. It is **NOT** updated
+ * when the local Thing Registry is updated, or when changes to this Thing are
+ * made in IoT Core. Consider calling the ThingRegistry to retrieve Thing
+ * objects as they are needed rather than storing references long term.
  */
-@Value
-public class Thing implements AttributeProvider {
+@Getter
+public final class Thing implements AttributeProvider {
     public static final String NAMESPACE = "Thing";
     private static final String thingNamePattern = "[a-zA-Z0-9\\-_:]+";
 
-    String thingName;
-    List<String> attachedCertificateIds;
+    private final int version;
+    private final String thingName;
+    private final List<String> attachedCertificateIds;
+    private boolean modified = false;
 
     /**
-     * Create a Thing object with no attached certificates.
+     * Create a new Thing.
      *
      * @param thingName AWS IoT ThingName
+     * @throws IllegalArgumentException If the given ThingName contains illegal characters
      */
-    public Thing(String thingName) {
-        this(thingName, null);
+    public static Thing of(String thingName) {
+        return Thing.of(0, thingName);
     }
 
     /**
-     * Create a Thing object with the provided attached certificate IDs.
+     * Create a new Thing.
      *
+     * @param version        Thing version
      * @param thingName      AWS IoT ThingName
-     * @param certificateIds AWS IoT Certificate IDs
      * @throws IllegalArgumentException If the given ThingName contains illegal characters
      */
-    public Thing(String thingName, List<String> certificateIds) {
+    public static Thing of(int version, String thingName) {
+        return Thing.of(version, thingName, new ArrayList<>());
+    }
+
+    /**
+     * Create a new Thing.
+     *
+     * @param version        Thing version
+     * @param thingName      AWS IoT ThingName
+     * @param certificateIds Attached certificate IDs
+     * @throws IllegalArgumentException If the given ThingName contains illegal characters
+     */
+    public static Thing of(int version, String thingName, List<String> certificateIds) {
+        if (version < 0) {
+            throw new IllegalArgumentException("Invalid version. Version must not be < 0");
+        }
         if (!Pattern.matches(thingNamePattern, thingName)) {
             throw new IllegalArgumentException("Invalid thing name. The thing name must match \"[a-zA-Z0-9\\-_:]+\".");
         }
+        return new Thing(version, thingName, new ArrayList<>(certificateIds));
+    }
+
+    /**
+     * Create a copy of a Thing.
+     *
+     * @param other Thing to copy
+     */
+    public static Thing of(Thing other) {
+        return Thing.of(other.getVersion(), other.getThingName(), other.getAttachedCertificateIds());
+    }
+
+    /**
+     * Attach a certificate ID.
+     * @param certificateId Certificate ID to attach
+     */
+    public void attachCertificate(String certificateId) {
+        if (attachedCertificateIds.contains(certificateId)) {
+            return;
+        }
+        attachedCertificateIds.add(certificateId);
+        modified = true;
+    }
+
+    /**
+     * Detach a certificate ID.
+     * @param certificateId Certificate ID to detach
+     */
+    public void detachCertificate(String certificateId) {
+        if (!attachedCertificateIds.contains(certificateId)) {
+            return;
+        }
+        attachedCertificateIds.remove(certificateId);
+        modified = true;
+    }
+
+    /**
+     * Returns copy of attached certificate IDs.
+     * </p>
+     * This list cannot be modified directly. Refer to {@link #attachCertificate(String) attachCertificate} and
+     * {@link #detachCertificate(String) detachCertificate}
+     *
+     * @return Certificate IDs
+     */
+    public List<String> getAttachedCertificateIds() {
+        return new ArrayList<>(attachedCertificateIds);
+    }
+
+    private Thing(int version, String thingName, List<String> certificateIds) {
+        this.version = version;
         this.thingName = thingName;
         this.attachedCertificateIds = certificateIds;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        }
+        if (!(o instanceof Thing)) {
+            return false;
+        }
+
+        Thing other = (Thing) o;
+        if (isModified() || other.isModified()) {
+            return false;
+        }
+
+        return version == other.version
+                && thingName.equals(other.thingName)
+                && attachedCertificateIds.equals(other.attachedCertificateIds);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(version, thingName, attachedCertificateIds);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/events/ThingEvent.java
```diff
@@ -0,0 +1,26 @@
...
+public class ThingEvent implements DomainEvent {
+    @Getter
+    private ThingEventType eventType;
+    @Getter
+    private String thingName;
+    @Getter
+    private List<String> attachedCertificateIds;
+
+    public enum ThingEventType {
+        THING_UPDATED
```
jbutler, line 24: I started with ADDED/UPDATED/DELETED, but we don't really have use cases for those other event types, so I removed them. This simplifies registry logic somewhat.

The idea is to try to model IoT Core registry events:
https://docs.aws.amazon.com/iot/latest/developerguide/registry-events.html

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,101 @@
...
+    }
+
+    /**
+     * Attach a certificate to Thing.
+     * </p>
+     * If the provided Thing does not exist, it will be created.
+     * @param thingName     ThingName.
+     * @param certificateId Certificate ID to attach.
+     */
+    public synchronized void attachCertificateToThing(String thingName, String certificateId) {
```
Nelsonochoam, line 63: It seems weird that the `ThingRegistry` operates on the certificateIds. The 2 arguments we are taking are already part of the `Thing` value object, also the thing value object has the added benefit of a validated `thingName` here we open the door for it being called with whatever string.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,101 @@
...
+        List<String> certificateIds = registry.get(thingName);
+
+        if (certificateIds != null && certificateIds.contains(certificateId)) {
```
Nelsonochoam, line 66: If we were using this value object then this messages could be asked to the `Thing` 

```java
thing.hasCertificateId(certificateId)
```

Also passing the value object directly would guarantee we don't have to worry about creating a new `ArrayList` since we would already have a thing. This method would become more of an  `update`

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,101 @@
...
+                        thingName,
+                        new ArrayList<>(certificateIds)));
+    }
+
+    /**
+     * Detach a certificate from Thing.
+     * @param thingName     ThingName.
+     * @param certificateId Certificate ID to detach.
+     */
+    public synchronized void detachCertificateFromThing(String thingName, String certificateId) {
```
Nelsonochoam, line 90: Similar to my comment above, if we were operating on `Things` this method would not exist we would just update the Thing and call the registry's `update` or `save` method. Also would remove the duplication in event emitting

src/main/java/com/aws/greengrass/clientdevices/auth/iot/events/ThingEvent.java
```diff
@@ -0,0 +1,26 @@
...
+@AllArgsConstructor
+public class ThingEvent implements DomainEvent {
+    @Getter
+    private ThingEventType eventType;
+    @Getter
+    private String thingName;
+    @Getter
+    private List<String> attachedCertificateIds;
+
+    public enum ThingEventType {
```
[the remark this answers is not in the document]
jbutler: I'm good with this approach. Will update

src/main/java/com/aws/greengrass/clientdevices/auth/iot/events/ThingEvent.java
jbutler: No JSON needed at this point. We can remove the type enum

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,101 @@
...
+    }
+
+    /**
+     * Retrieve a Thing based on ThingName.
+     *
+     * @param thingName ThingName
+     * @return Thing domain object
+     */
+    public Thing getThing(String thingName) {
+        List<String> certificateIds = registry.get(thingName);
```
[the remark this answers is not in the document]
jbutler: Don't get too hung up on the current internal storage. I didn't want to touch those bits more than I needed to, but I expect it will change later.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
jbutler: Let's chat about the best way to handle this one.

I am nervous about passing around Thing objects, because I am afraid of how they will be used. I want to treat anything provided by the registry as an immutable snapshot. They should be one time use objects, which are thrown away.

The reason for that is, that I don't want the registry to need to maintain references to every single Thing object that is ever returned. The Registry has no way of knowing if an external caller still has a reference to an object, so it doesn't know if it can "throw it away".

It would perhaps be better if only the Registry can instantiate Thing objects. And if we go that route, then we can't base an update API using it.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -8,29 +8,143 @@
...
+    /**
+     * Create a new Thing.
+     *
+     * @param version        Thing version
+     * @param thingName      AWS IoT ThingName
+     * @param certificateIds Attached certificate IDs
+     * @throws IllegalArgumentException If the given ThingName contains illegal characters
+     */
+    public static Thing of(int version, String thingName, List<String> certificateIds) {
+        if (version < 0) {
```
Nelsonochoam, line 66: Thought: What do you thing about versions just being a wrapper (maybe we can do it later if we ever need to have more that 1 versions thing) but I was thinking about `Versioned<Thing>` and the api to interact with it would be something like

```java
Versioned<Thing> versionedThing = Versioned.of(thing)
Versioned<Thing> versionedThing = Versioned.of(10, thing)

int version = versionedThing.getVersion();
Thing thing = versionedThing.get();
```

Then the wrapper would have this validation logic. **again, this is thought** this works just fine.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -8,29 +8,143 @@
...
         }
+        return new Thing(version, thingName, new ArrayList<>(certificateIds));
+    }
+
+    /**
+     * Create a copy of a Thing.
+     *
+     * @param other Thing to copy
+     */
+    public static Thing of(Thing other) {
```
Nelsonochoam, line 80: `copy`?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -8,29 +8,143 @@
...
+        attachedCertificateIds.add(certificateId);
+        modified = true;
+    }
+
+    /**
+     * Detach a certificate ID.
+     * @param certificateId Certificate ID to detach
+     */
+    public void detachCertificate(String certificateId) {
+        if (!attachedCertificateIds.contains(certificateId)) {
```
Nelsonochoam, line 101: Why not make `attachedCertificateIds` a `HashSet`?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,94 @@
...
+            // TODO: throw exception since caller is trying to update an old version
+            return null;
+        }
+
+        if (!thing.isModified()) {
+            // Nothing to modify - return old object
+            return thing;
+        }
+
+        Thing newThing = Thing.of(thing.getVersion() + 1, thing.getThingName(), thing.getAttachedCertificateIds());
```
Nelsonochoam, line 84: I like this - Just following up on my thought comment above this could be part of the `Versioned` class. 

```java
Versioned<Thing> newThings = versionedThing.increment();
```

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,94 @@
...
+     * Update a Thing.
+     * @param thing Thing which is being updated
+     * @return      New Thing version
+     */
+    public Thing updateThing(Thing thing) {
+        // TODO - need to throw exception if provided Thing is not the most recent version
+        Thing oldThing = getThingInternal(thing.getThingName());
+
+        if (oldThing == null) {
+            storeThing(thing);
```
Nelsonochoam, line 70: I think here we would also want to increment the version. Every time we store a thing or make a change in the registry we return a new version.

jbutler: I actually removed this code -- I'm forcing callers to call `createThing` first. I agree with the versioning approach.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
jbutler: Hmm, yes, it could make sense to extract out the version functionality

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
Nelsonochoam: Sure, I guess it will be rare to have too many certIds and we could always use `addIfAbsent`.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -8,29 +8,143 @@
...
+ */
+@Getter
+public final class Thing implements AttributeProvider {
     public static final String NAMESPACE = "Thing";
     private static final String thingNamePattern = "[a-zA-Z0-9\\-_:]+";
 
-    String thingName;
+    private final int version;
+    private final String thingName;
+    private final List<String> attachedCertificateIds;
```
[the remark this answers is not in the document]
jbutler: See comment below. We're expecting this list to be very short, so practically speaking runtime complexity doesn't matter. I'd rather optimize for memory usage

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -8,29 +8,138 @@
...
+     * @throws IllegalArgumentException If the given ThingName contains illegal characters
+     */
+    public static Thing of(int version, String thingName, List<String> certificateIds) {
+        if (version < 0) {
+            throw new IllegalArgumentException("Invalid version. Version must not be < 0");
+        }
         if (!Pattern.matches(thingNamePattern, thingName)) {
             throw new IllegalArgumentException("Invalid thing name. The thing name must match \"[a-zA-Z0-9\\-_:]+\".");
         }
+        return new Thing(version, thingName, new CopyOnWriteArrayList<>(certificateIds));
```
MikeDombo, line 73: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem:**
 This line of code lacks validation when processing input data through the following parameter: &#39;certificateIds&#39; (index: 2 | type: List). The parameter is exposed to external callers, because its enclosing class and method are publicly accessible. This means that upstream validation, if it exists, can be bypassed. Other validated parameters: &#39;thingName&#39;, &#39;version&#39;. Malicious, malformed, or unbounded inputs can cause unexpected runtime behavior or crashes, and can slow performance.

**Fix:**
 Add checks to ensure the validity of the parameter&#39;s value, such as testing it for nullness (for example, using the @NonNull annotation as [described in the Lombok library](https://projectlombok.org/features/NonNull)), emptiness, or equality. Or to prevent direct calls to it, [reduce the method&#39;s visibility](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html).

 Learn more about potential threats and guidance from the [Common Weakness Enumeration website](https://cwe.mitre.org/data/definitions/693.html) and the [OWASP Cheat Sheet series](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html).

Nelsonochoam (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,113 @@
...
+        return newThing.clone();
+    }
+
+    /**
+     * Retrieve a Thing based on ThingName.
+     *
+     * @param thingName ThingName
+     * @return Thing domain object, if it exists
+     */
+    public Thing getThing(String thingName) {
```
Nelsonochoam, line 71: Return Optional

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -5,33 +5,113 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Update a Thing.
+     * @param thing Thing which is being updated
+     * @return      New Thing version
+     */
+    public Thing updateThing(Thing thing) {
```
Nelsonochoam, line 84: Return Optional<Thing>

vaibhavmurkute (approved):

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
```diff
@@ -11,31 +11,107 @@
...
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing@"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing#"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing$"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing%"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing^"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing&"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing*"));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing("));
-        Assertions.assertThrows(IllegalArgumentException.class, () -> new Thing("Thing)"));
+    void GIVEN_thingNameWithInvalidCharacters_WHEN_Thing_THEN_exceptionThrown() {
```
Nelsonochoam, line 38: Great opportunity to have property bases testing

PR #140, opened 2022-09-26
fix: handler registration

Nelsonochoam:
**Description of changes:**
When exploring the race condition that was happening on tests it seemed (from the logs) the `CaConfigurationMonitor` could register its handler multiple times with the domain events. Registering a consumer using the `this::method` notation wraps the consumer on a new lambda every time is called.  Leaving the door open for the handler to be registered multiple times. 

Additionally the handler was getting registered on the `startup` method which can be potentially unsafe if called multiple times, given we are not unregistering handlers from the domain events.

Avoid the CA configuration change certificate rotation from being registered multiple times.

**Additional Notes**:
This doesn't necessarily addresses the race condition we saw on the failing tests but based on the information gathered from the test failure logs this would make the code have higher reliability given it would behave appropriately even if that race condition occurs (the ca configuration changed certificate rotation handler will only be able to get registered once).


src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -110,7 +112,8 @@ void setup(ExtensionContext context) throws DeviceConfigurationException {
         kernel = new Kernel();
         kernel.getContext().put(GroupManager.class, groupManager);
         kernel.getContext().put(SecurityService.class, securityServiceMock);
-        kernel.getContext().put(CertificateExpiryMonitor.class, certExpiryMonitor);
+        kernel.getContext().put(CertificateExpiryMonitor.class, certExpiryMonitorMock);
+        kernel.getContext().put(CISShadowMonitor.class, cisShadowMonitorMock);
```
Nelsonochoam, line 116: Mocking the cisShadowMonitor given it was always getting triggered trying to subscribe during this tests and always fails (we don't care about the cisShadowMonitor) in the context of this test.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CertificateRotationHandler.java
```diff
@@ -90,4 +66,25 @@ public void addToMonitor(CertificateGenerator cg) {
     public void removeFromMonitor(CertificateGenerator cg) {
         monitoredCertificateGenerators.remove(cg);
     }
+
+    @Override
+    public void accept(CAConfigurationChanged event) {
+        logger.debug("Received event {}", event.getName());
+
+        if (monitoredCertificateGenerators.isEmpty()) {
+            logger.info("No certificates to rotate, skipping");
```
MikeDombo, line 75: this log doesn't seem very useful. Drop to debug?

MikeDombo (dismissed):

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CertificateRotationHandler.java
```diff
@@ -37,42 +37,18 @@ public class CAConfigurationMonitor {
...
     @Inject
-    public CAConfigurationMonitor(ConnectivityInformation connectivityInformation, DomainEvents domainEvents) {
+    public CertificateRotationHandler(ConnectivityInformation connectivityInformation, DomainEvents domainEvents) {
         this.connectivityInformation = connectivityInformation;
         this.domainEvents = domainEvents;
     }
 
     /**
      * Start ca configuration monitor.
      */
```
vaibhavmurkute, line 47: need to update this doc?

vaibhavmurkute (approved):

PR #141, opened 2022-09-28
fix: increase wait time for security service to be ready

Nelsonochoam:
**Description of changes:**
This are some improvements required for the component to work with HSM.

We are not waiting long enough after the first failure to get the keys while the provider has not been loaded before retrying


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -250,9 +250,12 @@ void updateCACertificateConfig(List<String> caCerts) {
 
     @Override
     protected CompletableFuture<Void> close(boolean waitForDependers) {
-        // shutdown the threadpool in close, not in shutdown() because it is created
-        // and injected in the constructor and we won't be able to restart it after it stops.
-        cloudCallThreadPool.shutdown();
+        if (cloudCallThreadPool != null) {
```
Nelsonochoam, line 253: I am adding this here because I saw `NullPointer` exceptions get thrown occasionally when close was being called before this had been initialized

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -178,6 +178,28 @@ public  X509Certificate[] loadCaCertificateChain(URI privateKeyUri, URI certific
             throw new CertificateChainLoadingException("Unable to find aliases in the key manager with the given "
                     + "private key and certificate URIs");
         }
+
+        // TODO: We are doing this here to avoid changing the PKCS11 provider or the nucleus code base. It is not
+        //  pretty but
+        if (privateKeyUri.getScheme().equals("pkcs11")) {
```
Nelsonochoam, line 184: This is nasty it might be temporary. It is just to get reading certificates from HSM working without having to add this functionality to the PKCS11 provider, and SecurityService

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
```diff
@@ -65,13 +71,45 @@ public final class CertificateHelper {
...
+    public enum ProviderType {
+        HSM, DEFAULT
+    }
+
+
+    /**
+     * Computes and returns the provider names from the registered providers in JCA that are required for issuing the
+     * certificates. When a user uses the PKCS11 provider it registers a provider by the name SunPKCS11-[name].
+     */
+    private static void computeProviders() {
```
Nelsonochoam, line 96: If this seems odd I would understand you, but this is the explanation - The PKCS11 provider component will create a SunPKCS11 provider, the way it works is that when we create our PKCS11 provider, it gets registered with a specific notation see documentation for the name config value [here](https://docs.oracle.com/javase/7/docs/technotes/guides/security/p11guide.html)

Given we don't know which name the user used on their configuration (it is on the PKCS11 Provider config) but we know that custom SunPKCS11 providers get named `SunPKCS11-[name]` this method tries to get the name of the registered provider so that then we can use it when issuing certificates

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -290,7 +296,11 @@ public void configureCustomCA(CDAConfiguration configuration) throws InvalidConf
                     () -> certificateStore.loadCaCertificateChain(privateKeyUri, certificateUri),
                     "get-certificate-chain", logger);
 
-            certificateStore.setCaKeyAndCertificateChain(keyPair.getPrivate(), certificateChain);
+            CertificateHelper.ProviderType providerType = privateKeyUri.getScheme().contains(pkcs11Scheme)
```
jbutler, line 299: `file:///pkcs11.pem` ? :)

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -210,10 +216,12 @@ public X509Certificate getCACertificate() throws KeyStoreException {
       *
       * @param privateKey  leaf CA private key
       * @param caCertificateChain a CA chain
+      * @param providerType  provider type DEFAULT or HSM, used to map to the correct JCA provider
       *
       * @throws KeyStoreException  if privateKey is not instance of PrivateKey or no ca chain provided
       */
-     public synchronized void setCaKeyAndCertificateChain(Key privateKey, X509Certificate... caCertificateChain)
+     public synchronized void setCaKeyAndCertificateChain(
+             CertificateHelper.ProviderType providerType, Key privateKey, X509Certificate... caCertificateChain)
```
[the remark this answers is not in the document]
jbutler: I don't think we have the right interface here in the CertificateStore. Ideally we'd just encapsulate everything. But, I'm ok with this for now. I hate how we're leaking details like the underlying crypto provider

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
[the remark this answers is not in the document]
Nelsonochoam: There was not a really good way of checking if we should use the HSM provider since that is basically dictated by configuration and I didn't want to add more stuff. I am sure we could work something out to improve it.

vaibhavmurkute (approved):

PR #143, opened 2022-09-29
refactor: add repository for registries

vaibhavmurkute:
**Description of changes:** adds `Repository` for the registries that mediates between the registry domain and the data storage layer. This Repository can be backed with in-memory as well as filesystem storage. This PR includes only the in-memory repository implementation.

**Why is this change necessary:** To be able to switch storage layer for registries.

**How was this change tested:** unit, integ tests (`mvn verify` , `mvn package`)

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
@@ -13,35 +13,26 @@ import com.aws.greengrass.util.Digest;
 import com.aws.greengrass.util.Utils;
 
 import java.security.NoSuchAlgorithmException;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.Map;
 import java.util.Optional;
 import javax.inject.Inject;
 
 
 public class CertificateRegistry {
     private static final Logger logger = LogManager.getLogger(CertificateRegistry.class);
-    // holds mapping of certificateHash (SHA-256 hash of certificatePem) to IoT Certificate Id;
-    // size-bound by default cache size, evicts oldest written entry if the max size is reached
-    private final Map<String, String> registry = Collections.synchronizedMap(
-            new LinkedHashMap<String, String>(RegistryConfig.REGISTRY_CACHE_SIZE, 0.75f, false) {
-                @Override
-                protected boolean removeEldestEntry(Map.Entry eldest) {
-                    return size() > RegistryConfig.REGISTRY_CACHE_SIZE;
-                }
-            });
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -13,35 +13,26 @@
...
 
 import java.security.NoSuchAlgorithmException;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.Map;
 import java.util.Optional;
 import javax.inject.Inject;
 
 
 public class CertificateRegistry {
```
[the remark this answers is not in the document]
Nelsonochoam: Here is an article with some more details https://khalilstemmler.com/articles/typescript-domain-driven-design/repository-dto-mapper/#Repositories

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/InMemoryRepository.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Mediates between the registry domain and the in-memory data storage.
+ *
+ * @param <K> Registry key type
+ * @param <V> Registry value type
+ */
+public class InMemoryRepository<K, V> implements Repository<K, V> {
```
Nelsonochoam, line 19: This is close but the Registry above should be the one implementing this interface

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -13,35 +13,26 @@
 import com.aws.greengrass.util.Utils;
 
 import java.security.NoSuchAlgorithmException;
-import java.util.Collections;
```
jbutler, line 16: High level comments:

* Ideally these registry APIs would be an interface, but our DI framework isn't flexible enough. We'll inject the registry objects into use cases. The registry itself can be constructed with whatever storage layer we need
* I'm not too concerned about an interface for the storage layer. The only thing using it is the registry itself. We can trivially refactor this later if we need other storage options.
* Don't touch CertRegistry yet. That's just creating more work for us (I'm making significant changes to it)
* ThingRegistry - skip the InMemory repo. We're not going to use it. I think you can just implement the runtime configuration storage and logic to convert from Thing to Thing configuration

vaibhavmurkute:
Revised in https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/150

PR #144, opened 2022-09-30
fix: ensure we don't allow subscribers to get rotated if the ca setup is not ready

Nelsonochoam:
**Description of changes:**
Given that now a customer can provide a configuration for CDA, it might happen that they provide invalid values upon initial setup. If we other components try to subscribe before a CA has properly been setup they will just fail to subscribe multiple times in a row causing them to go into a broken state.

This change just makes it so that if a CA has not been setup we don't trigger the subscriptions. When the CA is configured properly we emit an event that will cause them to rotate their certificates instead.

Part of increasing our component reliability and fault tolerance. Next step is making sure we can recover from failures when invalid configuration values are provided.

* Wrote a test and manually deployed the changes to test everything was still working


MikeDombo:
> other components try to subscribe before a CA has properly been setup they will just fail to subscribe multiple times in a row

Why do they fail to subscribe? Surely they should be able to subscribe and then just get the certificates if and when the configuration is correct, right?

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -218,12 +218,10 @@ private void subscribeToServerCertificateUpdatesNoCSR(@NonNull GetCertificateReq
         cisShadowMonitor.addToMonitor(certificateGenerator);
         caConfigurationMonitor.addToMonitor(certificateGenerator);
 
-        // TODO: Doing this here is wrong. We are assuming that we can call this and it would work but this should
-        //  only be called after we are sure a certificate authority has been configured. Which might not be true
-        //  before this gets called. Having this fail several times in a row (if no CA has been configured) will cause
-        //  downstream components that rely on this to fail
-        certificateGenerator.generateCertificate(connectivityInformation::getCachedHostAddresses,
-                "initialization of server cert subscription");
+        if (certificateStore.isReady()) {
```
MikeDombo, line 221: this call is happening when creating the subscription, so when will the cert get generated now that this code may not execute?


Also, if this code will not execute sometimes, why not remove it completely and let the cert get generated by whatever workflow is fixing it in the case when `isReady()` is false?

Nelsonochoam: If customers use a config to provide their own certificate and launch the component for the first time. If they provide a wrong path for the cert in the config we fail to load it and no CA is configured. Still, components will call this method which will in turn raise an exception because there is no CA, causing them to go eventually into a broken state.

To your point, we could completely remove the conditionals once we refactor all this subscription stuff to work reacting to events.

MikeDombo: So you described the current state I believe. But what happens with this change, now that generate certificates isn't called?

How do the certs eventually get generated?

Nelsonochoam: When we set the ca chain with the pk, we emit an event. We have a handler that listens to it and triggers rotation every time a new CA is configured

MikeDombo (approved):

vaibhavmurkute (approved):

PR #145, opened 2022-09-30
feat(wip): verify iot certificate usecase

jbutler:
**Issue #, if available:**

**Description of changes:**
* Introduce verify iot certificate usecase
* compute certificate ID based on encoded certificate instead of PEM
* enable offline authN of previously authenticated certificates

* manual testing

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
@@ -8,18 +8,77 @@ package com.aws.greengrass.clientdevices.auth.iot;
 import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
 import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
 import com.aws.greengrass.clientdevices.auth.session.attribute.StringLiteralAttribute;
-import lombok.NonNull;
-import lombok.Value;
+import lombok.Getter;
 
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Map;
 
-@Value
+@Getter
 public class Certificate implements AttributeProvider {
     public static final String NAMESPACE = "Certificate";
 
-    @NonNull
-    String iotCertificateId; // Needed for certificate revocation
+    public enum Status {
+        ACTIVE,
+        INACTIVE
+    }
+
+    String iotCertificateId;
+    Status status;
+    Instant lastUpdated;
+
+
+    // TODO: Needed?
+    public Certificate(String iotCertificateId) {
+        this(iotCertificateId, Status.INACTIVE);
+    }
+
+    /**
+     * Construct certificate using the current time as last updated.
+     * @param iotCertificateId Certificate ID
+     * @param status           Certificate status
+     */
+    public Certificate(String iotCertificateId, Status status) {
+        this(iotCertificateId, status, Instant.MIN);
+    }
+
+    /**
+     * Construct Certificate.
+     * @param iotCertificateId Certificate ID
+     * @param status           Certificate status
+     * @param lastUpdated      Certificate last updated timestamp
+     */
+    public Certificate(String iotCertificateId, Status status, Instant lastUpdated) {
+        this.iotCertificateId = iotCertificateId;
+        this.status = status;
+        this.lastUpdated = lastUpdated;
+    }
+
+    /**
+     * Update certificate status as of the current time.
+     * @param status Certificate status
+     */
+    public void updateStatus(Status status) {
+        updateStatus(status, Instant.now());
+    }
+
+    /**
+     * Update certificate status as of the provided time.
+     * @param status      Certificate status
+     * @param lastUpdated Timestamp
+     */
+    public void updateStatus(Status status, Instant lastUpdated) {
+        this.status = status;
+        this.lastUpdated = lastUpdated;
+    }
+
+    /**
+     * Check certificate status.
+     * @return true if this certificate is active in IoT Core.
+     */
+    public boolean isActive() {
+        return status == Status.ACTIVE;
+    }
 
     @Override
     public String getNamespace() {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
@@ -5,155 +5,81 @@
 
 package com.aws.greengrass.clientdevices.auth.iot.registry;
 
-import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
-import com.aws.greengrass.clientdevices.auth.iot.IotAuthClient;
+import com.aws.greengrass.clientdevices.auth.iot.Certificate;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
-import com.aws.greengrass.util.Digest;
-import com.aws.greengrass.util.Utils;
+import org.bouncycastle.util.encoders.Hex;
+import software.amazon.awssdk.utils.StringInputStream;
 
+import java.io.UnsupportedEncodingException;
+import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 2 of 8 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -8,18 +8,77 @@
...
+        ACTIVE,
+        INACTIVE
+    }
+
+    String iotCertificateId;
+    Status status;
+    Instant lastUpdated;
+
+
+    // TODO: Needed?
```
jbutler, line 31: Will clean this up. Right now there are some tests which instantiate certificates directly

src/test/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistryTest.java
```diff
@@ -8,7 +8,7 @@
 import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
 import com.aws.greengrass.clientdevices.auth.iot.IotAuthClient;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
```
jbutler, line 11: need to re-write these tests still. don't worry, i'll get to it

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -5,155 +5,81 @@
...
-            Optional<String> certId = getAssociatedCertificateId(certificatePem).map(Optional::of)
-                    .orElseGet(() -> fetchActiveCertificateId(certificatePem));
-            updateRegistryForCertificate(certificatePem, certId);
-            return certId;
-        } catch (CloudServiceInteractionException e) {
-            return getAssociatedCertificateId(certificatePem).map(Optional::of)
-                    .orElseThrow(() -> e);
+    public Certificate getCertificateById(String certificateId) {
+        Instant lastValid = registry.get(certificateId);
+        if (lastValid != null) {
```
Nelsonochoam, line 54: I don't know about this. This is not actually doing a lookup of a certificate by its id on the registry it is more of creating a certificate and having some business logic (It knows that is last valid is not null the status should be "ACTIVE"). Ideally that logic is part of the certificate itself and we store certificates in the repository. Maybe what should be stored is a `Map<String, Certificate>` and then this would

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -5,155 +5,81 @@
...
-     * @param certificatePem Certificate PEM
-     * @param iotCertificateId Optional of IoT Certificate ID
-     */
-    private void updateRegistryForCertificate(String certificatePem, Optional<String> iotCertificateId) {
-        if (iotCertificateId.isPresent()) {
-            registerCertificateIdForPem(iotCertificateId.get(), certificatePem);
+    public void updateCertificate(Certificate certificate) {
+        // NOTE: The storage aspect of this will change soon. But for now
+        //  only store ACTIVE certificates in the registry
+        if (certificate.getStatus() == Certificate.Status.ACTIVE) {
```
Nelsonochoam, line 67: Similar to my comment above this logic is making the repository, registry a bit smart and doing an unexpected operation. When I call update on something I wouldn't expect it to get removed from storage

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -5,155 +5,81 @@
...
-                .ifPresent(registry::remove);
-    }
-
-    /**
-     * Returns SHA-256 hash of given CertificatePem.
-     * @param certificatePem certificate pem
-     * @return Certificate hash or null if hash could not be calculated
-     */
-    private String getCertificateHash(String certificatePem) {
+    private String getCertificateId(String certificatePem) {
```
Nelsonochoam, line 74: This is ok as long as we don't expose it as public. Ideally ids should get set to `Certificate` objects when they are stored. I was thinking we would have `Certificate` objects hold a reference to their pem file and call something like

```java
Certificate cert = certificateRegistry.create(String pem); 
// create would compute the id
// Create a certificate instance with that id and store it
// return the instance
```

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param networkState        Network state infrastructure
+     * @param iotAuthClient       IoT dataplane API client
+     * @param certificateRegistry cert registry
+     */
+    @Inject
+    public VerifyIotCertificate(NetworkState networkState, IotAuthClient iotAuthClient,
+                                CertificateRegistry certificateRegistry) {
+        this.networkState = networkState;
+        this.iotAuthClient = iotAuthClient;
+        this.certificateRegistry = certificateRegistry;
```
Nelsonochoam, line 42: This is just a comment but ideally we would inject into this a `CertificateRepository` which would just be an interface and have `certificateRegistry` adhere to that interface then we could swap the implementation any time without having to modify everywhere we inject this

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public Result apply(String certificatePem) {
+        // If we think we are online, the certificate is not ACTIVE, or if certificate
+        // status information is stale, then go to the cloud.
+        // This ensures that we eventually go to IoT Core even if there are shenanigans
+        // going on where we have HTTP connectivity but MQTT is disconnected.
+        String source = "local";
+        Certificate cert = certificateRegistry.getCertificateByPem(certificatePem);
```
Nelsonochoam, line 67: Related to some of the comments above this would just internally compute the `id` from the pem file and try to get a certificate directly from the Repository and if not present create a new one like

```java
Optional<Certificate> cert = certificateRepository.getByPem(certificatePem);
// the isStale method would have cert.getStatus() != Certificate.Status.ACTIVE || !isTimestampRecent(cert) check

if (networkState.isUp() || certificate.isStale()) {
          // This would return a certificate that has source "cloud"
          Optional<Certificate> Certificate = getCertificateIdFromIot(certificatePem);
         // etc etc...
}
```

The main problem I see with this is that the Certificate object is not encapsulating business logic we have scattered methods that determine its logic across the Registry and this useCase

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
jbutler: Don't worry about the storage bits. Same comment as thing registry. This is all changing anyway.
But fair point about keeping business logic out of this. Agree there

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
jbutler: Ack. Can add a remove API

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
jbutler: The id or the method?

ID _is_ the public interface. I'm aligning with IoT core here instead of inventing out own thing

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistry.java
```diff
@@ -5,155 +5,81 @@
...
-            Optional<String> certId = fetchActiveCertificateId(certificatePem);
-            updateRegistryForCertificate(certificatePem, certId);
-            return certId.isPresent();
-        } catch (CloudServiceInteractionException e) {
-            return getAssociatedCertificateId(certificatePem)
-                    .map(certId -> true)
-                    .orElseThrow(() -> e);
-        }
+    public Certificate getCertificateByPem(String certificatePem) {
+        return getCertificateById(getCertificateId(certificatePem));
```
vaibhavmurkute, line 44: Possible NPE here. `getCertificateId()` returns `null` in case of `NoSuchAlgorithmException | CertificateException | UnsupportedEncodingException` while calculating certificate id.

PR #146, opened 2022-09-30
feat: improve byoca reliability, recover from bad config

Nelsonochoam:
**Description of changes:**
This PR improves the reliability of BYOCA by allowing CDA to recover from a bad configuration provided by a customer. For example if a customer was running a previous version of CDA and they switch to managed with a bad config, the service will keep on running with the old config and report the errors on the logs, once the customer fixes the configuration and it becomes valid, the changes will get applied.

There are still issues that may arise that we don't recover from like a customer providing and invalid private key for a certificate that they provide. In those scenarios the service will break. We could improve that in the future and further check that the key they provided is valid for the certificate they provided before setting it as the CA

Part of improving the customer experience is allowing them to easily recover from mistakes and allowing the software to keep running even if there are faults.

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
@@ -64,6 +64,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
@@ -133,20 +134,30 @@ public class CustomCaConfigurationTest {
         kernel.shutdown();
     }
 
-    // TODO: Consolidate this test helpers with ClientDevicesAuthServiceTest
-    private void givenNucleusRunningWithConfig(String configFileName) throws InterruptedException {
+    private void givenNucleusRunningWithConfig(String configFileName, Consumer<State> consumer) throws InterruptedException {
         CountDownLatch authServiceRunning = new CountDownLatch(1);
         kernel.parseArgs("-r", rootDir.toAbsolutePath().toString(), "-i",
                 getClass().getResource(configFileName).toString());
         kernel.getContext().addGlobalStateChangeListener((service, was, newState) -> {
-            if (ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME.equals(service.getName()) && service.getState()
-                    .equals(State.RUNNING)) {
-                authServiceRunning.countDown();
+            if (ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME.equals(service.getName())) {
+                State serviceState = service.getState();
+                consumer.accept(serviceState);
+
+                if (serviceState.equals(State.RUNNING)) {
+                    authServiceRunning.countDown();
+                }
+
             }
         });
         kernel.launch();
+
         assertThat(authServiceRunning.await(30L, TimeUnit.SECONDS), is(true));
     }
+
+    private void givenNucleusRunningWithConfig(String configFileName) throws InterruptedException {
+        givenNucleusRunningWithConfig(configFileName, (State s) -> {});
+    }
+
 
     private static Pair<X509Certificate[], KeyPair[]> givenRootAndIntermediateCA() throws NoSuchAlgorithmException,
             CertificateException,
@@ -167,7 +178,7 @@ public class CustomCaConfigurationTest {
     /**
      * Simulates an external party configuring their custom CA.
      */
-    private void givenCDAWithCustomCertificateAuthority(URI privateKeyUri, URI certificateUri) throws
+    private void configureCDAWithCustomCertificateAuthority(URI privateKeyUri, URI certificateUri) throws
             ServiceLoadException {
         // Service Configuration
         Topics topics = kernel.locate(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME).getConfig();
@@ -219,7 +230,7 @@ public class CustomCaConfigurationTest {
                 GetCertificateRequestOptions.CertificateType.CLIENT, asyncCall.getRight());
 
         givenNucleusRunningWithConfig("config.yaml");
-        givenCDAWithCustomCertificateAuthority(privateKeyUri, certificateUri);
+        configureCDAWithCustomCertificateAuthority(privateKeyUri, certificateUri);
         subscribeToCertificateUpdates(request);
 
         asyncCall.getLeft().get(1, TimeUnit.SECONDS);
@@ -249,7 +260,7 @@ public class CustomCaConfigurationTest {
                 GetCertificateRequestOptions.CertificateType.CLIENT, asyncCall.getRight());
 
         givenNucleusRunningWithConfig("config.yaml");
-        givenCDAWithCustomCertificateAuthority(privateKeyUri, certificateUri);
+        configureCDAWithCustomCertificateAuthority(privateKeyUri, certificateUri);
         subscribeToCertificateUpdates(request);
 
         asyncCall.getLeft().get(1, TimeUnit.SECONDS);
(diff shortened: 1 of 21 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
```diff
@@ -27,28 +38,49 @@ public static <V> Result<V> ok() {
         return ok(null);
     }
 
-    public static <E extends Exception> Result<E> error(E value) {
-        return new Result<>(Status.ERROR, value);
+    public static <T, E extends Exception> Result<T> error(E err) {
```
Nelsonochoam, line 41: This changes to result are such that we apply generic more easily and do things like

```java
Result<CAConfiguration> someMethod() {
     if (badThing) {
          Result.error(new SomeException("blah");
     }

     return Result.ok(CAConfiguration);
}
```

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
```diff
@@ -58,18 +62,36 @@ private CAConfiguration(List<String> caTypes, CertificateStore.CAType caType,
      * Factory method for creating an immutable CAConfiguration from the service configuration.
      *
      * @param configurationTopics the configuration key of the service configuration
-     *
-     * @throws URISyntaxException if invalid certificateUri or privateKeyUri provided.
      */
-    public static CAConfiguration from(Topics configurationTopics) throws URISyntaxException {
+    public static Result<CAConfiguration> from(Topics configurationTopics) {
```
Nelsonochoam, line 66: With this changes we are being more specific about what went wrong in the configuration and reflect what is a valid configuration.

src/test/java/com/aws/greengrass/clientdevices/auth/api/UseCasesTest.java
```diff
@@ -65,20 +64,6 @@ public Result<String> apply(String dto) {
         }
     }
 
-    static class UseCaseUpdatingDependency implements UseCases.UseCase<String, Void> {
```
Nelsonochoam, line 68: Was not being used

jbutler:
> For example if a customer was running a previous version of CDA and they switch to managed with a bad config, the service will keep on running with the old config and report the errors on the logs, once the customer fixes the configuration and it becomes valid, the changes will get applied.

I haven't looked at the change yet, but are you actually discarding configuration if it's bad? Logs are not an appropriate mechanism to convey errors to customers. They're there to understand failures after they've been discovered.

In this case of bad CA configuration we absolutely need to fail deployment.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -131,15 +133,31 @@ private void initializeHandlers() {
...
         onConfigurationChanged();
         config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe(this::configChangeHandler);
     }
 
     private void onConfigurationChanged() {
-        try {
-            cdaConfiguration = CDAConfiguration.from(cdaConfiguration, getConfig());
-        } catch (URISyntaxException e) {
-            serviceErrored(e);
+        Result<CDAConfiguration> result = CDAConfiguration.from(cdaConfiguration, getConfig());
```
Nelsonochoam, line 142: This will return a result with a configuration regardless of wether it has errored or not, we want to track the configuration even if it is bad so that we can check if it has changed from a bad one to a good one or from a bad one to another bad one.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -131,15 +133,31 @@ private void initializeHandlers() {
...
+        CDAConfiguration configuration = result.get();
+
+        // Nothing has changed
+        if (configuration.hasChanged(cdaConfiguration)) {
```
Nelsonochoam, line 146: The onConfigurationChanged method might be called multiple times due to the nature of how the Nucleus emits multiple events per key changed on the configuration

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -131,15 +133,31 @@ private void initializeHandlers() {
...
+            return;
+        }
+
+        cdaConfiguration = configuration;
+
+        if (result.isOk() && inState(State.BROKEN)) {
```
Nelsonochoam, line 152: If the new configuration we are getting is ok but we were broken due to a bad config we try to re install.

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -310,11 +326,94 @@ void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serve
...
+        X509Certificate[] chain = credentials.getLeft();
+        KeyPair[] certificateKeys = credentials.getRight();
+        KeyPair intermediateKeyPair = certificateKeys[0];
+
+        CountDownLatch authServiceBroken = new CountDownLatch(1);
+        CountDownLatch recoveredFromBroken = new CountDownLatch(1);
+        AtomicBoolean wasBroken = new AtomicBoolean(false);
+        Consumer<State> serviceStateChangeListener = (State s) -> {
+            System.out.println(s);
+            System.out.println(wasBroken.get() && s.equals(State.RUNNING));
```
vaibhavmurkute, line 388: need to remove these prints?

Nelsonochoam: Haha great catch

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -131,15 +133,31 @@ private void initializeHandlers() {
...
+        if (configuration.isEqual(cdaConfiguration)) {
+            return;
+        }
+
+        cdaConfiguration = configuration;
+
+        if (result.isOk() && inState(State.BROKEN)) {
+            logger.info("Service is {} and configuration changed. Attempting to reinstall.", State.BROKEN);
+            requestReinstall();
+            return;
```
vaibhavmurkute, line 155: I may have misunderstood it, but it seems like we're _silently_ reinstalling the service when bad configuration. Customer may not look at this log when we silently rollback the service. Should we just keep the service in broken state and fail the deployment?

Nelsonochoam: We are not silently reinstalling. Basically every time the customer provides a wrong configuration we will call `serviceErrored` (on the section below). If they keep providing bad configuration values we will keep erroring (which will show the state and errors on the logs). If they provide a bad configuration enough times, we will eventually go into a BROKEN state.

The only way of exiting out of a BROKEN state is re installing the component. Hence, if we are in a broken state and the customer does a deployment with a new working config we basically tell it to reinstall. Without this step, (and this is also how the current version of CDA in production works), if the component goes into a broken state, deploying a configuration change won't simply fix it, the customers would have to manually restart the component.

src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
```diff
@@ -27,28 +39,56 @@ public static <V> Result<V> ok() {
...
 
     public static <V> Result<V> warning() {
         return new Result<>(Status.WARNING, null);
     }
 
-
+    /**
+     * If a value is present in this Result, returns the value, otherwise throws NoSuchElementException.
+     *
+     * @throws NoSuchElementException when no value present
```
jbutler, line 62: Does it? It looks like it just returns null

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
jbutler: Hmmm.. this seems unnecessarily complicated. How does it convey more specific information back to the caller?
I'd prefer just returning a `CAConfiguration` or throwing here. This is kind of a weird pattern to have a factory method return a Result.

src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
```diff
@@ -27,28 +39,56 @@ public static <V> Result<V> ok() {
...
 
     public boolean isOk() {
         return status == Status.OK;
     }
+
+    /**
+     * Calls a consumer with the stored value if the Result isOk.
+     * @param consumer a consumer function.
+     */
+    public void ifOk(Consumer<T> consumer) {
```
jbutler, line 84: These don't seem necessary. A simple if statement is easier to grok

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
jbutler: It doesn't seem like most of these changes are necessary. There's really no need to track bad configuration, is there? If configuration is null and state is broken, request reinstall.

PR #148, opened 2022-10-04
feat: add certificate domain logic

jbutler:
**Issue #, if available:**

**Description of changes:**
Revision of #145 which removes the registry changes and focuses primarily on Certificate.

Future revisions will flesh out the Registry, as well as add better test support so Registry changes don't break our integ tests (which they currently do).

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
@@ -12,7 +12,8 @@ import lombok.Getter;
 import org.bouncycastle.util.encoders.Hex;
 
 import java.io.ByteArrayInputStream;
-import java.io.UnsupportedEncodingException;
+import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -58,7 +59,7 @@ public class Certificate implements AttributeProvider {
         try {
             String certificateId = computeCertificateId(certificatePem);
             return new Certificate(certificateId);
-        } catch (CertificateException | NoSuchAlgorithmException | UnsupportedEncodingException e) {
+        } catch (CertificateException | NoSuchAlgorithmException | IOException e) {
             throw new InvalidCertificateException("Unable to parse certificate PEM", e);
         }
     }
@@ -109,12 +110,13 @@ public class Certificate implements AttributeProvider {
     }
 
     private static String computeCertificateId(String certificatePem)
-            throws CertificateException, NoSuchAlgorithmException, UnsupportedEncodingException {
+            throws CertificateException, NoSuchAlgorithmException, IOException {
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
-        X509Certificate cert = (X509Certificate) cf.generateCertificate(
-                new ByteArrayInputStream(certificatePem.getBytes(StandardCharsets.UTF_8)));
-        MessageDigest digest = MessageDigest.getInstance("SHA-256");
-        digest.reset();
-        return new String(Hex.encode(digest.digest(cert.getEncoded())), "UTF-8");
+        try (InputStream is = new ByteArrayInputStream(certificatePem.getBytes(StandardCharsets.UTF_8))) {
+            X509Certificate cert = (X509Certificate) cf.generateCertificate(is);
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            digest.reset();
+            return new String(Hex.encode(digest.digest(cert.getEncoded())), StandardCharsets.UTF_8);
+        }
     }
 }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
@@ -8,18 +8,95 @@ package com.aws.greengrass.clientdevices.auth.iot;
 import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
 import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
 import com.aws.greengrass.clientdevices.auth.session.attribute.StringLiteralAttribute;
-import lombok.NonNull;
-import lombok.Value;
+import lombok.Getter;
+import org.bouncycastle.util.encoders.Hex;
 
+import java.io.ByteArrayInputStream;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Map;
 
-@Value
+@Getter
 public class Certificate implements AttributeProvider {
     public static final String NAMESPACE = "Certificate";
 
-    @NonNull
-    String iotCertificateId; // Needed for certificate revocation
+    public enum Status {
+        ACTIVE,
+        INACTIVE,
+        UNKNOWN
+    }
+
+    String certificateId;
+    Status status;
+    Instant lastUpdated;
+
+
+    /**
+     * TODO: Make this private.
+     * @param certificateId certificate ID
+     */
+    public Certificate(String certificateId) {
+        this.certificateId = certificateId;
+        this.status = Status.UNKNOWN;
+        this.lastUpdated = Instant.MIN;
+    }
+
+    /**
+     * Factory method to construct a Certificate object from certificate PEM.
+     * @param certificatePem Certificate PEM
+     * @return Certificate
+     * @throws InvalidCertificateException if certificate PEM is invalid
+     */
+    public static Certificate fromPem(String certificatePem) throws InvalidCertificateException {
+        try {
+            String certificateId = computeCertificateId(certificatePem);
+            return new Certificate(certificateId);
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -8,18 +8,94 @@
 import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
 import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
 import com.aws.greengrass.clientdevices.auth.session.attribute.StringLiteralAttribute;
-import lombok.NonNull;
-import lombok.Value;
+import lombok.Getter;
+import org.bouncycastle.util.encoders.Hex;
+import software.amazon.awssdk.utils.StringInputStream;
```
jbutler, line 13: Going to see if I can replace these with vanilla java

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -53,50 +56,81 @@ public Optional<String> getActiveCertificateId(String certificatePem) {
...
                                 + "policy grants the greengrass:VerifyClientDeviceIdentity permission.");
                 throw new CloudServiceInteractionException("Failed to verify client device identity", e);
             }
         }
 
+        @Override
+        @SuppressWarnings("PMD.AvoidCatchingGenericException")
+        public Certificate getIotCertificate(String certificatePem) throws InvalidCertificateException {
+            if (Utils.isEmpty(certificatePem)) {
+                throw new IllegalArgumentException("Certificate PEM is empty");
```
jbutler, line 74: Woops - can be InvalidCertificateException

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -0,0 +1,50 @@
...
+    @Inject
+    public VerifyIotCertificate(CertificateRegistry certificateRegistry) {
+        this.certificateRegistry = certificateRegistry;
+    }
+
+    @Override
+    public Result<Boolean> apply(String certificatePem) {
+        return Result.ok(certificateRegistry.isCertificateValid(certificatePem));
+
+        /*
```
jbutler, line 32: This is the next step logic. Unfortunately doing this breaks a lot of fragile tests which need to be fixed first.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -28,6 +105,16 @@ public String getNamespace() {
...
+    }
+
+    private static String computeCertificateId(String certificatePem)
+            throws CertificateException, NoSuchAlgorithmException, UnsupportedEncodingException {
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        X509Certificate cert = (X509Certificate) cf.generateCertificate(
+                new ByteArrayInputStream(certificatePem.getBytes(StandardCharsets.UTF_8)));
+        MessageDigest digest = MessageDigest.getInstance("SHA-256");
+        digest.reset();
+        return new String(Hex.encode(digest.digest(cert.getEncoded())), "UTF-8");
```
vaibhavmurkute, line 118: `Digest.calculate()` from Nucleus does exactly this. Can we reuse that?

jbutler: It doesn't though. That base64 encodes. This hex encodes

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -53,50 +56,81 @@ public Optional<String> getActiveCertificateId(String certificatePem) {
...
+            VerifyClientDeviceIdentityRequest request =
+                    VerifyClientDeviceIdentityRequest.builder().clientDeviceCertificate(certificatePem).build();
+            try {
+                // We can ignore the response since it contains only the cert ID, which we directly compute
+                clientFactory.getGreengrassV2DataClient().verifyClientDeviceIdentity(request);
+                cert.setStatus(Certificate.Status.ACTIVE);
+            } catch (ValidationException | ResourceNotFoundException e) {
+                logger.atWarn().cause(e).kv(CERTPEM_KEY, certificatePem)
+                        .log("Certificate doesn't exist or isn't active");
+                throw new InvalidCertificateException("Invalid IoT certificate", e);
```
vaibhavmurkute, line 89: Instead of throwing `InvalidCertificateException`, update cert status to `INACTIVE` here?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -28,6 +105,16 @@ public String getNamespace() {
...
     public Map<String, DeviceAttribute> getDeviceAttributes() {
-        return Collections.singletonMap("CertificateId", new StringLiteralAttribute(getIotCertificateId()));
+        return Collections.singletonMap("CertificateId", new StringLiteralAttribute(getCertificateId()));
+    }
+
+    private static String computeCertificateId(String certificatePem)
+            throws CertificateException, NoSuchAlgorithmException, UnsupportedEncodingException {
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        X509Certificate cert = (X509Certificate) cf.generateCertificate(
+                new ByteArrayInputStream(certificatePem.getBytes(StandardCharsets.UTF_8)));
```
MikeDombo, line 115: need to close the stream?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
MikeDombo, line 74: checked exception

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -53,50 +56,81 @@ public Optional<String> getActiveCertificateId(String certificatePem) {
...
+            } catch (Exception e) {
+                // TODO: don't log at error level for network failures
+                logger.atError().cause(e).kv(CERTPEM_KEY, certificatePem)
+                        .log("Failed to verify client device identity with cloud. Check that the core device's IoT "
+                                + "policy grants the greengrass:VerifyClientDeviceIdentity permission.");
```
MikeDombo, line 94: no period at the end of log messages

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -0,0 +1,50 @@
...
+            return Result.ok(false);
+        }
+
+        logger.atDebug().kv("CertificateID", iotCert.getCertificateId())
+                .kv("status", iotCert.getStatus())
+                .kv("source", source)
+                .log("Verifying client device certificate");
+
+        return Result.ok(iotCert.isActive());
+         */
```
vaibhavmurkute, line 48: Need to remove this block?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
jbutler: `ByteArrayInputStream` is GC'd without needing to be closed. But, probably best to do it just in case the stream type changes in the future

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -8,18 +8,95 @@
...
+    public boolean isActive() {
+        return status == Status.ACTIVE;
+    }
+
+    /**
+     * Retrieve certificate PEM.
+     * @return certificate PEM as a UTF-8 encoded string
+     * @throws UnsupportedOperationException since this is not yet supported
+     */
+    public String getCertificatePem() {
```
Nelsonochoam, line 98: Is this a not supported or not implemented? - Why even have this method then?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
jbutler: That implies the certificate exists in the customer account, which is not the case.

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
Nelsonochoam: Mmmm I think there is a distinction between a certificate not existing or not being active. Following @vaibhavmurkute comment if it is not active it might be better to just set it as inactive and throw an error if it is not found. 

I know also this might be a debated suggestion but if would be nice to model this in terms of states. I see in your commented code that we catch it so it is something that is OK to happen. Why not return a result instead?.  So

* If it is empty return just `Result.ok()` you could add a method on the Result called `isEmpty` if you'd like (that represent an ok state but nothing returned.
* For the scenario of an invalid certificate maybe  having `cert.setStatus(Invalid)` `Result.ok(cert)` better represents the state than an exception
* Finally if it is valid we `cert.setStatus(Active)` return `Result.ok(cert)`

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
jbutler: I'm open to making tweaks to this in the future, but for now I don't want to model this if we don't get a positive response. The cloud API doesn't give us enough information to do this.

PR #149, opened 2022-10-05
feat: certificate registry

jbutler:
**Issue #, if available:**

**Description of changes:**
This change removes IoT Core logic from the Certificate Registry -- it now serves as a repository for certificates.

The VerifyIotCertificate use case now inserts certificates into the registry after it validates them.

The MqttSessionFactory retrieves certificates from the Certificate Registry, which removes one cloud API call. However, it is still calling IoT Core to validate  Thing <-> Certificate attachment. Removing that can come later.

This change also introduces a new CertificateFake object helps somewhat with tests, as it allows us to instantiate Certificates without a valid PEM. There are also now test utils for generating client certs. We can slowly improve our test quality over time by migrating to these. Currently the tests are still quite fragile.

Manual / unit tests

### DIFF
--- a/src/test/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistryTest.java
+++ b/src/test/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistryTest.java
@@ -5,114 +5,116 @@
 
 package com.aws.greengrass.clientdevices.auth.iot.registry;
 
-import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
-import com.aws.greengrass.clientdevices.auth.iot.IotAuthClient;
+import com.aws.greengrass.clientdevices.auth.certificate.CertificateHelper;
+import com.aws.greengrass.clientdevices.auth.certificate.CertificateStore;
+import com.aws.greengrass.clientdevices.auth.helpers.CertificateTestHelpers;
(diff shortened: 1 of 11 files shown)


src/test/java/com/aws/greengrass/clientdevices/auth/iot/registry/CertificateRegistryTest.java
```diff
@@ -5,114 +5,116 @@
...
+            KeyPair rootKeyPair = CertificateStore.newRSAKeyPair(2048);
+            KeyPair clientKeyPair = CertificateStore.newRSAKeyPair(2048);
+            X509Certificate rootCA = CertificateTestHelpers.createRootCertificateAuthority("root", rootKeyPair);
+
+            validClientCertificate = CertificateTestHelpers.createClientCertificate(rootCA, "Client", clientKeyPair.getPublic(),
+                    rootKeyPair.getPrivate());
+            validClientCertificatePem = CertificateHelper.toPem(validClientCertificate);
+        } catch (NoSuchAlgorithmException | CertificateException | OperatorCreationException | IOException e) {
+            Assertions.fail("Unable to initialize test certificates!");
+        }
```
MikeDombo, line 55: just let this method throw, it will do the same thing, without swallowing the exception

MikeDombo (dismissed):
LGTM

MikeDombo (approved):

PR #150, opened 2022-10-05
feat: add Thing configuration store adapter

vaibhavmurkute:
**Description of changes:** 
- Adds Thing configuration store adapter for ThingRegistry to store and retrieve client device Thing metadata from the Runtime config. Updates to runtime config are flushed to the disk immediately. Newly added runtime configuration has following structure:
```
 * |---- runtime
 * |    |---- clientDeviceThings:
 * |          |---- <schemaVersion>:
 * |                   |---- <thingName>:
 * |                            |---- version: "..."
 * |                            |---- attachedCertificateIds: [...]
 * |
```
- This config schema is versioned for better backward compatibility with future schema changes.
- ThingRegistry writes to or reads directly off of the runtime config store. Metadata written to the config store is opaque to the Config store. 

**Why is this change necessary:** To persist Thing metadata across device power-cycle or restart.

**How was this change tested:** unit tests, integ tests (`mvn package`, `mvn verify`)

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
@@ -21,8 +21,9 @@ import java.util.Map;
  * |    |---- certificates:
  * |         |---- authorities: [...]
  * |    |---- clientDeviceThings:
- * |          |---- thingName:
- * |                |---- key: value
+ * |          |---- schemaVersion:
+ * |                |---- thingName:
+ * |                      |---- key: value
  * </p>
  */
 public final class RuntimeConfiguration {
@@ -73,31 +74,37 @@ public final class RuntimeConfiguration {
     /**
      * Retrieves opaque client device thing configuration.
      *
-     * @param thingName Thing name
+     * @param thingName     Thing name
+     * @param schemaVersion Thing config schema version
      * @return config topic mapped to given thing name
      */
-    public Topics getClientDeviceThing(String thingName) {
-        return config.findTopics(CLIENT_DEVICE_THINGS_KEY, thingName);
+    public Topics getClientDeviceThing(String thingName, String schemaVersion) {
+        return config.findTopics(CLIENT_DEVICE_THINGS_KEY, schemaVersion, thingName);
     }
 
     /**
      * Create or replace client device thing details in the config.
-     * @param thingName    client device thing name
-     * @param thingDetails thing details
+     *
+     * @param thingName     client device thing name
+     * @param thingDetails  thing details
+     * @param schemaVersion Thing config schema version
      */
-    public void createOrReplaceClientDeviceThing(String thingName, Map<String, Object> thingDetails) {
-        Topics clientDeviceThing = config.lookupTopics(CLIENT_DEVICE_THINGS_KEY, thingName);
+    public void createOrReplaceClientDeviceThing(String thingName,
+                                                 Map<String, Object> thingDetails,
+                                                 String schemaVersion) {
+        Topics clientDeviceThing = config.lookupTopics(CLIENT_DEVICE_THINGS_KEY, schemaVersion, thingName);
         clientDeviceThing.replaceAndWait(thingDetails);
     }
 
     /**
      * Update client device thing details in the config.
      *
-     * @param thingName    client device Thing name
-     * @param thingDetails map of client device Thing details to be updated
+     * @param thingName     client device Thing name
+     * @param thingDetails  map of client device Thing details to be updated
+     * @param schemaVersion Thing config schema version
      */
-    public void updateClientDeviceThing(String thingName, Map<String, Object> thingDetails) {
-        Topics clientDeviceThing = config.findTopics(CLIENT_DEVICE_THINGS_KEY, thingName);
+    public void updateClientDeviceThing(String thingName, Map<String, Object> thingDetails, String schemaVersion) {
+        Topics clientDeviceThing = config.findTopics(CLIENT_DEVICE_THINGS_KEY, schemaVersion, thingName);
         if (clientDeviceThing == null) {
             return;
         }
@@ -108,7 +115,7 @@ public final class RuntimeConfiguration {
     private void updateClientDeviceThingDetail(Topics clientDeviceThing, String detailKey, Object newValue) {
         Topic thingDetail = clientDeviceThing.lookup(detailKey);
         IllegalArgumentException ex = new IllegalArgumentException(
-                String.format("Unsupported thing detail value %s", newValue));
+                String.format("Unsupported thing detail value: %s", newValue));
         if (newValue instanceof Integer) {
             thingDetail.withValue((Integer) newValue);
         } else if (newValue instanceof String) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -66,4 +72,31 @@ public void updateCAPassphrase(String passphrase) {
...
+    /**
+     * Update (or create if not available) opaque client device thing details in the config.
+     *
+     * @param thingName    client device Thing name
+     * @param thingDetails opaque client device Thing details
+     */
+    public void updateClientDeviceThing(String thingName, Map<String, Object> thingDetails) {
+        Topics clientDeviceThings = config.lookupTopics(CLIENT_DEVICE_THINGS_KEY);
+        Topics clientDeviceThing = clientDeviceThings.lookupTopics(thingName);
+        clientDeviceThing.replaceAndWait(thingDetails);
```
MikeDombo, line 84: This can be expensive. I'd recommend only updating the parts that have actually changed.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -66,4 +72,31 @@ public void updateCAPassphrase(String passphrase) {
...
+    /**
+     * Retrieves opaque client device thing details from the config.
+     *
+     * @param thingName Thing name
+     * @return map of thing details
+     */
+    public Map<String, Object> getClientDeviceThing(String thingName) {
+        Topics clientDeviceThings = config.findTopics(CLIENT_DEVICE_THINGS_KEY);
+        if (clientDeviceThings == null || clientDeviceThings.findTopics(thingName) == null
+                || clientDeviceThings.findTopics(thingName).isEmpty()) {
```
MikeDombo, line 96: `Utils.isEmpty` would work

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -66,4 +72,31 @@ public void updateCAPassphrase(String passphrase) {
...
+            return Collections.emptyMap();
+        } else {
+            return clientDeviceThings.findTopics(thingName).toPOJO();
```
MikeDombo, line 99: Do we need to do this? This will be copying data into a map. Would be best if we don't need to duplicate data.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingConfigStoreAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return optional of client device Thing
+     */
+    Optional<Thing> getThing(String thingName) {
+        Map<String, Object> thingDetails = runtimeConfig.getClientDeviceThing(thingName);
+        if (thingDetails.isEmpty()) {
+            return Optional.empty();
+        }
+
+        Thing thing = Thing.of((int) thingDetails.get(THING_VERSION_KEY),
+                thingName, (List<String>) thingDetails.get(ATTACHED_CERTIFICATE_IDS_KEY));
```
MikeDombo, line 53: don't blindly cast, check first

Nelsonochoam: +1 to what Michael said. If think the runtime config should return a shape with this already validated (validate that there is a list of string) and then `Thing.of` can have the validation for business logic and things of sorts

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -66,4 +72,31 @@ public void updateCAPassphrase(String passphrase) {
...
+
+    /**
+     * Retrieves opaque client device thing details from the config.
+     *
+     * @param thingName Thing name
+     * @return map of thing details
+     */
+    public Map<String, Object> getClientDeviceThing(String thingName) {
+        Topics clientDeviceThings = config.findTopics(CLIENT_DEVICE_THINGS_KEY);
+        if (clientDeviceThings == null || clientDeviceThings.findTopics(thingName) == null
```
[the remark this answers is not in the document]
jbutler: Should be able to do it in one shot

```
config.findTopics(CLIENT_DEVICE_THINGS_KEY, thingName);
```

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingConfigStoreAdapter.java
jbutler: I am concerned about domain logic sneaking in to the `RuntimeConfiguration`. Considering that we'll also be storing certificate metadata, and that the information we're storing may evolve over time, I hesitate to say we should perform validation there.

I'm almost inclined to say that `RuntimeConfiguration` should just get/put `Topics` and that this class should perform the mapping from `Thing` to `Topics`. Thoughts?

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
vaibhavmurkute: Right, agreed. I added new api to update only the changed thing details: 
```
updateClientDeviceThing(String thingName, Map<String, Object> thingDetails, String schemaVersion) {}
```
Let me create a separate PR where adapter figures out what details changed and calls this RuntimeConfig api accordingly ^

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
vaibhavmurkute: right, refactored to return `Topics` directly, instead of a map.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingConfigStoreAdapter.java
vaibhavmurkute: I agree that domain logic should not sneak into `RuntimeConfiguration`. I refactored the adapter to deal with Topics directly and type-check (coerce) results appropriately. Also versioned the thing runtime config schema for better backward compatibility for future schema changes.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -66,4 +71,66 @@ public void updateCAPassphrase(String passphrase) {
...
+                    thingDetail.withValue((List<String>) newValue);
+                } else {
+                    throw ex;
+                }
+                break;
+            }
+        } else {
+            throw ex;
+        }
+    }
```
vaibhavmurkute, line 135: - This update api will let us update only the changed thing detail. Currently its not used, I'll create a separate PR for the adapter to figure out what thing details changed and call this api to only update that changed detail.
- Considering the `Topic` lets us update [only (int, String, boolean, List\<String\>) values](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/config/Topic.java#L139-L157), the `updateClientDeviceDetail()` logic may seem _somewhat_ ugly! I'm open to suggestions.

PR #151, opened 2022-10-05
feat: recover from configuration failures

Nelsonochoam:
**Description of changes:**
Enables CDA to recover from certificateAuthority configuration failures with requiring a manual restart.

Without this when a customer provides a bad config the component will error and can't recover without un installing and re installing

Integration test


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -136,11 +141,38 @@ private void subscribeToConfigChanges() {
...
 
     private void onConfigurationChanged() {
-        try {
-            cdaConfiguration = CDAConfiguration.from(cdaConfiguration, getConfig());
-        } catch (URISyntaxException e) {
-            serviceErrored(e);
+        // Note: The nucleus emits multiple configuration changed events, one per key that changed. It will also
+        // keep emitting them regardless of the state it is current in. If the configuration was incorrect, we want the
+        // service to error, but we don't want to check again until the nucleus has run the remediation steps (when the
+        // service errors, the nucleus will try to call shutdown -> install -> startup).
```
jbutler, line 147: Can you help me understand this? My mental model is that we shouldn't need to care about what Nucleus is doing, and that we can rely entirely on configuration updates. If we are broken, and a configuration change fixes us, then request reinstall? Are there scenarios where that will not work?

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -183,6 +215,13 @@ private void configChangeHandler(WhatHappened whatHappened, Node node) {
     protected void startup() throws InterruptedException {
         context.get(CertificateManager.class).startMonitors();
         super.startup();
+
+        synchronized (configLock) {
+            if (configurationErrored) {
+                configurationErrored = false;
+                onConfigurationChanged();
+            }
+        }
```
jbutler, line 224: I thought the issue was that if we are broken, then Nucleus _doesn't_ call `startup`?

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
Nelsonochoam: @jbutler and I chatted about this offline - we are going to dig deeper and see if it is required

PR #155, opened 2022-10-07
feat: persist things to runtime config store

vaibhavmurkute:
**Description of changes:** ThingRegistry now persists client device Thing metadata directly to the runtime config store.

**Why is this change necessary:** To enable offline auth capabilities across restarts

**How was this change tested:** Unit tests (`mvn package`, `mvn verify`)


src/test/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistryTest.java
```diff
@@ -100,7 +107,8 @@ void GIVEN_certificateWithUpdate_WHEN_updateCertificate_THEN_updatedCertificateI
         assertThat(cert.isPresent(), is(true));
         assertThat(cert.get().getCertificateId(), equalTo(newCert.getCertificateId()));
         assertThat(cert.get().getStatus(), equalTo(Certificate.Status.ACTIVE));
-        assertThat(cert.get().getStatusLastUpdated(), equalTo(now));
+        // asserting on epochMillis since Instant resolution varies with JDK versions
+        assertThat(cert.get().getStatusLastUpdated().toEpochMilli(), equalTo(now.toEpochMilli()));
```
vaibhavmurkute, line 111: On JDK 11:
```
Instant set : 2022-10-07T04:03:07.467315100Z (Epoch millis = 1665115387467)
Instant (parsed from Config store): 2022-10-07T04:03:07.467Z (Epoch millis = 1665115387467)
```

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -85,37 +85,43 @@ public Thing clone() {
...
+     * <p>
+     * This map should not be modified directly. Refer to {@link #attachCertificate(String) attachCertificate} and
      * {@link #detachCertificate(String) detachCertificate}
-     *
+     * </p>
      * @return Certificate IDs
      */
-    public List<String> getAttachedCertificateIds() {
-        return new ArrayList<>(attachedCertificateIds);
+    public Map<String, Instant> getAttachedCertificateIds() {
```
Nelsonochoam, line 113: I don't know if this can be improved, but this breaks encapsulation and exposes how we internally structure this class. Ideally we are not returning the attached certificates but rather asking this class to do something for us which is what the [tell don't ask principle](https://martinfowler.com/bliki/TellDontAsk.html#:~:text=Tell%2DDon't%2DAsk,an%20object%20what%20to%20do.) is all about. 

It is important because it allows us to change how we store data inside of this class without exposing the externals giving us the freedom to make any change we want inside of this class without affecting the external world.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -135,12 +139,30 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
                 updateThing(thing);
             }
         } catch (CloudServiceInteractionException e) {
-            List<String> certIds = thing.getAttachedCertificateIds();
-            if (certIds.contains(certificate.getCertificateId())) {
+            Map<String, Instant> certIds = thing.getAttachedCertificateIds();
+            if (certIds.containsKey(certificate.getCertificateId())) {
```
Nelsonochoam, line 143: Following to my comment above this should be something you can live inside of a `Thing`

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
```diff
@@ -135,12 +139,30 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
                 return true;
             }
             throw e;
         }
         return false;
     }
+
+    private Thing dtoToThing(ThingV1 dto) {
```
Nelsonochoam, line 151: Now I get it... If ThingV1 is a dto could we make  the name `ThingV1DTO` as a name?

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
```diff
@@ -50,65 +62,71 @@ void GIVEN_thingNameWithInvalidCharacters_WHEN_Thing_THEN_exceptionThrown() {
...
     void GIVEN_thing_WHEN_attachCertificate_THEN_certAttachedAndDirtyBitSet() {
-        Thing thing = Thing.of(0, "Thing");
-        thing.attachCertificate("cert-id");
+        Thing thing = Thing.of(0, mockThingName);
+        thing.attachCertificate(mockCertId);
 
         assertThat(thing.isModified(), is(true));
-        List<String> certIds = thing.getAttachedCertificateIds();
-        assertThat(certIds, equalTo(Collections.singletonList("cert-id")));
+        Map<String, Instant> certIds = thing.getAttachedCertificateIds();
```
Nelsonochoam, line 75: This should not be exposed, add methods inside the thing class that can ask questions about the certificates it has.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
```diff
@@ -50,65 +62,71 @@ void GIVEN_thingNameWithInvalidCharacters_WHEN_Thing_THEN_exceptionThrown() {
...
-        assertThat(version0_Thing_NoList, equalTo(version0_Thing_EmptyList));
-        assertThat(version0_Thing_NoList, not(equalTo(version0_Thing2_NoList)));
+        Map<String, Instant> certA = new HashMap<>(mockCertIdMap);
+        Map<String, Instant> certB = new HashMap<>(mockCertIdMap);
+        Instant instant = Instant.now();
+        certA.put("Cert-A", instant);
+        certB.put("Cert-B", instant); // same time instant for both certs
+
+        Thing version0_Thing_NoMap = Thing.of(0, mockThingName);
+        Thing version0_Thing2_NoMao = Thing.of(0, "Thing2", Collections.emptyMap());
```
Nelsonochoam, line 118:
```suggestion
        Thing version0_Thing2_NoMap = Thing.of(0, "Thing2", Collections.emptyMap());
```

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -47,6 +48,7 @@ public final class RuntimeConfiguration {
     static final String THINGS_KEY = "clientDeviceThings";
     static final String THINGS_V1_KEY = "v1";
     static final String THINGS_CERTIFICATES_KEY = "c";
+    static final String THINGS_VERSION_KEY = "v";
```
MikeDombo, line 51: what is the purpose of this?

We're already versioning by using the parent key.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -85,37 +85,43 @@ public Thing clone() {
...
+        Map<String, Instant> certIds;
+        synchronized (attachedCertificateIds) {
+            certIds = new HashMap<>(attachedCertificateIds);
```
MikeDombo, line 116: why are we copying?

src/test/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistryTest.java
```diff
@@ -88,7 +95,7 @@ void GIVEN_invalidCertificate_WHEN_createCertificate_THEN_exceptionThrown() {
     void GIVEN_certificateWithUpdate_WHEN_updateCertificate_THEN_updatedCertificateIsRetrievable()
             throws InvalidCertificateException {
         Certificate newCert = registry.createCertificate(validClientCertificatePem);
-        Instant now = Instant.now();
+        Instant now = Instant.now();System.out.println(now);
```
MikeDombo, line 98: remove

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -104,16 +106,19 @@ public Optional<ThingV1> getThingV1(String thingName) {
             return Optional.empty();
         }
 
+        // defaults to 0 for invalid config values
```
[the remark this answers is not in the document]
vaibhavmurkute: Coerce to int defaults it to 0 : https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/util/Coerce.java#L69-L87

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -104,16 +106,19 @@ public Optional<ThingV1> getThingV1(String thingName) {
...
+        int thingVersion = Coerce.toInt(v1ThingTopics.find(THINGS_VERSION_KEY));
+
         Topics certTopics = v1ThingTopics.findTopics(THINGS_CERTIFICATES_KEY);
         if (certTopics == null || certTopics.isEmpty()) {
```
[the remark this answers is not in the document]
vaibhavmurkute: I think `topics.isEmpty()` would be appropriate here as it would check if there are no children under this topic.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
vaibhavmurkute: Agreed. How about the `Thing` internally tracks the Instant for each certId ( with `Map<String, Instant>`) and expose only the list of certIds to the external world?

vaibhavmurkute: lol, or the other way around: Thing only tracks list of attached certIds and something else (`ThingRegistry`) tracks certId <-> Instant mapping in `ThingV1` dto

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
vaibhavmurkute: This copy ensures that a clone of `Thing` has its own copy of certIds that will not refer to the other `Thing`'s. ThingRegistry maintains the latest version of locally persisted Thing.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
MikeDombo: What is the purpose of that version?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
MikeDombo: Why do we have clones of Things?

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
jbutler: The idea is that this would be used for optimistic locking on Thing updates, though the original idea is that we wouldn't need to persist it. But, now I'm actually not seeing how it could be avoided.

We may want to rethink this one. I'd like to keep this memory footprint as small as we can.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -85,37 +85,43 @@ public Thing clone() {
      * @param certificateId Certificate ID to attach
      */
     public void attachCertificate(String certificateId) {
```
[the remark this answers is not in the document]
jbutler: It's a hex encoded string of a SHA256 hash. We could definitely do some sort of validation but TBH I'm not sure it's worth it.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -85,37 +85,43 @@ public Thing clone() {
...
-        if (attachedCertificateIds.addIfAbsent(certificateId)) {
-            modified = true;
+        if (attachedCertificateIds.containsKey(certificateId)) {
```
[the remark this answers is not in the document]
jbutler: Definitely. Need a way to update metadata. Either via this method or something else.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
jbutler: I think we're missing a layer here. These are low level primitives within the iot domain, but external callers don't care about these details.

We want callers to be able to say "authenticate this connection" and let us figure out what that means in terms of Things and Certificates being attached.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
jbutler: +1, we put it in a dto package, but you don't see that when reading the code. Good suggestion

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
vaibhavmurkute: @jbutler can correct me here, I believe it is so that ThingRegistry can vend Thing copies to anyone else but Registry is the sole entity that can manipulate (attach/detach certificate) or keep track of latest persisted `Thing` version.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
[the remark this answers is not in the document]
Nelsonochoam: Is there a way this could be explicit here? - I know there is `.dflt` method on the topics that we might want to use.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
Nelsonochoam: Regardless of the layer I think we should not expose the internals. Callers can still do what they want and interact with the thing through its API.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
Nelsonochoam: We wouldn't need to copy it if we followed why I was recommending above and keep the data structure encapsulated.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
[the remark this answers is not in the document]
vaibhavmurkute: Changed to update the Instant when attached same certificate.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
vaibhavmurkute: Thing now exposes this method: `Thing.isCertificateAttached(certId)`, cert registry can use that.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistry.java
vaibhavmurkute: Renamed DTOs

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
vaibhavmurkute: We'll get rid of this Thing version.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/RuntimeConfiguration.java
```diff
@@ -97,34 +99,38 @@ public void updateCAPassphrase(String passphrase) {
...
-            return Optional.of(new ThingV1(thingName, Collections.emptyMap()));
+            return Optional.of(new ThingV1DTO(thingVersion, thingName, Collections.emptyMap()));
         }
         Map<String, Long> certMap = new HashMap<>();
         certTopics.forEach(node -> {
             certMap.put(node.getName(), Coerce.toLong(node));
         });
 
-        return Optional.of(new ThingV1(thingName, certMap));
+        return Optional.of(new ThingV1DTO(thingVersion, thingName, certMap));
```
Nelsonochoam, line 121: Why do we still want to pass the thing version here if we are saying it is version1? We are not passing the version for the `CertificateV1DTO`

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
jbutler: Can refactor this later. Right now we have callers depending on this API. Removing it breaks a bunch of stuff

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
jbutler: Oh.... we're copying this because of clone?? That's not right. The constructor should copy. Let me remove this

MikeDombo (approved):

jbutler (approved):

PR #161, opened 2022-10-13
feat: verify registry data against configured trust duration

vaibhavmurkute:
**Description of changes:** 
- add a configurable clientDeviceTrustDuration that indicates how long a client device can be trusted before requiring it to be reverified.
- Locally stored ACTIVE Certificates and Thing <-> certificate association is verified to be within configured trust duration.

**Why is this change necessary:** To add a TTL to the offline auth cache entries

**How was this change tested:** Unit tests (`mvn package`, `mvn verify`)


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -198,6 +199,8 @@ protected void shutdown() throws InterruptedException {
     @Override
     public void postInject() {
         super.postInject();
+        context.put(SessionConfig.class, SessionConfig.from(getConfig()));
```
vaibhavmurkute, line 202: this _needs_ to be in postInject() to be able to inject the same SessionConfig (with the subscription to config changes) in all the CDA service dependencies.

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionConfig.java
```diff
@@ -75,10 +100,33 @@ private int getConfiguredSessionCapacity() {
...
+                MIN_SESSION_CAPACITY, MAX_SESSION_CAPACITY);
+    }
+
+    /**
+     * Retrieves configured client device trust duration.
+     * Invalid values are clamped to the valid range
+     *
+     * @return session capacity value
+     */
+    private int getConfiguredTrustDuration() {
```
Nelsonochoam, line 113: Similar comment to the one above we should make this configuration immutable.

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionConfig.java
```diff
@@ -75,10 +100,33 @@ private int getConfiguredSessionCapacity() {
...
+        }
+
+        int configValue = Coerce.toInt(configuration.findOrDefault(DEFAULT_CLIENT_DEVICE_TRUST_DURATION_HOURS,
+                SECURITY_TOPIC, CLIENT_DEVICE_TRUST_DURATION_HOURS_TOPIC));
+
+        return getClampedConfigValue(CLIENT_DEVICE_TRUST_DURATION_HOURS_TOPIC, configValue,
+                MIN_CLIENT_DEVICE_TRUST_DURATION_HOURS, MAX_CLIENT_DEVICE_TRUST_DURATION_HOURS);
+    }
+    
+    private int getClampedConfigValue(String configKey, int configValue, int min, int max) {
```
Nelsonochoam, line 125: Should we do this things for them or fail? - Specially for the TTL I think it might be better to break than just showing a log and tell them all is good.

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionManager.java
```diff
@@ -40,7 +41,17 @@ protected boolean removeEldestEntry(Map.Entry<String, Session> eldest) {
...
-    private SessionConfig sessionConfig;
+    private final SessionConfig sessionConfig;
+
+    /**
+     * Construct Session Manager.
+     *
+     * @param sessionConfig Session configuration
+     */
+    @Inject
+    public SessionManager(SessionConfig sessionConfig) {
```
Nelsonochoam, line 52: I liked how we where setting it before rather than injecting it. The problem with injecting it is that now we can't make this immutable given injected instances are singletons.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistryTest.java
```diff
@@ -3,16 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package com.aws.greengrass.clientdevices.auth.iot.registry;
+package com.aws.greengrass.clientdevices.auth.iot;
 
 import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
 import com.aws.greengrass.clientdevices.auth.configuration.RuntimeConfiguration;
 import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
-import com.aws.greengrass.clientdevices.auth.iot.Certificate;
```
Nelsonochoam, line 11: Why did this change?

src/test/java/com/aws/greengrass/clientdevices/auth/session/MqttSessionFactoryTest.java
```diff
@@ -107,4 +119,31 @@ void GIVEN_componentWithValidClientId_WHEN_createSession_THEN_returnsSession() t
         assertThat(session, is(IsNull.notNullValue()));
         assertThat(session.getSessionAttribute(Component.NAMESPACE, "component"), notNullValue());
     }
+
+    @Test
+    void GIVEN_credentialsWithCertificateBeyondTrustDuration_WHEN_createSession_THEN_throwsAuthenticationException()
+            throws InvalidCertificateException {
+        Certificate activeCert = CertificateFake.activeCertificate("fake-cert", Instant.EPOCH);
+        when(mockCertificateRegistry.getCertificateFromPem(any()))
```
Nelsonochoam, line 127: I am not a big fan of mocking when we could avoid it (generally I only prefer mocks for things that our outside of out control like an external lib).  Do we really need to be mocking the Certificate, creating a fake? We already have tooling for testing that creates certificates for us to test with.

This is (optional) so don't pay too much attention to it but just a thought

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistryTest.java
vaibhavmurkute: Registries were moved from `iot.registry` package to `iot` where all the IoT domain classes reside. We got rid of separate `registry` package.

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionConfig.java
```diff
@@ -34,16 +41,21 @@ public class SessionConfig {
      *
      * @param configuration Configuration topic for this service
      */
-    public SessionConfig(Topics configuration) {
+    private SessionConfig(Topics configuration) {
         this.configuration = configuration;
         this.sessionCapacity.set(getConfiguredSessionCapacity());
 
         this.configuration.subscribe((whatHappened, node) -> {
```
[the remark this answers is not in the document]
vaibhavmurkute: refactored now to contain new configurations within existing `CDAConfiguration` infra.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
```diff
@@ -109,6 +118,12 @@ public void deleteCertificate(Certificate certificate) {
         runtimeConfiguration.removeCertificateV1(certificate.getCertificateId());
     }
 
+    private boolean isTrusted(Certificate certificate) {
```
Nelsonochoam, line 121: Ideally we don't have registries trigger use cases, they should just deal with querying data from the store and returning domain objects and saving data. (See previous comment). When we do this we are mixing business logic inside of infra

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
```diff
@@ -55,7 +64,7 @@ public CertificateRegistry(RuntimeConfiguration runtimeConfiguration, ClientCert
     public Optional<Certificate> getCertificateFromPem(String certificatePem) throws InvalidCertificateException {
         Certificate cert = Certificate.fromPem(certificatePem);
         Optional<CertificateV1DTO> dto = runtimeConfiguration.getCertificateV1(cert.getCertificateId());
-        return dto.map(this::certificateV1DTOToCert);
+        return dto.map(this::certificateV1DTOToCert).filter(this::isTrusted);
```
Nelsonochoam, line 67: This is just filtering out what we don't trust anymore but not really changing the value we have stored. I don't think the registry is the place that should decide wether something is trusted or not. There are a few options I can think of.

1) We have a some sort of certificate factory that knows about this setting and when creating a certificate it assigns the TLL to the certificate it is returning. Then the certificate itself would have a method called `certificate.isValid` (this one is preferred)

2) We have something that tracks periodically checks for devices and their TTL and marks them as expired when they are no longer active.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistry.java
```diff
@@ -123,9 +127,13 @@ private Thing storeThing(Thing thing) {
      * @throws CloudServiceInteractionException when thing <-> certificate association cannot be verified
      */
     public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate) {
-        // If we have a local cache hit, then return true. Else, go to cloud
+        // If we have a local cache hit and it is trusted, then return true. Else, go to cloud
         if (thing.isCertificateAttached(certificate.getCertificateId())) {
-            return true;
+            Instant lastVerified = thing.getAttachedCertificateIds().get(certificate.getCertificateId());
```
Nelsonochoam, line 132: This should not go here. Same comment as the one above.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CDAConfiguration.java
```diff
@@ -94,6 +103,8 @@ public static CDAConfiguration from(Topics topics) throws URISyntaxException {
     private void triggerChanges(CDAConfiguration current, CDAConfiguration prev) {
         if (hasCAConfigurationChanged(prev)) {
             domainEvents.emit(new CAConfigurationChanged(current));
+        } else if (hasSecurityConfigurationChanged(prev)) {
```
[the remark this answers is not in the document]
vaibhavmurkute: Good catch

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/events/SecurityConfigurationChanged.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.aws.greengrass.clientdevices.auth.api.DomainEvent;
+import com.aws.greengrass.clientdevices.auth.configuration.CDAConfiguration;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+public class SecurityConfigurationChanged implements DomainEvent {
+    @Getter
+    private CDAConfiguration configuration;
```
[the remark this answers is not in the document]
Nelsonochoam: I fixed that for the CAConfiguration on this PR https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/164

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -147,4 +154,18 @@ public String getNamespace() {
...
         return Collections.singletonMap("ThingName", new WildcardSuffixAttribute(thingName));
     }
+
+    /**
+     * Updates the duration for which a certificate attachment can be trusted.
+     *
+     * @param newTrustDuration desired trust duration in hours
+     */
+    public static void updateMetadataTrustDurationHours(int newTrustDuration) {
+        metadataTrustDurationHours.set(newTrustDuration);
```
Nelsonochoam, line 164: Should we synchronize on this variable basically here and on the `isCertAttachementTrusted` method

src/test/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfigurationTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    public void GIVEN_noConfiguredTrustDuration_WHEN_getClientDeviceTrustDurationHours_THEN_returnDefaultTrustDuration() {
+        assertThat(securityConfig.getClientDeviceTrustDurationHours(), 
+                is(equalTo(DEFAULT_CLIENT_DEVICE_TRUST_DURATION_HOURS)));
+    }
+
+    @Test
+    public void GIVEN_invalid_configured_session_capacity_WHEN_getSessionCapacity_THEN_returns_clamped_capacity() {
```
Nelsonochoam, line 53: This tests are great for the SecurityConfig but we are missing some tests that ensure the new functionality works and the certificates are getting excluded.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
vaibhavmurkute: These updates to `AtomicInteger` are guaranteed to be thread-safe, without using explicit synchronization.

src/test/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfigurationTest.java
vaibhavmurkute: agreed, let me add a test on domain objects.

vaibhavmurkute: added unit tests

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyMetadataTrust.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Updates the duration for which a client device metadata can be trusted.
+     *
+     * @param newTrustDuration desired trust duration in hours
+     */
+    public void updateClientDeviceTrustDurationHours(int newTrustDuration) {
+        clientDeviceTrustDurationHours.set(newTrustDuration);
+    }
+
+    private boolean isWithinTrustDuration(Instant lastVerifiedInstant) {
```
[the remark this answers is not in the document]
vaibhavmurkute: moved trust factor inside the domain class itself

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
vaibhavmurkute: went with #1, thanks

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CDAConfiguration.java
```diff
@@ -22,6 +23,8 @@
  * <p>
  * Certificate Manager Service uses the following topic structure:
  * |---- configuration
+ * |    |---- security:
+ * |         |---- clientDeviceTrustDurationHours: "..."
```
MikeDombo, line 27: is hours the right unit we should use?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -116,6 +122,20 @@ public Map<String, DeviceAttribute> getDeviceAttributes() {
...
+     *
+     * @param newTrustDuration desired trust duration in hours
+     */
+    public static void updateMetadataTrustDurationHours(int newTrustDuration) {
+        metadataTrustDurationHours.set(newTrustDuration);
+    }
+
+    private boolean isStatusTrusted() {
+        Instant validTill = statusLastUpdated.plus(metadataTrustDurationHours.get(), ChronoUnit.HOURS);
+        return validTill.isAfter(Instant.now());
```
MikeDombo, line 136: if we can inject a `Clock` that'd be great for testing. `clock.instant()`

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -147,4 +154,18 @@ public String getNamespace() {
...
+    }
+
+    private boolean isCertAttachmentTrusted(Instant lastVerified) {
+        Instant validTill = lastVerified.plus(metadataTrustDurationHours.get(), ChronoUnit.HOURS);
+        return validTill.isAfter(Instant.now());
```
MikeDombo, line 169: likewise, recommend injecting a clock

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
Nelsonochoam: That would be awesome. We can also mock the clock though and make `Instant.now` return any value we want

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
```diff
@@ -126,4 +134,20 @@ void testEquals() {
         assertThat(Thing_CertA, not(equalTo(Thing_CertB)));
         assertThat(Thing_MultiCert, equalTo(Thing_MultiCert_copy));
     }
+
```
Nelsonochoam, line 137: We are missing a few tests here. I am ok with doing them later but testing that you attachSomething and then set a ttl and then wait for the time to pass and ask for it again and it should no longer be active. See how we are doing it here https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java#L120

Nelsonochoam (approved):
Approving this: we will handle making it default to 0 on the next PR

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CDAConfiguration.java
vaibhavmurkute: I was thinking _seconds_ or _minutes_ would be too low level of granularity that the customers won't need. I updated the PR to `clientDeviceTrustDurationMinutes` now.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
vaibhavmurkute: agreed, let me have that in a separate PR.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
vaibhavmurkute: will have these tests in a separate PR.

vaibhavmurkute:
Default `clientDeviceTrustDurationMinutes`  is currently set to a non-zero value. This default should be changed to zero in a separate PR. Need to decide on whether to treat it as a special case to disable offline auth.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/events/SecurityConfigurationChanged.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.certificate.events;
```
jbutler, line 6: nit: seems like more of an iot domain thing to me?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -79,10 +85,10 @@ public void setStatus(Status status, Instant lastUpdated) {
 
     /**
      * Check certificate status.
-     * @return true if this certificate is active in IoT Core.
+     * @return true if this certificate is verified and active in IoT Core.
      */
     public boolean isActive() {
-        return status == Status.ACTIVE;
+        return status == Status.ACTIVE && isStatusTrusted();
```
jbutler, line 91: Based on the method name, I would assume this is essentially just shorthand for `cert.getStatus() == Status.ACTIVE`. That's not the case though. We should consider overriding `getStatus` as well

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistryTest.java
```diff
@@ -57,6 +52,7 @@ void beforeEach() throws InvalidCertificateException {
         configTopics = Topics.of(new Context(), "config", null);
         registry = new ThingRegistry(mockIotAuthClient, domainEvents, RuntimeConfiguration.from(configTopics));
         mockCertificate = CertificateFake.of("mock-certificateId");
+        mockThing.detachCertificate(mockCertificate.getCertificateId());
```
jbutler, line 55: Red flag right here. We're obviously missing tests here, but this is a biiiig no no. There is almost never a good reason to mutate object state between tests like this. In this case, `mockThing` shouldn't be a static final like it is.

You're just asking for trouble if your setup fixtures are dependent on some previous object state. At best, this can make it hard to debug test failures. At worst, it can mask legitimate failures.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
```diff
@@ -38,6 +41,11 @@ void GIVEN_validThingName_WHEN_Thing_THEN_objectIsCreated() {
         Assertions.assertDoesNotThrow(() -> Thing.of("0123456789"));
     }
 
+    @AfterEach
```
jbutler, line 44: Similar thing here. Configure this up front so you have a known state of the world prior to your test running. In this case I think it's probably okay, but it's a dangerous pattern to follow.

Also, nit, move this method before the first test case?

src/test/java/com/aws/greengrass/clientdevices/auth/iot/CertificateTest.java
```diff
@@ -0,0 +1,68 @@
...
+        X509Certificate rootCA = CertificateTestHelpers.createRootCertificateAuthority("root", rootKeyPair);
+
+        validClientCertificate =
+                CertificateTestHelpers.createClientCertificate(rootCA, "Client", clientKeyPair.getPublic(),
+                        rootKeyPair.getPrivate());
+        validClientCertificatePem = CertificateHelper.toPem(validClientCertificate);
+    }
+
+    @AfterEach
+    void afterEach() {
```
jbutler, line 49: Same comment. BeforeEach

src/test/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfigurationTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    public void GIVEN_noConfiguredTrustDuration_WHEN_getClientDeviceTrustDurationMinutes_THEN_returnDefaultTrustDuration() {
+        assertThat(securityConfig.getClientDeviceTrustDurationMinutes(),
+                is(equalTo(DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES)));
+    }
+
+    @Test
+    public void GIVEN_invalid_configured_session_capacity_WHEN_getSessionCapacity_THEN_returns_clamped_capacity() {
```
jbutler, line 53: nit: let's try to stay consistent with naming. I've been doing camelCase not snake_case

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+ * |    |---- security:
+ * |          |---- clientDeviceTrustDurationMinutes: "..."
+ * </p>
+ */
+@Getter
+public final class SecurityConfiguration {
+    private static final Logger logger = LogManager.getLogger(SecurityConfiguration.class);
+    public static final String SECURITY_TOPIC = "security";
+    public static final String CLIENT_DEVICE_TRUST_DURATION_MINUTES_TOPIC = "clientDeviceTrustDurationMinutes";
+    public static final int DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 24;
```
jbutler, line 29: 24?

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final int MIN_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 0;
+    public static final int MAX_CLIENT_DEVICE_TRUST_DURATION_MINUTES = Integer.MAX_VALUE;
```
jbutler, line 31: Can be removed. You're reading the value as an int. By definition, there no value larger than `Integer.MAX_VALUE` so clamping will never do anything. I'm guessing the compiler will probably remove this.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/events/SecurityConfigurationChanged.java
Nelsonochoam: All this actually should go on the configuration domain we have. I think the iot and certificate "domain" would become more of of the authentication domain (maybe not?)

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
Nelsonochoam: Yeah I thought it would be 60 minutes

jbutler: We wanted this to be opt in, so 0 by default

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
vaibhavmurkute: agreed, removing this max-duration and also the clamping logic. The overflown integer values are now set to minimum trust duration value, zero, with an error log.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
Nelsonochoam: That is correct. @vaibhavmurkute mentioned he would like to handle it on a separate PR given setting it to 0 basically rejects everything, event when requesting things from the cloud at the moment... There are some minor tweaks that need to happen to make this a reality

vaibhavmurkute: yeah, will update this to an _opt-in_ kinda default (0) in a separate PR.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingTest.java
vaibhavmurkute: addressed

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistryTest.java
vaibhavmurkute: agreed, removed this. Thanks

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
vaibhavmurkute: should we have a method something like `getTrustedStatus()` ?

src/test/java/com/aws/greengrass/clientdevices/auth/iot/CertificateTest.java
vaibhavmurkute: addressed

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
vaibhavmurkute: overidden `getStatus()` to return the trusted status.

Nelsonochoam (approved):

Nelsonochoam (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -36,10 +37,12 @@ public class Certificate implements AttributeProvider {
 
     public enum Status {
         ACTIVE,
-        UNKNOWN
+        UNKNOWN,
+        TRUST_EXPIRED
```
jbutler, line 41: We already have `UNKNOWN` - just use that?

Nelsonochoam: I have recommended @vaibhavmurkute to do this. @jbutler the reasoning is that UNKNOW does not really represent what state the certificate is in. We actually know what happened which represents a valid state in the domain

Nelsonochoam (approved):

jbutler (approved):

PR #162, opened 2022-10-13
feat: refresh certificates in the background

Nelsonochoam:
**Description of changes:**
This PR introduces the `BackgrounCertificateRefresh` infra class whose sole purpose is to re verify stored certificates on a cadence, so when a client goes offline the stored value is one that fairly up to date. We are using a polling mechanism but it might change in the future to a push mechanism connecting to IoT Core and listening for certificate revocations.

We want to give customers a way to ensure their cached values are fairly recent in case the go offline, their devices can still communicate.

Integration test.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.certificate.infra;
+
+import com.aws.greengrass.clientdevices.auth.api.UseCases;
+import com.aws.greengrass.clientdevices.auth.infra.NetworkState;
+import com.aws.greengrass.clientdevices.auth.iot.CertificateRegistry;
+import com.aws.greengrass.clientdevices.auth.iot.usecases.VerifyIotCertificate;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import javax.inject.Inject;
+
+/**
+ * Periodically calls the VerifyIotCertificate to update the
+ * locally cached certificates on a cadence.
+ */
+public class BackgroundCertificateRefresh implements Runnable {
+    private final CertificateRegistry registry;
+    private final UseCases useCases;
+    private final NetworkState networkState;
+    private static final int DEFAULT_INTERVAL_SECONDS = 60 * 60; // 1H
+    private static final Logger logger = LogManager.getLogger(BackgroundCertificateRefresh.class);
+
+    private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    private final AtomicReference<ScheduledFuture<?>> scheduledFuture = new AtomicReference<>(null);
+
+
+    /**
+     * Creates an instance of the BackgroundCertificateRefresh.
+     * @param certificateRegistry - A certificate registry
+     * @param networkState - A network state
+     * @param useCases - useCases service
+     */
+    @Inject
+    public BackgroundCertificateRefresh(CertificateRegistry certificateRegistry, NetworkState networkState,
+            UseCases useCases) {
+        this.registry = certificateRegistry;
+        this.networkState = networkState;
+        this.useCases = useCases;
+    }
+
+    /**
+     * Start running the task every DEFAULT_INTERVAL_SECONDS.
+     */
+    public void start() {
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Periodically calls the VerifyIotCertificate to update the
+ * locally cached certificates on a cadence.
+ */
+public class BackgroundCertificateRefresh implements Runnable {
+    private final CertificateRegistry registry;
+    private final UseCases useCases;
+    private final NetworkState networkState;
+    private static final int DEFAULT_INTERVAL_SECONDS = 60 * 60; // 1H
```
Nelsonochoam, line 30: This default might be changed. This will probably work with the configuration stuff @vaibhavmurkute is working on.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final Logger logger = LogManager.getLogger(BackgroundCertificateRefresh.class);
+
+    private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    private final AtomicReference<ScheduledFuture<?>> scheduledFuture = new AtomicReference<>(null);
```
Nelsonochoam, line 34: I am using an atomic ref here to avoid the issue of checkstyle complaining I can't set values to null... I could add suppressions though if that is the route we want to take.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
```diff
@@ -100,6 +103,19 @@ public void updateCertificate(Certificate certificate) {
         runtimeConfiguration.putCertificate(certificateToCertificateV1DTO(certificate));
     }
 
+    /**
+     * Returns the PEMs that are stored for all the certificate ids that are being tracked.
+     */
+    public List<String> getAllCertificatePems() {
```
jbutler, line 109: I have concerns about this approach. Let's avoid loading everything in to memory all at once

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
jbutler: That's not checkstyle complaining... it's just a compiler error, right? If you update this variable then it isn't final.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -0,0 +1,109 @@
...
+ * locally cached certificates on a cadence.
+ */
+public class BackgroundCertificateRefresh implements Runnable {
+    private final CertificateRegistry registry;
+    private final UseCases useCases;
+    private final NetworkState networkState;
+    private static final int DEFAULT_INTERVAL_SECONDS = 60 * 60; // 1H
+    private static final Logger logger = LogManager.getLogger(BackgroundCertificateRefresh.class);
+
+    private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
jbutler, line 33: Let's inject this. Nucleus already has an SES we can use.

Nelsonochoam: Ahh interesting didn't know that

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Nope if I don't make it final and set variables to null checkstyle complains

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
Nelsonochoam: mmm what do you suggest ? - The reason I am doing this is because we need to update all the pems that we have stored. Do you want me to read one by one and execute the update?

jbutler (approved):

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -0,0 +1,121 @@
...
+        return scheduledFuture != null;
+    }
+
+    /**
+     * Runs verifyIotCertificate useCase for all the registered client certificate PEMs.
+     */
+    @Override
+    public void run() {
+        if (isNetworkDown()) {
+            logger.debug("Network is down - not refreshing certificates");
```
vaibhavmurkute, line 103: should this be logged at `TRACE`?

vaibhavmurkute: considering it will be run on a cadence

vaibhavmurkute (approved):

PR #163, opened 2022-10-13
refactor: extract iot client from registry and decouple update logic

Nelsonochoam:
**Description of changes:**
This changes extract the iot client from the registry effectively decoupling them. Before this change we had a boolean method that would mutate the registry. Now we moved the logic into a use case that can be called separately I will further decouple the logic that updates the registry in the next PR but this is the first step.

Refreshing the attached certificates to things is really hard to do without decoupling things. In the next PR I will split things a bit more so I can individually call something that would refresh the thing <-> cert attachements.

Existing tests pass

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
@@ -10,6 +10,8 @@ public final class Result<T> {
     private final Status status;
     private final T value;
 
+    private Exception err;
+
     private enum Status {
         OK, WARNING, ERROR
     }
@@ -17,6 +19,12 @@ public final class Result<T> {
     private Result(Status status, T value) {
         this.value = value;
         this.status = status;
+    }
+
+    private Result(Status status, T value, Exception e) {
+        this.value = value;
+        this.status = status;
+        this.err = e;
     }
 
     public static <V> Result<V> ok(V value) {
@@ -27,8 +35,8 @@ public final class Result<T> {
         return ok(null);
     }
 
-    public static <E extends Exception> Result<E> error(E value) {
-        return new Result<>(Status.ERROR, value);
+    public static <T, E extends Exception> Result<T> error(E value) {
+        return new Result<>(Status.ERROR, null, value);
     }
 
     public static <E extends Exception> Result<E> warning(E value) {
@@ -42,6 +50,10 @@ public final class Result<T> {
 
     public T get() {
         return value;
+    }
+
+    public Exception getError() {
+        return this.err;
     }
 
     public boolean isError() {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
@@ -10,8 +10,6 @@ public final class Result<T> {
     private final Status status;
     private final T value;
 
-    private Exception err;
-
     private enum Status {
         OK, WARNING, ERROR
     }
@@ -19,12 +17,6 @@ public final class Result<T> {
     private Result(Status status, T value) {
         this.value = value;
         this.status = status;
-    }
-
-    private Result(Status status, T value, Exception e) {
-        this.value = value;
-        this.status = status;
-        this.err = e;
     }
 
     public static <V> Result<V> ok(V value) {
@@ -35,8 +27,8 @@ public final class Result<T> {
         return ok(null);
     }
 
-    public static <T, E extends Exception> Result<T> error(E value) {
-        return new Result<>(Status.ERROR, null, value);
+    public static <T extends Exception> Result<T> error(T value) {
+        return new Result<>(Status.ERROR, value);
     }
 
     public static <E extends Exception> Result<E> warning(E value) {
@@ -47,13 +39,8 @@ public final class Result<T> {
         return new Result<>(Status.WARNING, null);
     }
 
-
     public T get() {
         return value;
-    }
-
-    public Exception getError() {
-        return this.err;
     }
 
     public boolean isError() {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
@@ -16,6 +16,9 @@ public class UseCases {
         Result<R> apply(D var1) throws Exception;
     }
 
+    public UseCases() {
+    }
+
     public UseCases(Context context) {
         this.context = context;
     }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
@@ -12,7 +12,15 @@ public class UseCases {
     private Context context;
 
     public interface UseCase<R, D>  {
-        Result<R> apply(D var1);
+        @SuppressWarnings("PMD.SignatureDeclareThrowsException")
+        Result<R> apply(D var1) throws Exception;
+    }
+
+    public UseCases() {
+    }
+
+    public UseCases(Context context) {
+        this.context = context;
     }
 
     public void init(Context context) {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
@@ -16,9 +16,6 @@ public class UseCases {
         Result<R> apply(D var1) throws Exception;
     }
 
-    public UseCases() {
-    }
-
     public UseCases(Context context) {
         this.context = context;
     }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/dto/CreateSessionDTO.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/dto/CreateSessionDTO.java
@@ -10,7 +10,7 @@ import lombok.Value;
 
 @Value
 @AllArgsConstructor
-public class CreateSessionCommand {
+public class CreateSessionDTO {
     String thingName;
     String certificatePem;
 }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/infra/ThingRegistry.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/infra/ThingRegistry.java
@@ -3,11 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package com.aws.greengrass.clientdevices.auth.iot;
+package com.aws.greengrass.clientdevices.auth.iot.infra;
 
 import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
 import com.aws.greengrass.clientdevices.auth.configuration.RuntimeConfiguration;
-import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
+import com.aws.greengrass.clientdevices.auth.iot.Thing;
 import com.aws.greengrass.clientdevices.auth.iot.dto.ThingV1DTO;
 import com.aws.greengrass.clientdevices.auth.iot.events.ThingUpdated;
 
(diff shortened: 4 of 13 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/CreateIoTThingSession.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Verifies if a certificate is attached to a thing. When the device is online it will try to verify
+     * it from the cloud and update the local values in case the device goes offline. When offline, the assertion
+     * will be based on the locally stored values.
+     * @param dto - VerifyCertificateAttachedToThingDTO
+     */
+    @Override
+    public Result<SessionImpl> apply(CreateSessionCommand dto) throws AuthenticationException {
```
Nelsonochoam, line 77: Somewhat convoluted use case that creates a session. I will further decouple this so I can simply call the part that refreshes the certs from the cloud. This got extracted from the thingRegistry

src/main/java/com/aws/greengrass/clientdevices/auth/session/MqttSessionFactory.java
```diff
@@ -54,19 +46,17 @@ public Session createSession(Map<String, String> credentialMap) throws Authentic
...
-                throw new AuthenticationException("Certificate isn't active");
-            }
-            Thing thing = thingRegistry.getOrCreateThing(mqttCredential.clientId);
-            if (!thingRegistry.isThingAttachedToCertificate(thing, cert.get())) {
-                throw new AuthenticationException("unable to authenticate device");
-            }
-            return new SessionImpl(cert.get(), thing);
-        } catch (CloudServiceInteractionException | InvalidCertificateException e) {
-            throw new AuthenticationException("Failed to verify certificate with cloud", e);
+        // NOTE: We should remove  calling this useCase from here, but for now it serves its purpose. We will
```
Nelsonochoam, line 49: This is temporarily living here. In the future this will get moved away.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/registry/ThingRegistryTest.java
```diff
@@ -1,115 +0,0 @@
...
-    void GIVEN_unmodifiedThing_WHEN_updateThing_THEN_oldThingReturned() {
-        Thing createdThing = registry.createThing(mockThingName);
-        Thing returnedThing = registry.updateThing(createdThing);
-        assertThat(createdThing, equalTo(returnedThing));
-
-        // TODO: no update event
-    }
-
-    @Test
-    void GIVEN_validThingAndCertificate_WHEN_isThingAttachedToCertificate_THEN_pass() {
```
Nelsonochoam, line 93: This file just got moved to a different folder. It is the same as the one above. I remove this last 2 tests that where directly testing the `registry.isThingAttachedToCertificate` that doesn't exist anymore. I will cover all this with integration tests for the offline flow

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistryTest.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.hamcrest.Matchers.is;
+
+@ExtendWith({MockitoExtension.class, GGExtension.class})
+class ThingRegistryTest {
+    private static final String mockThingName = "mock-thing";
+    private Topics configTopics;
+    private DomainEvents domainEvents;
+    private ThingRegistry registry;
+
+    @BeforeEach
```
Nelsonochoam, line 34: This is not new I just moved the file and removed some tests that no longer where valid.

src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
```diff
@@ -19,6 +21,12 @@ private Result(Status status, T value) {
         this.status = status;
     }
 
+    private Result(Status status, T value, Exception e) {
```
Nelsonochoam, line 24: This is done so we can throw errors from useCases and the type signature remains the same... Currently needed to allow us to move code embedded in classes to use cases without having to change the old code. We just make the old code call a use case.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/CreateIoTThingSession.java
```diff
@@ -0,0 +1,80 @@
...
+        String certificatePem = dto.getCertificatePem();
+        String thingName = dto.getThingName();
+        Optional<Certificate> certificate;
+
+        try {
+            certificate = certificateRegistry.getCertificateFromPem(certificatePem);
+
+            if (!certificate.isPresent()) {
+                throw new AuthenticationException("Certificate isn't active");
+            }
```
vaibhavmurkute, line 62: Just a heads up: we also need to check `cert.isActive()` here, once this PR is in: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/161

Nelsonochoam: True great observation

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
```diff
@@ -12,7 +12,15 @@ public class UseCases {
     private Context context;
 
     public interface UseCase<R, D>  {
-        Result<R> apply(D var1);
+        @SuppressWarnings("PMD.SignatureDeclareThrowsException")
+        Result<R> apply(D var1) throws Exception;
+    }
+
+    public UseCases() {
+    }
```
vaibhavmurkute, line 20: do we need this one?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/dto/CreateSessionDTO.java
```diff
@@ -0,0 +1,16 @@
...
+package com.aws.greengrass.clientdevices.auth.iot.dto;
+
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@Value
+@AllArgsConstructor
+public class CreateSessionDTO {
+    String thingName;
+    String certificatePem;
```
vaibhavmurkute, line 15: Should we tie these credentials to a protocol (MQTT)? MQTT sessions are currently based on these fields: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/main/java/com/aws/greengrass/clientdevices/auth/session/MqttSessionFactory.java#L76-L88

Should we add them to this DTO?

Nelsonochoam: Well that is a detail the infra should hide a domain event doesn't need to know about mqtt

src/test/java/com/aws/greengrass/clientdevices/auth/iot/ThingRegistryTest.java
```diff
@@ -83,28 +67,4 @@ void GIVEN_unmodifiedThing_WHEN_updateThing_THEN_oldThingReturned() {
 
         // TODO: no update event
     }
-
-    @Test
```
Nelsonochoam, line 87: Removed this tests given they where testing a that bad api. I will write some integration tests to cover this

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/CreateIoTThingSession.java
```diff
@@ -0,0 +1,80 @@
...
+import com.aws.greengrass.clientdevices.auth.iot.dto.CreateSessionDTO;
+import com.aws.greengrass.clientdevices.auth.iot.dto.VerifyThingAttachedToCertificateDTO;
+import com.aws.greengrass.clientdevices.auth.iot.infra.ThingRegistry;
+import com.aws.greengrass.clientdevices.auth.session.SessionImpl;
+
+import java.util.Optional;
+import javax.inject.Inject;
+
+public class CreateIoTThingSession
+        implements UseCases.UseCase<SessionImpl, CreateSessionDTO> {
```
vaibhavmurkute, line 25: should this UseCase return `Session` interface, instead of `SessionImpl`?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/infra/ThingRegistry.java
```diff
@@ -3,11 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package com.aws.greengrass.clientdevices.auth.iot;
+package com.aws.greengrass.clientdevices.auth.iot.infra;
```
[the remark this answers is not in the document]
Nelsonochoam: We should

src/main/java/com/aws/greengrass/clientdevices/auth/api/UseCases.java
Nelsonochoam: Yes we need it, if we add another constructor like the one below and do no add this one then it can't be auto created by the `Context` and a whole bunch of tests fail

vaibhavmurkute (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
```diff
@@ -0,0 +1,85 @@
...
+        this.iotAuthClient = iotAuthClient;
+        this.thingRegistry = thingRegistry;
+        this.networkState = networkState;
+    }
+
+    private boolean verifyLocally(Thing thing, Certificate certificate) {
+        return thing.isCertificateAttached(certificate.getCertificateId());
+    }
+
+    private boolean verifyFromCloud(Thing thing, Certificate certificate) {
```
jbutler, line 48: I think we need some logging in here. But that can be done separately.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Verifies if a certificate is attached to a thing. When the device is online it will try to verify
+     * it from the cloud and update the local values in case the device goes offline. When offline, the assertion
+     * will be based on the locally stored values.
+     * @param dto - VerifyCertificateAttachedToThingDTO
+     */
+    @Override
+    public Result<Boolean> apply(VerifyThingAttachedToCertificateDTO dto) throws AuthenticationException {
```
jbutler, line 71: What is throwing?

src/main/java/com/aws/greengrass/clientdevices/auth/session/MqttSessionFactory.java
```diff
@@ -54,19 +46,17 @@ public Session createSession(Map<String, String> credentialMap) throws Authentic
...
+        //  refactor this later
+        CreateIoTThingSession useCase = useCases.get(CreateIoTThingSession.class);
+        CreateSessionDTO command = new CreateSessionDTO(mqttCredential.clientId, mqttCredential.certificatePem);
+        Result<Session> sessionResult = useCase.apply(command);
+
+        if (sessionResult.isOk()) {
+            return sessionResult.get();
         }
+
+        throw new AuthenticationException(sessionResult.getError());
```
jbutler, line 59: I know we've gone back and forth about this functional style, but masking exceptions like this violates our Greengrass coding guidelines.

src/main/java/com/aws/greengrass/clientdevices/auth/api/Result.java
Nelsonochoam: See here https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/163/files#diff-4cbc61e8fc9bd73ada6b4ddb43f7e84858fbc178aed0134d410876e0fd382362R53

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
Nelsonochoam: This is a copy past from what we had but I agree. More logs would be nice

vaibhavmurkute (dismissed):

vaibhavmurkute (approved):

PR #166, opened 2022-10-14
Refresh thing attachment

Nelsonochoam:
**Description of changes:**
This is a follow up to https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/163 so check that one first. This PR enables the refresh of certificate thing attachments as part of the refresh cadence.

Keep cache in sync so we can keep working when offline

**Note**
I am yet to test paginated results. I will do that on the next PR

Integration test


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -104,17 +129,73 @@ public void run() {
...
+            useCase.apply(dto);
+        } catch (AuthenticationException e) {
+            logger.warn("Failed to refresh certificate thing attachment {}", dto.getThing().getThingName(), e);
         }
     }
 
+    private void refreshCertificateValidity(String pem) {
+        VerifyIotCertificate useCase = useCases.get(VerifyIotCertificate.class);
+        useCase.apply(pem);
+    }
```
vaibhavmurkute, line 197: should we also remove PEM from the filesystem? I see `VerifyIoTCertificate` usecase only updates metadata in the registry.

vaibhavmurkute: ^ for the certificates found invalid while _refresh_

Nelsonochoam: We could do that check the ones that we have that are not on the list of associated to the core and delete them.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -7,27 +7,45 @@
...
 
 /**
  * Periodically calls the VerifyIotCertificate to update the
  * locally cached certificates on a cadence.
  */
 public class BackgroundCertificateRefresh implements Runnable {
-    private final CertificateRegistry registry;
     private final UseCases useCases;
     private final NetworkState networkState;
     private static final int DEFAULT_INTERVAL_SECONDS = 60 * 60; // 1H
```
jbutler, line 44: Default was once a day

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -122,12 +141,67 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+            }
+        }
+
+        // TODO: This should not live here ideally it should be returned by the clientFactory but we
+        //  are adding it here to avoid introducing new changes to the nucleus
+        private GreengrassV2Client getGGV2Client(DeviceConfiguration deviceConfiguration) {
+            ApacheHttpClient.Builder httpClient = ClientConfigurationUtils
+                    .getConfiguredClientBuilder(deviceConfiguration);
+            GreengrassV2ClientBuilder clientBuilder = GreengrassV2Client.builder()
+                    .credentialsProvider(AnonymousCredentialsProvider.create())
```
jbutler, line 182: @MikeDombo may have comments on if we're doing this correctly

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -104,15 +128,66 @@ public void run() {
...
+                    "Access denied to list things associated with core device. Please make sure you "
+                    + "have setup the correct permissions.");
+        } catch (Exception e) {
+            logger.warn("Failed to get things associated to the core device. Retry will be schedule later");
         }
+
+        return Stream.empty();
+    }
+
+    private Consumer<Thing> refresh() {
```
jbutler, line 161: Like we talked about in the huddle, let's maybe rework this a bit so we have the opportunity to remove things once they're no longer active.

If we throw attached certificates into a set, then we can refresh those first. Then verifying thing attachments can first query the cert registry to make sure the cert is still active.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -46,7 +46,11 @@ public enum Status {
     private Instant statusLastUpdated;
 
 
-    Certificate(String certificateId) {
+    /**
+     * Builds a new Certificate instance given its id generated from PEM file.
+     * @param certificateId - a certificateId
+     */
+    public Certificate(String certificateId) {
```
jbutler, line 53: I actually had this package private as a hint that nothing outside of the package should instantiate a Certificate directly. It looks like you just had to make this public for tests. Maybe we can move that test into the same package?

I probably should have commented my motivation for doing that to make it clear.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
```diff
@@ -101,13 +101,6 @@ public void updateCertificate(Certificate certificate) {
         runtimeConfiguration.putCertificate(certificateToCertificateV1DTO(certificate));
     }
 
-    /**
-     * Returns the PEMs that are stored for all the certificate ids that are being tracked.
-     */
-    public String[] getAllCertificateIds() {
```
jbutler, line 107: As much as I :heart: seeing red, I think we'll still want this for the removal flow.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
Nelsonochoam: It was not for the tests but for the BackgroundRefresh and avoid getting the PEM first but I can rework this

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Yup I'll take care of removal

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
Nelsonochoam: I am still removing it and basing it off the certificates attached to things. See https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/166/files#diff-6161e346c7f85b92277837560e988c3304ad8a2f83eb92b5525a2e92e7fc3dd3R214

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
Nelsonochoam: @MikeDombo could you check if this looks good? - If not what other way do we have to get a GreengrassV2Client?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -104,18 +129,144 @@ public void run() {
...
+    }
+
+    private void refreshCertificateThingAttachment(Thing thing) {
+        VerifyThingAttachedToCertificate useCase = useCases.get(VerifyThingAttachedToCertificate.class);
+
+        thing.getAttachedCertificateIds().keySet().forEach(certificateId -> {
+            Optional<String> certPem = pemStore.getPem(certificateId);
+
+            if (!certPem.isPresent()) {
+                return;
```
vaibhavmurkute, line 186: any thoughts on having a log here (may be TRACE)? It'd be useful to know why a certain certificate attachment was not refreshed. At this point if we don't have PEM for it, we may never have it until the client reconnects and until that time this cert won't be refreshed.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -104,18 +129,144 @@ public void run() {
...
             }
+
+            try {
+                Certificate certificate = Certificate.fromPem(certPem.get());
+                useCase.apply(new VerifyThingAttachedToCertificateDTO(thing, certificate));
+            } catch (InvalidCertificateException e) {
+                logger.atWarn().kv("thing", thing.getThingName()).kv("certificate", certificateId)
+                        .log("Failed to verify thing attached to certificate");
```
vaibhavmurkute, line 194: also add the cause (stacktrace) of the exception?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -104,18 +129,144 @@ public void run() {
...
+       associations.getCertificateAssociations().forEach(this::refreshCertificateValidity);
+       associations.getStaleCertificateAssociations().forEach(certificateRegistry::deleteCertificate);
+       associations.getStaleThingAssociations().forEach(thingRegistry::deleteThing);
+    }
+
+    private void refreshCertificateThingAttachment(Thing thing) {
+        VerifyThingAttachedToCertificate useCase = useCases.get(VerifyThingAttachedToCertificate.class);
+
+        thing.getAttachedCertificateIds().keySet().forEach(certificateId -> {
+            Optional<String> certPem = pemStore.getPem(certificateId);
```
vaibhavmurkute, line 183: We don't need PEM for verifying Thing <-> Certificate attachment. Just CertId and thing name would do. How about updating `VerifyThingAttachedToCertificateDTO` to have only thingName and CertId instead of complete domain objects?

Nelsonochoam: Yeah it was just that we don't expose creating a certificate without the pem and if for some reason we don't have the pem stored and we can't update the cert validity would we still want to update its attachements?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
MikeDombo: What's the purpose of this change? You already have access to a client from the Nucleus itself so that you don't need to handle this on your own.

Yes, this all looks fine otherwise, especially if you've taken it from the Nucleus.

Nelsonochoam: The main reason to do this is that the Nucleus provides a GGV2DataClient which provides access to data plain apis but it does not provide a GGV2Client to access the control plain api that has the operation we want to run.

The reason we need to call that API is to know what is still attached to the core vs what has been detached to update the local registries

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -104,18 +130,142 @@ public void run() {
...
+        associations.getStaleThings().forEach(thingRegistry::deleteThing);
 
-        for (String certificateId : certificateIds) {
-            if (pemStore.exists(certificateId)) {
-                String pem = pemStore.getPem(certificateId).get();
-                verifyIotCertificate.apply(pem);
+        certificateRegistry.getAllCertificates()
+                .forEach(certificate -> {
+            if (associations.isAttachedToThing(certificate)) {
+                refreshCertificateValidity(certificate);
```
vaibhavmurkute, line 175: should we also deleteCertificate() if a certificate is found invalid during `refreshCertificateValidity()` ?

(rest of the changes look good to me)

Nelsonochoam: We do that already as part of the useCase

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
MikeDombo: `RegionUtils.getGreengrassControlPlaneEndpoint(awsRegion, this.envStage)` should make getting the endpoint simpler.

MikeDombo: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/dd8e059e802f44a763819a01e2088263ae56f3b3/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java#L160

src/main/java/com/aws/greengrass/clientdevices/auth/iot/CertificateRegistry.java
```diff
@@ -102,19 +103,22 @@ public void updateCertificate(Certificate certificate) {
...
+     * Deletes a certificate from the repository provided its id.
      *
-     * @param certificate certificate to remove
+     * @param certificateId - certificateId of the certificate to remove
      */
-    public void deleteCertificate(Certificate certificate) {
-        runtimeConfiguration.removeCertificateV1(certificate.getCertificateId());
+    public void deleteCertificate(String certificateId) {
+        runtimeConfiguration.removeCertificateV1(certificateId);
     }
```
vaibhavmurkute, line 122: make this function `private` ? so that exposed `public void deleteCertificate(Certificate certificate) {}` will always remove registry entry as well as PEM.

vaibhavmurkute (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -173,8 +185,10 @@ void GIVEN_clientDevice_WHEN_verifyingItsIdentity_THEN_pemStored() throws Interr
    @Test
    void GIVEN_storedCertificates_WHEN_refreshEnabled_THEN_storedCertificatesRefreshed() throws
            NoSuchAlgorithmException, CertificateException, OperatorCreationException, IOException,
-           InvalidCertificateException, InterruptedException {
+           InvalidCertificateException, InterruptedException, AuthenticationException {
```
MikeDombo, line 188: nit: since it is just a test, just throw all Exceptions.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/ClientCertificateStore.java
```diff
@@ -76,12 +82,28 @@ public void storePem(String certificateId, String certificatePem) throws Certifi
...
 
+    /**
+     * Removes the PEM for a certificateId alias.
+     * @param certificateId - a certificate id
+     */
+    public void removePem(String certificateId) {
+        try {
+            keyStore.deleteEntry(certificateId);
+        } catch (KeyStoreException e) {
+            logger.atError().cause(e).kv("certificateId", certificateId).log("Failed to remove certificate PEM");
```
MikeDombo, line 93: when would this happen? If it does happen, is there anything the customer can do about it?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -122,12 +145,58 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
             } catch (Exception e) {
                 logger.atError().cause(e).kv("thingName", thing.getThingName())
                         .kv("certificateId", certificate.getCertificateId())
-                        .log("Failed to verify certificate thing association. Check that the core device's IoT policy"
-                                + " grants the greengrass:VerifyClientDeviceIoTCertificateAssociation permission.");
+                        .log("Failed to verify certificate thing association. Check that the core device's "
+                                + "IoT policy grants the greengrass:VerifyClientDeviceIoTCertificateAssociation "
+                                + "permission.");
```
MikeDombo, line 150: I know this was there already, but remove the period at the end of the log. Our logger appends a period automatically.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -122,12 +145,58 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+                            .build();
+
+            try (GreengrassV2Client client = getGGV2Client(configuration)) {
+                ListClientDevicesAssociatedWithCoreDeviceResponse response =
+                        client.listClientDevicesAssociatedWithCoreDevice(request);
+
+                return response.associatedClientDevices().stream()
+                        .map(AssociatedClientDevice::thingName)
+                        .map(Thing::of);
+            }  catch (SdkClientException | AwsServiceException e) {
```
MikeDombo, line 174: what about other exceptions? They're just going to bubble up. Safer to just catch all exceptions here?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -122,12 +145,58 @@ public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate
...
+            String awsRegion = Coerce.toString(deviceConfiguration.getAWSRegion());
+            GreengrassV2ClientBuilder clientBuilder = GreengrassV2Client.builder()
+                    .httpClient(httpClient.build())
+                    .credentialsProvider(lazyCredentialProvider)
+                    .overrideConfiguration(
+                            ClientOverrideConfiguration.builder().retryPolicy(RetryMode.STANDARD).build());
+
+            if (!Utils.isEmpty(awsRegion)) {
+                String greengrassServiceEndpoint = RegionUtils.getGreengrassControlPlaneEndpoint(
+                        awsRegion,  IotSdkClientFactory.EnvironmentStage.PROD);
```
MikeDombo, line 194: don't hardcode the environment, this must pull the environment from the device config.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -7,48 +7,73 @@
...
- * locally cached certificates on a cadence.
+ * Periodically updates the certificates and its relationships to things
+ * (whether they are still attached or not to a thing) to keep them in sync with the cloud.
  */
 public class BackgroundCertificateRefresh implements Runnable {
-    private final CertificateRegistry registry;
     private final UseCases useCases;
     private final NetworkState networkState;
-    private static final int DEFAULT_INTERVAL_SECONDS = 60 * 60; // 1H
+    private static final int DEFAULT_INTERVAL_SECONDS = 60 * 60 * 24; // Once a day
```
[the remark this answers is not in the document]
MikeDombo: and if it fails after the 3 times, then the device is allowed to stay connected right?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/ClientCertificateStore.java
Nelsonochoam: Based on the docs it could happen if the entry can't be removed or if the keystore has not been initialized. This is not supper critical as this keystore lives in memory and it might happen you that we try to delete a certificate but the alias is not in there.

I guess we could make the case for throwing the exception and letting the client handle it

MikeDombo: OK, maybe go for WARN here if there's nothing the customer can do about it.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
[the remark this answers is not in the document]
Nelsonochoam: If it fails after 3 times and we go offline after that then the value will be based on what is on the cache and all those values have a TTL. Worst case scenario when this fails is we don't refresh the locally stored certificates and we keep what the cache has which also gets updated each time the device connects.

I agree we could make retries even smarter, than just retrying 3 times. I could handle it on the next PR when I handle paginated responses

vaibhavmurkute (approved):

MikeDombo (approved):

PR #168, opened 2022-10-18
refactor: remove result

Nelsonochoam:
**Description of changes:**
If became apparent after using `Result` for a bit that it is an unnecessary pattern for the style of error handling the team follows. `Result` is a pattern for functional style error handling which is basically a monad with the intention of keeping the domain free of side effects, but that comes with some undesired side effect that do not comply with the style the team prefers, using checked exceptions.

This PR basically removes the Result class and makes UseCases follow the pattern for reporting errors using checked exceptions.

Result is somewhat new and it is not yet in production. It is better to remove this now and avoid spreading it if it is something that is not adding any apparent benefit and does not align with the preferred style of error handling.


Nelsonochoam:
Tests are failing on this one, but this PR https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/169 fixes the issue with the flaking tests

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CACertificateChainChangedHandler.java
```diff
@@ -43,7 +44,10 @@ public void listen() {
      */
     @Override
     public void accept(CACertificateChainChanged event) {
-        // TODO: Move retry logic from the domain to here, based on the result returned from the useCase
-        useCases.get(RegisterCertificateAuthorityUseCase.class).apply(null);
+        try {
+            useCases.get(RegisterCertificateAuthorityUseCase.class).apply(null);
+        } catch (UseCaseException e) {
+            // TODO: Move retry logic from the domain to here, based on the result returned from the useCase
```
jbutler, line 50: We should have a backlog item to track this if we don't currently have a mechanism to retry failures. At the very least, do we need to log here?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CAConfigurationChangedHandler.java
```diff
@@ -47,12 +52,15 @@ public void listen() {
...
-        } else {
-            useCases.get(ConfigureManagedCertificateAuthority.class).apply(configuration);
+        try {
+            if (configuration.isUsingCustomCA()) {
+                useCases.get(ConfigureCustomCertificateAuthority.class).apply(configuration);
+            } else {
+                useCases.get(ConfigureManagedCertificateAuthority.class).apply(configuration);
+            }
+        } catch (UseCaseException e) {
+            logger.error("Failed to configure certificate authority");
```
jbutler, line 62: Stack trace will be useful to understand why this failed

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -46,7 +45,7 @@ public VerifyIotCertificate(IotAuthClient iotAuthClient, CertificateRegistry cer
     }
 
     @Override
-    public Result<Boolean> apply(String certificatePem) {
+    public Boolean apply(String certificatePem) {
```
jbutler, line 48: nit: can use `boolean` instead of `Boolean`, which is an object wrapping the native type

Nelsonochoam: I can't the Generic type doesn't allow me to use primitives there

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CACertificateChainChangedHandler.java
vaibhavmurkute, line 50: better to have a failure log here until we finish this TODO

Nelsonochoam: We did not have anything before, but I agree

jbutler (approved):

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/usecases/RegisterCertificateAuthorityUseCase.java
```diff
@@ -47,39 +48,38 @@ public RegisterCertificateAuthorityUseCase(
...
             // NOTE: uploadCoreDeviceCAs should not block execution.
-            certificateManager.uploadCoreDeviceCAs(thingName, certificateR.get());
-            return Result.ok();
+            certificateManager.uploadCoreDeviceCAs(thingName, certificate.get());
+            return  null;
         } catch (CloudServiceInteractionException e) {
             logger.atError().cause(e).kv("coreThingName", thingName)
                     .log("Unable to upload core CA certificates to the cloud");
-            return Result.warning(e);
+            throw  new UseCaseException(e);
```
MikeDombo, line 82: nit: format, double spaced

MikeDombo (approved):

PR #169, opened 2022-10-18
fix: race condition

Nelsonochoam:
**Description of changes:**
Fixes a race condition happening on some of the integration tests where the custom CA configuration change where not fully applied before we continued subscribed resulting on the test getting a certificate rotated by the old CA

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
@@ -176,7 +176,6 @@ public class CustomCaConfigurationTest {
                 .withValue(privateKeyUri.toString());
         topics.lookup(CONFIGURATION_CONFIG_KEY, CERTIFICATE_AUTHORITY_TOPIC, CA_CERTIFICATE_URI)
                 .withValue(certificateUri.toString());
-
         kernel.getContext().waitForPublishQueueToClear();
     }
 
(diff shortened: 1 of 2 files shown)


vaibhavmurkute (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -304,6 +312,8 @@ void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serve
 
         URI privateKeyUri = new URI("file:///private.key");
         URI certificateUri = new URI("file:///certificate.pem");
+        when(securityServiceMock.getKeyPair(privateKeyUri, certificateUri)).thenReturn(intermediateKeyPair);
```
jbutler, line 315: Ideally we wouldn't mock this, seeing how it's an integration test. But we can do that later

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -304,6 +312,8 @@ void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serve
...
+        doReturn(chain).when(certificateStoreSpy).loadCaCertificateChain(privateKeyUri, certificateUri);
 
         AtomicReference<CertificateUpdateEvent> eventRef = new AtomicReference<>();
         Pair<CompletableFuture<Void>, Consumer<CertificateUpdateEvent>> asyncCall =
```
jbutler, line 319: I realize this is a smaller refactor, but I'll just make a note of this.

This test function is really confusing IMO, and I'm really failing to see how it makes anything easier. It just seems to add layers of indirection. Also, I don't think it's correct to assert on how many times the consumer was "accepted"

I'd vote for removing it from our tests and make whatever this is doing under the hood explicit. Thoughts?

jbutler (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
Nelsonochoam: Yeah I agree. I am pro not mocking but when this was written we had a few things missing

PR #170, opened 2022-10-18
feat: add opt-in metadata trust verification

vaibhavmurkute:
**Description of changes:** 
- Adds opt-in trust verification: metadata trust verification is enabled when configured trust duration is non-zero.
- Trust verification is disabled by default.

**Why is this change necessary:** Enable/disable offline auth as necessary

**How was this change tested:**  Unit tests (`mvn package`, `mvn verify`)

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
@@ -26,8 +26,9 @@ public final class SecurityConfiguration {
     private static final Logger logger = LogManager.getLogger(SecurityConfiguration.class);
     public static final String SECURITY_TOPIC = "security";
     public static final String CLIENT_DEVICE_TRUST_DURATION_MINUTES_TOPIC = "clientDeviceTrustDurationMinutes";
-    // TODO: change default value to zero (opt-in offline auth)
-    public static final int DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 60;
+    // opt-in trust verification: metadata trust verification is enabled when configured trust duration is non-zero
+    // trust verification is disabled by default
+    public static final int DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 0;
     public static final int MIN_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 0;
 
     private int clientDeviceTrustDurationMinutes;
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
@@ -145,9 +145,20 @@ public class Certificate implements AttributeProvider {
         metadataTrustDurationMinutes.set(newTrustDuration);
     }
 
+    /**
+     * Following opt-in offline auth, verifies the certificate status trust.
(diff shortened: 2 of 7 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
```diff
@@ -145,9 +145,20 @@ public static void updateMetadataTrustDurationMinutes(int newTrustDuration) {
...
     }
 
+    /**
+     * Following opt-in offline auth, verifies the certificate status trust.
+     * Metadata trust verification is enabled when configured trust duration is non-zero.
+     * Trust verification is disabled by default.
+     *
+     * @return whether the status is verified within configured trust duration
+     */
     private boolean isStatusTrusted() {
```
Nelsonochoam, line 155: It would be nice to move this into an interface that both of those object implement kind of 

```java
interface Expirable {
      default boolean hasExpired(int maxMinutesToLive) {
        if (maxMinutesToLive > 0) {
            Instant validTill = statusLastUpdated.plus(maxMinutesToLive, ChronoUnit.MINUTES);
            return validTill.isAfter(Instant.now());
        }
        return true;
      }
}
```

and then have 

```java
    public Status getStatus() {
        if (hasExpired(metadataTrustDurationMinutes.get())) {
            return status;
        }
        return Status.UNKNOWN;
    }
```

```java
class Certificate implements Expirable
```

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfiguration.java
```diff
@@ -26,8 +26,9 @@ public final class SecurityConfiguration {
     private static final Logger logger = LogManager.getLogger(SecurityConfiguration.class);
     public static final String SECURITY_TOPIC = "security";
     public static final String CLIENT_DEVICE_TRUST_DURATION_MINUTES_TOPIC = "clientDeviceTrustDurationMinutes";
-    // TODO: change default value to zero (opt-in offline auth)
-    public static final int DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 60;
+    // opt-in trust verification: metadata trust verification is enabled when configured trust duration is non-zero
+    // trust verification is disabled by default
+    public static final int DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES = 0;
```
Nelsonochoam, line 31: Should we add one test that ensures that when not provided this defaults to 0?

Nelsonochoam: Also, checking this again, we are now allowing people to set the value to 0.1 which might cause undesired behaviour when offline

vaibhavmurkute: Test for fallback to default when not provided is here: [link](https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/test/java/com/aws/greengrass/clientdevices/auth/configuration/SecurityConfigurationTest.java#L45-L49)

We currently allow only integer values for trust duration config. Decimal numbers are floor-rounded to (cast to) Integers. I just added a unit test for that.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Certificate.java
vaibhavmurkute: Good call. Added a common `Trustable` interface. Thanks

PR #175, opened 2022-10-20
feat: refresh certificates when network comes back up

Nelsonochoam:
**Description of changes:**
This makes it so that the background refresh runs when the network connectivity is resumed

Improves when we refresh the certificate cache so we opportunistically update the values in it.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -33,6 +33,7 @@ import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -54,7 +55,7 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
 
     private ScheduledFuture<?> scheduledFuture = null;
     private final ScheduledThreadPoolExecutor scheduler;
-    private Instant nextRun;
+    private final AtomicReference<Instant> nextRun = new AtomicReference<>();
 
 
     /**
@@ -126,20 +127,18 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
      * Runs verifyIotCertificate useCase for all the registered client certificate PEMs.
      */
     @Override
-    public void run() {
+    public synchronized void run() {
         if (isNetworkDown()) {
             logger.debug("Network is down - not refreshing certificates");
             return;
         }
 
         if (!canRun()) {
-            logger.info("Certificates where refreshed recently - retrying later");
+            logger.info("Certificates were refreshed recently - retrying later");
             return;
         }
 
-        // Why the -10? This is just to avoid a scenario where the scheduler triggers it and the current time
-        // is not exactly after DEFAULT_INTERVAL_SECONDS because this variable gets set a few ms after run is called
-        nextRun = Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS - 10));
+        nextRun.set(Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS)));
         Optional<ThingAssociations> associations = getThingsAssociatedWithCoreDevice();
         associations.ifPresent(this::refresh);
     }
@@ -158,11 +157,12 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
     }
 
     private boolean canRun() {
-        if (nextRun == null) {
+        if (nextRun.get() == null) {
             return true;
         }
 
-        return Instant.now().isAfter(nextRun);
+        Instant now = Instant.now();
+        return now.compareTo(nextRun.get()) >= 0;
     }
 
     /**
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -24,6 +24,7 @@ import software.amazon.awssdk.services.greengrassv2data.model.InternalServerExce
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
 
 import java.time.Duration;
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -53,6 +54,7 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
 
     private ScheduledFuture<?> scheduledFuture = null;
     private final ScheduledThreadPoolExecutor scheduler;
+    private Instant nextRun;
 
 
     /**
@@ -95,7 +97,6 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
            return;
         }
 
-        networkState.registerHandler(this);
         logger.info("Starting background refresh of client certificates every {} seconds", intervalSeconds);
         scheduledFuture =
             scheduler.scheduleAtFixedRate(this, intervalSeconds, intervalSeconds, TimeUnit.SECONDS);
@@ -131,6 +132,14 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
             return;
         }
 
+        if (!canRun()) {
+            logger.info("Certificates where refreshed recently - retrying later");
+            return;
+        }
+
+        // Why the -10? This is just to avoid a scenario where the scheduler triggers it and the current time
+        // is not exactly after DEFAULT_INTERVAL_SECONDS because this variable gets set a few ms after run is called
+        nextRun = Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS - 10));
         Optional<ThingAssociations> associations = getThingsAssociatedWithCoreDevice();
         associations.ifPresent(this::refresh);
     }
@@ -146,6 +155,14 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
             logger.atInfo("Network back up - refreshing certificates");
             run();
         }
+    }
+
+    private boolean canRun() {
+        if (nextRun == null) {
+            return true;
+        }
+
+        return Instant.now().isAfter(nextRun);
     }
 
     /**
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -134,13 +134,14 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
         }
 
         if (!canRun()) {
-            logger.info("Certificates were refreshed recently - retrying later");
             return;
         }
 
-        nextRun.set(Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS)));
         Optional<ThingAssociations> associations = getThingsAssociatedWithCoreDevice();
-        associations.ifPresent(this::refresh);
+        associations.ifPresent(a -> {
+            this.refresh(a);
+            nextRun.set(Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS)));
+        });
     }
 
 
@@ -162,7 +163,7 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
         }
 
         Instant now = Instant.now();
-        return now.compareTo(nextRun.get()) >= 0;
+        return now.equals(nextRun.get()) || now.isAfter(nextRun.get());
     }
 
     /**
@@ -185,13 +186,17 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
             ThingAssociations associations = ThingAssociations.create(cloudThings);
             associations.setLocalThings(localThings);
             return Optional.of(associations);
-        } catch (AccessDeniedException e) {
-            logger.atInfo().cause(e).log(
-                "Did not refresh local certificates. To enable certificate refresh add a policy to the core device"
-                        + "that grants the greengrass:ListClientDevicesAssociatedWithCoreDevice permission");
         } catch (Exception e) {
-            logger.atWarn().cause(e).log(
-                "Failed to get things associated to the core device. Retry will be scheduled later");
+            // We can't explicitly catch this given we always throw a CloudServiceInteractionException, hence we do
+            // have to check it here
+            if (e.getCause() instanceof AccessDeniedException) {
+                logger.atInfo().log(
+                    "Did not refresh local certificates. To enable certificate refresh add a policy to the core device"
+                            + " that grants the greengrass:ListClientDevicesAssociatedWithCoreDevice permission");
+            } else {
+                logger.atWarn().cause(e).log(
+                        "Failed to get things associated to the core device. Retry will be scheduled later");
+            }
         }
 
         return Optional.empty();
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -228,6 +228,7 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
         });
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private void refreshCertificateThingAttachment(Thing thing) {
         VerifyThingAttachedToCertificate useCase = useCases.get(VerifyThingAttachedToCertificate.class);
 
@@ -245,13 +246,14 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
             try {
                 Certificate certificate = Certificate.fromPem(certPem.get());
                 useCase.apply(new VerifyThingAttachedToCertificateDTO(thing, certificate));
-            } catch (InvalidCertificateException e) {
+            } catch (InvalidCertificateException | RuntimeException e) {
                 logger.atWarn().cause(e).kv("thing", thing.getThingName()).kv("certificate", certificateId)
                         .log("Failed to verify thing attached to certificate");
             }
         });
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private void refreshCertificateValidity(Certificate certificate) {
         Optional<String> certPem = pemStore.getPem(certificate.getCertificateId());
 
@@ -262,7 +264,12 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
         }
 
         VerifyIotCertificate useCase = useCases.get(VerifyIotCertificate.class);
-        useCase.apply(certPem.get());
+        try {
+            useCase.apply(certPem.get());
+        } catch (RuntimeException e) {
+            logger.atWarn().cause(e).kv("certificateId", certificate.getCertificateId())
+                    .log("Failed to verify certificate validity");
+        }
     }
 
     private boolean isNetworkDown() {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -152,7 +152,6 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
     @Override
     public void accept(NetworkState.ConnectionState connectionState) {
         if (connectionState == NetworkState.ConnectionState.NETWORK_UP) {
-            logger.atInfo("Network back up - refreshing certificates");
             run();
         }
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -93,6 +95,7 @@ public void start(int intervalSeconds) {
            return;
         }
 
+        networkState.registerHandler(this);
```
jbutler, line 98: Register in infra init?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -132,6 +135,19 @@ public void run() {
         associations.ifPresent(this::refresh);
     }
 
+
+    /**
+     * Handler to react to network changes.
+     * @param connectionState - A network state
+     */
+    @Override
+    public void accept(NetworkState.ConnectionState connectionState) {
```
jbutler, line 144: Let's still cap this at once a day. Maybe track last sync time?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: It seemed kind of weird putting it there given already have access to the service here, but I can so that we keep registration in one place

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: I had that same thought but decided not to add it. I think that is fine. I'll change it

Nelsonochoam: Done

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -128,10 +132,39 @@ public void run() {
             return;
         }
 
+        if (!canRun()) {
+            logger.info("Certificates where refreshed recently - retrying later");
```
vaibhavmurkute, line 136: `where` --> `were`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -128,10 +132,39 @@ public void run() {
...
+            return;
+        }
+
+        // Why the -10? This is just to avoid a scenario where the scheduler triggers it and the current time
+        // is not exactly after DEFAULT_INTERVAL_SECONDS because this variable gets set a few ms after run is called
+        nextRun = Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS - 10));
```
vaibhavmurkute, line 142: Non-blocking:
This sets a limit on `DEFAULT_INTERVAL_SECONDS`. May be we should document this in a comment beside this default that its value should be set greater than 10. 
(Also I'd like this offset to be as small as possible, considering we're making this tradeoff for few ms. Can we go lower than 10?)

Nelsonochoam: I don't really think it matters given it runs once a day and it is a non deterministic value.

vaibhavmurkute (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,16 +127,44 @@ public boolean isRunning() {
...
+        if (connectionState == NetworkState.ConnectionState.NETWORK_UP) {
+            logger.atInfo("Network back up - refreshing certificates");
+            run();
+        }
+    }
+
+    private boolean canRun() {
+        if (nextRun.get() == null) {
+            return true;
+        }
```
vaibhavmurkute, line 162: `nextRun` will never be `null`

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -190,4 +215,79 @@ void GIVEN_certsAndThings_WHEN_certificateAssociatedToMoreThanOneThing_THEN_itDo
...
+            throws InterruptedException {
+       Thread t1 = new Thread(() -> { backgroundRefresh.run(); });
+       Thread t2 = new Thread(() -> { backgroundRefresh.run(); });
+       t1.start();
+       t2.start();
+       t1.join();
+       t2.join();
+
+        verify(iotAuthClientFake, times(1)).getThingsAssociatedWithCoreDevice();
+    }
```
vaibhavmurkute, line 257: nice!

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -146,15 +179,19 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
+                    "get-things-associated-with-core-device", logger);
 
             Stream<Thing> localThings = thingRegistry.getAllThings();
-            ThingAssociations associations =  ThingAssociations.create(cloudThings);
+            ThingAssociations associations = ThingAssociations.create(cloudThings);
             associations.setLocalThings(localThings);
             return Optional.of(associations);
+        } catch (AccessDeniedException e) {
+            logger.atInfo().cause(e).log(
+                "Did not refresh local certificates. To enable certificate refresh add a policy to the core device"
```
MikeDombo, line 190: space between device and that

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,16 +127,44 @@ public boolean isRunning() {
...
+        }
+    }
+
+    private boolean canRun() {
+        if (nextRun.get() == null) {
+            return true;
+        }
+
+        Instant now = Instant.now();
+        return now.compareTo(nextRun.get()) >= 0;
```
MikeDombo, line 165: use isbefore or isafter instead of compareto to be more clear about the intention.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,16 +127,44 @@ public boolean isRunning() {
...
     @Override
-    public void run() {
+    public synchronized void run() {
         if (isNetworkDown()) {
             logger.debug("Network is down - not refreshing certificates");
             return;
         }
 
+        if (!canRun()) {
+            logger.info("Certificates were refreshed recently - retrying later");
```
MikeDombo, line 137: drop to debug, like with network being down?

How often would you think this condition would trigger?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -146,15 +179,19 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
-                            "get-things-associated-with-core-device", logger);
+                    "get-things-associated-with-core-device", logger);
 
             Stream<Thing> localThings = thingRegistry.getAllThings();
-            ThingAssociations associations =  ThingAssociations.create(cloudThings);
+            ThingAssociations associations = ThingAssociations.create(cloudThings);
             associations.setLocalThings(localThings);
             return Optional.of(associations);
+        } catch (AccessDeniedException e) {
+            logger.atInfo().cause(e).log(
```
MikeDombo, line 189: in this case, logging the stacktrace is not every useful, so omit the `cause`.

This should probably be at warn too, not just info.

Verify with testing that we actually get an `AccessDeniedException`, we have found that that exception isn't actually thrown in many cases, and we needed to change our code to catch any sdk exception and then look at the http status code.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Which debug you want me to drop, the one above?.

This condition would trigger based on how choppy the device connection is. If it flickers a lot between up and down this could get triggered multiple times .

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Well I want a `greaterThanOrEqual` behaviour. It is either that of comparing if it is after or equal like  `now.equals(nextRun) || now.isAfter(nextRun)`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
MikeDombo: No, this one which is at info for `!canRun`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Yup, the only problem is how it does the equality comparison of instant. I could move it to a method like `isGreatherThanOrEquals(now, nextRun)`. Thoughts?

```java
Instant now = Instant.now();
now.equals(Instant.ofEpochMilli(now.toEpochMilli())); // Returns false
```

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,16 +127,44 @@ public boolean isRunning() {
...
+    public synchronized void run() {
         if (isNetworkDown()) {
             logger.debug("Network is down - not refreshing certificates");
             return;
         }
 
+        if (!canRun()) {
+            logger.info("Certificates were refreshed recently - retrying later");
+            return;
+        }
```
jbutler, line 139: Customer networks can be quite chatty - this isn't appropriate for INFO level. I'd actually suggest just removing the log entirely and only log when we do refresh.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
MikeDombo: Why is equality a problem? Just use `.equals`, it does the right thing.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,16 +127,44 @@ public boolean isRunning() {
...
 
+
+    /**
+     * Handler to react to network changes.
+     * @param connectionState - A network state
+     */
+    @Override
+    public void accept(NetworkState.ConnectionState connectionState) {
+        if (connectionState == NetworkState.ConnectionState.NETWORK_UP) {
+            logger.atInfo("Network back up - refreshing certificates");
```
jbutler, line 154: Same here - remove

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
MikeDombo:
```
public boolean equals(Object otherInstant) {
        if (this == otherInstant) {
            return true;
        }
        if (otherInstant instanceof Instant) {
            Instant other = (Instant) otherInstant;
            return this.seconds == other.seconds &&
                   this.nanos == other.nanos;
        }
        return false;
    }
```

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Good point. I will write a test for that. I think info is fine, we intentionally wanted to make it so that if that permission is not set the refresh is "disabled". I could go with warn though, don't really know

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,16 +127,44 @@ public boolean isRunning() {
...
             logger.debug("Network is down - not refreshing certificates");
             return;
         }
 
+        if (!canRun()) {
+            logger.info("Certificates were refreshed recently - retrying later");
+            return;
+        }
+
+        nextRun.set(Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS)));
```
jbutler, line 141: What if the refresh fails?

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -190,4 +215,79 @@ void GIVEN_certsAndThings_WHEN_certificateAssociatedToMoreThanOneThing_THEN_itDo
...
+
+        backgroundRefresh.run();
+
+        Instant aDayLater = now.plus(Duration.ofHours(24));
+        mockInstant(aDayLater.toEpochMilli());
+
+        backgroundRefresh.run();
+        verify(iotAuthClientFake, times(2)).getThingsAssociatedWithCoreDevice();
+    }
+
```
jbutler, line 272: I think there's a missing test case here. If the network call fails, it should be retried on the next network up event

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
MikeDombo: You can't convert to millis and back again, it loses information. Instants contain nanoseconds too, so that granularity of data will be lost, which is why you're seeing that behavior.

If you've converted from millis, then you need to also convert what you're comparing from to millis and back to round off the nanoseconds.

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
Nelsonochoam: K this should be testing that scenario now https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/175/files#diff-c6f02a506e0f67706541eefa03ba3a070eb33963da172e13ee67b590423d1f83R232

Nelsonochoam: Basically if it fails it can retry the net time it gets called (which could be triggered by an network up or the scheduler)

Nelsonochoam: This gets tested now here https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/175/files#diff-c6f02a506e0f67706541eefa03ba3a070eb33963da172e13ee67b590423d1f83R231. I did not use a InternalServer exception because then it enters the retry policy and hangs in there forever

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -170,12 +169,6 @@ public Stream<Thing> getThingsAssociatedWithCoreDevice() {
                 return response.associatedClientDevices().stream()
                         .map(AssociatedClientDevice::thingName)
                         .map(Thing::of);
-            } catch (Exception e) {
```
Nelsonochoam, line 173: This got removed because otherwise using the RetryUtils doesn't work very well given we don't just simply want to retry on `CloudServiceInteractionException` we want the caller to have more control over what to do with the exceptions

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -122,14 +127,42 @@ public boolean isRunning() {
...
 
+        if (!canRun()) {
+            return;
+        }
+
         Optional<ThingAssociations> associations = getThingsAssociatedWithCoreDevice();
-        associations.ifPresent(this::refresh);
+        associations.ifPresent(a -> {
+            this.refresh(a);
+            nextRun.set(Instant.now().plus(Duration.ofSeconds(DEFAULT_INTERVAL_SECONDS)));
```
jbutler, line 143: Now this may potentially not refresh for close to 48 hours. The next scheduled refresh needs to be pushed when this happens, right?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -118,18 +115,72 @@ public boolean isRunning() {
...
+    public Instant getNextScheduledRun() {
+       return nextScheduledRun.get();
+    }
+
+    /**
+     * Returns the last Instant the task ran successfully.
+     */
+    public Instant getLastRan() {
+        return lastRan.get();
+    }
```
jbutler, line 131: Not a huge fan of these methods - right now you're just using these in tests. That's ok, but I'd prefer to test actual observable behavior instead of asking the class what it did / intends to do.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -81,21 +87,12 @@ public BackgroundCertificateRefresh(
...
         if (scheduledFuture != null) {
-           return;
+            return;
         }
 
-        logger.info("Starting background refresh of client certificates every {} seconds", intervalSeconds);
-        scheduledFuture =
-            scheduler.scheduleAtFixedRate(this, intervalSeconds, intervalSeconds, TimeUnit.SECONDS);
+        logger.info("Starting background refresh of client certificates every {} seconds", DEFAULT_INTERVAL_SECONDS);
+        scheduleNextRun();
```
jbutler, line 95: Did you consider running immediately on boot instead of scheduling the first refresh in 24 hours? I think we could defend either. Curious about your thoughts.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -208,7 +264,12 @@ private void refreshCertificateValidity(Certificate certificate) {
         }
 
         VerifyIotCertificate useCase = useCases.get(VerifyIotCertificate.class);
-        useCase.apply(certPem.get());
+        try {
+            useCase.apply(certPem.get());
```
jbutler, line 268: When is this throwing a RuntimeException? We should fix that

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: The main reason I did not do it on bootup is because a lot of being able to refresh certificates relies on the fact that we have the certificate PEMs stored. Given the current implementation stores things in memory (for the pems) if we can't find the PEMs then we can't update them

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Agree... the tricky part was testing the scheduler. Technically I could make it run every second but the non determinism would make the test flaky this way is not perfect but gave me a sense of whether it was doing the right thing or not?. Do you know of a better way we could test it?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Here is the reasoning. Currently it does not if you check the code but if someone were to make changes that introduce that and we don't catch it here we run the risk of the thread dying and not getting rescheduled for a next run

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -94,16 +113,30 @@ void setup() throws DeviceConfigurationException, KeyStoreException {
...
     }
 
     @AfterEach
     void cleanup() throws IOException {
+        this.clockMock.ifPresent(ScopedMock::close);
         configurationTopics.getContext().close();
     }
 
+    @SuppressWarnings("PMD.CloseResource")
+    private void mockInstant(long expected) {
```
jcosentino11, line 130: thought for a potential follow-up:   as an alternative, we could have our own implementation of `Clock` that lets us mutate the instant and then inject that in `BackgroundCertificateRefresh`.   Would express intent more clearly and avoid the need to override standard library methods :)

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -190,4 +222,184 @@ void GIVEN_certsAndThings_WHEN_certificateAssociatedToMoreThanOneThing_THEN_itDo
...
+    void GIVEN_triggeredByMultipleThreads_WHEN_called_THEN_itShouldOnlyRunOnceForThatTimeWindow()
+            throws InterruptedException {
+       Thread t1 = new Thread(() -> { backgroundRefresh.run(); });
+       Thread t2 = new Thread(() -> { backgroundRefresh.run(); });
+       t1.start();
+       t2.start();
+       t1.join();
+       t2.join();
+
+        verify(iotAuthClientFake, times(1)).getThingsAssociatedWithCoreDevice();
```
jcosentino11, line 275: this isn't necessarily guaranteed, there would still be a (small) possibility that t1 completes before t2 starts.   it's probably ok in practice, just wanted to call it out

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -118,18 +115,72 @@ public boolean isRunning() {
...
     /**
      * Runs verifyIotCertificate useCase for all the registered client certificate PEMs.
      */
     @Override
-    public void run() {
+    public synchronized void run() {
         if (isNetworkDown()) {
             logger.debug("Network is down - not refreshing certificates");
             return;
         }
```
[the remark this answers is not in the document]
Nelsonochoam: Could we do that on a separate PR? We don't have a test that checks what happens if the code below runs when the network is down probably nothing but I'd rather put something in place to test it

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
jcosentino11: https://github.com/mockito/mockito/wiki/Using-Spies-(and-Fakes)#case-study-2-to-test-a-jobscheduler-that-uses-scheduledexecutorservice has an example

jbutler (approved):

jcosentino11 (approved):
bit late to the party here, just some side comments:

PR #176, opened 2022-10-21
fix: dispatch network change events asynchronously

jbutler:
Network state change events are running in a CRT thread, which means it is not safe to do any meaningful work since we'll be blocking the CRT. This change emits network state change events on an executor service so that callees can safely do whatever work is needed.

**Issue #, if available:**

**Description of changes:**

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/infra/NetworkState.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/infra/NetworkState.java
@@ -11,10 +11,13 @@ import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
 import javax.inject.Inject;
 
 public final class NetworkState {
+    private final ExecutorService executorService;
+
     public enum ConnectionState {
         NETWORK_UP,
         NETWORK_DOWN,
@@ -41,12 +44,14 @@ public final class NetworkState {
 
     /**
      * Represents the current network state.
-     * @param mqttClient   MqttClient wrapper.
+     * @param mqttClient      MqttClient wrapper.
+     * @param executorService Executor service used to dispatch connection state change events.
      */
     @Inject
-    public NetworkState(MqttClient mqttClient) {
+    public NetworkState(MqttClient mqttClient, ExecutorService executorService) {
         this.mqttClient = mqttClient;
         mqttClient.addToCallbackEvents(onConnect, callbacks);
+        this.executorService = executorService;
     }
 
     public void registerHandler(Consumer<ConnectionState> networkChangeHandler) {
@@ -83,7 +88,7 @@ public final class NetworkState {
     private void emitNetworkUp() {
         if (isRunning()) {
             for (Consumer<ConnectionState> handler : handlers) {
-                handler.accept(ConnectionState.NETWORK_UP);
+                executorService.submit(() -> handler.accept(ConnectionState.NETWORK_UP));
             }
         }
     }
@@ -91,7 +96,7 @@ public final class NetworkState {
     private void emitNetworkDown() {
         if (isRunning()) {
             for (Consumer<ConnectionState> handler : handlers) {
-                handler.accept(ConnectionState.NETWORK_DOWN);
+                executorService.submit(() -> handler.accept(ConnectionState.NETWORK_DOWN));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/infra/NetworkState.java
```diff
@@ -11,10 +11,13 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
 import javax.inject.Inject;
 
 public final class NetworkState {
+    private final ExecutorService executorService;
+
```
Nelsonochoam, line 20: Test please :D

src/main/java/com/aws/greengrass/clientdevices/auth/infra/NetworkState.java
```diff
@@ -83,15 +88,15 @@ public ConnectionState getConnectionStateFromMqtt() {
...
+                executorService.submit(() -> handler.accept(ConnectionState.NETWORK_UP));
             }
         }
     }
 
     private void emitNetworkDown() {
         if (isRunning()) {
             for (Consumer<ConnectionState> handler : handlers) {
-                handler.accept(ConnectionState.NETWORK_DOWN);
+                executorService.submit(() -> handler.accept(ConnectionState.NETWORK_DOWN));
```
MikeDombo, line 99: cancel future if network state flips again, or something like that? Any concerns for duplicating calls?

Should each callback run in another thread (which is what this is doing), or should it run the loop in the executor? The current implementation will end up most likely starting _n_ threads for _n_ handlers.

Nelsonochoam (dismissed):

src/test/java/com/aws/greengrass/clientdevices/auth/infra/NetworkStateTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @AfterEach
+    void afterEach() {
+        fjp.shutdown();
+    }
+
+    void assertNetworkDomainEventOnNetworkEvent(Runnable networkEvent, NetworkState.ConnectionState expectedState)
+            throws InterruptedException {
+        CountDownLatch cdl = new CountDownLatch(1);
```
Nelsonochoam, line 60: I like this helper but it hides what the steps of the tests are. When we read the test we just see a call to an assertion  without any step that setup pre conditions, run the system under test and then assert on things. I think we might want to split it into


```java
private Runnable givenSubscribreToNetworkChanges() {
          CountDownLatch cdl = new CountDownLatch(1);


        Consumer<NetworkState.ConnectionState> networkStateConsumer = (state) -> {
            connState.set(state);
            callbackThreadId.set(Thread.currentThread().getId());
            cdl.countDown();
        };
        networkState.registerHandler(networkStateConsumer);

      return () -> {
             networkEvent.run();
            cdl.await(1, TimeUnit.SECONDS);
      }
}

```

Then your test becomes

```java
 // Given
  Runnable onNetworkUp = givenSubscribreToNetworkChanges(() -> onConnectCaptor.getValue().onConnect(true),
                NetworkState.ConnectionState.NETWORK_UP);

// When 
         onNetworkUp.run();

// Then
        assertThat(callbackThreadId.get(), is(not(curThreadId)));
        assertThat(connState.get(), is(NetworkState.ConnectionState.NETWORK_UP));
```

src/test/java/com/aws/greengrass/clientdevices/auth/infra/NetworkStateTest.java
```diff
@@ -0,0 +1,103 @@
...
+        networkEvent.run();
+
+        // Assert that NETWORK_UP event was emitted, and that it was emitted in a separate thread.
+        // We're checking this because network change events are triggered in a CRT thread, which
+        // we should not ever block
+        assertTrue(cdl.await(1, TimeUnit.SECONDS));
+        assertThat(callbackThreadId.get(), is(not(curThreadId)));
+        assertThat(connState.get(), is(expectedState));
+    }
+
```
Nelsonochoam, line 81: There is also one test missing. It seems from the code that the intent is that multiple handlers get ran by the same thread rather than trigger each handler on a single thread. If that is a decision we are firm on the we should have a test with more than one handler and check they both got triggered by the same thread

src/test/java/com/aws/greengrass/clientdevices/auth/infra/NetworkStateTest.java
Nelsonochoam: The snippet was meant as an example I did not run it.  The `@ParametrizedTest` looks good as well it is achieving something similar, the only thing I would change if the method name to something like `GIVEN_subscriptionToNetworkChanges_WHEN_connectionStateChanges_THEN_subscribersRunOnASeparateThread`

Nelsonochoam (approved):

vaibhavmurkute (approved):

PR #177, opened 2022-10-21
feat: support pagination on refreshing things and certificates

Nelsonochoam:
**Description of changes:**
This changes the call we make to get the things attached to the core device to use the paginator provided by the aws sdk. It provides a stream that will call all the pages and allow us to process the received response in chunks. I reworked the refresh logic so that we could take advantage of the fact that the response pages can be streamed so we can process one page at a time.

There are still some additional improvements that could be made to improve performance now that we have this streams such as parallel processing of things and certificate updates as well as the cleanup (which can be done on a separate thread).

For clients with cores with more than 100 clients this is required. Otherwise we only ever refresh the first 100d

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
@@ -52,6 +52,7 @@ import java.util.HashMap;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 import static com.aws.greengrass.clientdevices.auth.helpers.CertificateTestHelpers.createClientCertificate;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
@@ -194,10 +195,10 @@ public class OfflineAuthenticationTest {
        String clientBPem = CertificateHelper.toPem(clientBCrt);
        iotAuthClientFake.activateCert(clientBPem);
        when(networkStateMock.getConnectionStateFromMqtt()).thenReturn(NetworkState.ConnectionState.NETWORK_UP);
-       String thingOne = "ThingOne";
-       iotAuthClientFake.attachCertificateToThing(thingOne, clientAPem);
-       String thingTwo = "ThingTwo";
-       iotAuthClientFake.attachCertificateToThing(thingTwo, clientBPem);
+       Supplier<String> thingOne =  () -> "ThingOne";
+       iotAuthClientFake.attachCertificateToThing(thingOne.get(), clientAPem);
+       Supplier<String> thingTwo = () -> "ThingTwo";
+       iotAuthClientFake.attachCertificateToThing(thingTwo.get(), clientBPem);
 
        iotAuthClientFake.attachThingToCore(thingOne);
        iotAuthClientFake.attachThingToCore(thingTwo);
@@ -214,13 +215,13 @@ public class OfflineAuthenticationTest {
 
        // Simulate a client connecting and generating a session
        api.getClientDeviceAuthToken("mqtt",   new HashMap<String, String>() {{
-           put("clientId", thingOne);
+           put("clientId", thingOne.get());
            put("certificatePem", clientAPem);
            put("username", "foo");
            put("password", "bar");
        }});
        api.getClientDeviceAuthToken("mqtt", new HashMap<String, String>() {{
-           put("clientId", thingTwo);
+           put("clientId", thingTwo.get());
            put("certificatePem", clientBPem);
            put("username", "baz");
            put("password", "fizz");
@@ -235,8 +236,8 @@ public class OfflineAuthenticationTest {
 
        // Check state before refresh of thing attachments
        ThingRegistry thingRegistry = kernel.getContext().get(ThingRegistry.class);
-       Thing ogThingA = thingRegistry.getOrCreateThing(thingOne);
-       Thing ogThingB = thingRegistry.getOrCreateThing(thingTwo);
+       Thing ogThingA = thingRegistry.getOrCreateThing(thingOne.get());
+       Thing ogThingB = thingRegistry.getOrCreateThing(thingTwo.get());
        assertEquals(ogThingA.certificateLastAttachedOn(ogCertA.getCertificateId()).get().toEpochMilli(), now.toEpochMilli());
        assertEquals(ogThingB.certificateLastAttachedOn(ogCertB.getCertificateId()).get().toEpochMilli(), now.toEpochMilli());
 
@@ -262,8 +263,8 @@ public class OfflineAuthenticationTest {
        assertFalse(certB.isPresent());
 
        // Verify thing certificate attachments got updated after refresh
-       Thing thingA = thingRegistry.getThing(thingOne);
-       Thing thingB = thingRegistry.getThing(thingTwo);
+       Thing thingA = thingRegistry.getThing(thingOne.get());
+       Thing thingB = thingRegistry.getThing(thingTwo.get());
        assertEquals(
            thingA.certificateLastAttachedOn(ogCertA.getCertificateId()).get().toEpochMilli(),
            anHourLater.toEpochMilli()
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -20,7 +20,8 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Pair;
 import com.aws.greengrass.util.RetryUtils;
-import software.amazon.awssdk.services.greengrassv2data.model.AccessDeniedException;
+import software.amazon.awssdk.services.greengrassv2.model.AccessDeniedException;
+import software.amazon.awssdk.services.greengrassv2.model.AssociatedClientDevice;
 import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
 
@@ -28,6 +29,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -163,7 +165,6 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
         scheduledFuture = scheduler.schedule(this, duration.getSeconds(), TimeUnit.SECONDS);
     }
 
-
     /**
      * Handler to react to network changes.
      * @param connectionState - A network state
@@ -196,9 +197,13 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
                 .retryableExceptions(Arrays.asList(ThrottlingException.class, InternalServerException.class)).build();
 
         try {
-            Stream<Thing> cloudThings = RetryUtils.runWithRetry(
+            Stream<List<AssociatedClientDevice>> cloudAssociatedDevices = RetryUtils.runWithRetry(
                     retryConfig, iotAuthClient::getThingsAssociatedWithCoreDevice,
-                    "get-things-associated-with-core-device", logger);
+                            "get-things-associated-with-core-device", logger);
+
+            Stream<Thing> cloudThings = cloudAssociatedDevices.flatMap(List::stream)
+                        .map(AssociatedClientDevice::thingName)
+                        .map(name -> Thing.of(name, Thing.Source.CLOUD));
 
             return Optional.of(cloudThings);
         } catch (AccessDeniedException e) {
@@ -239,14 +244,14 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
     private void refreshCertificateAttachments(Thing thing) {
         AtomicReference<Thing> thingWithCerts = new AtomicReference<>(thing);
 
-        if (thingWithCerts.get().getSource() != Thing.Source.LOCAL) {
-            thingWithCerts.set(thingRegistry.getThing(thing.getThingName()));
-        }
-
         if (thingWithCerts.get() == null) {
-            logger.atInfo().kv("thingName", thing.getThingName())
+            logger.atDebug().kv("thingName", thing.getThingName())
                     .log("No local version found for thing. Not refreshing thing certificate attachments");
             return;
+        }
+
+        if (thingWithCerts.get().getSource() != Thing.Source.LOCAL) {
+            thingWithCerts.set(thingRegistry.getThing(thing.getThingName()));
         }
 
         Set<String> thingCertificateIds = thingWithCerts.get().getAttachedCertificateIds().keySet();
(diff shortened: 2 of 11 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -190,77 +210,65 @@ private void refreshCertificateThingAttachment(Thing thing) {
...
+                    return;
+                }
 
-        /**
-         * Gets a stream of all the things that are still associated to the core.
-         */
-        public Stream<Thing> getThings() {
-            return attachedThings.stream();
-        }
+                alreadyRefreshed.add(certificateId);
```
vaibhavmurkute, line 250: should we have this after L#215 `verifyCertificate.apply(certPem.get())`,  considering verifyCertificate usecase might fail.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -160,24 +160,44 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
+
+        if (thingWithCerts.get().getSource() != Thing.Source.LOCAL) {
+            thingWithCerts.set(thingRegistry.getThing(thing.getThingName()));
+        }
+
+        if (thingWithCerts.get() == null) {
+            logger.atInfo().kv("thingName", thing.getThingName())
+                    .log("No local version found for thing. Not refreshing thing certificate attachments");
+            return;
+        }
```
vaibhavmurkute, line 196: this null-check should be before L#188 checking the source of this thing.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -160,24 +160,44 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
+
+        // Clean up the registries by providing the names of the things that should still be attached and
+        // the ids of the certificates that are still attached to a thing.
+        cleanUpRegistries(attachedThingNames, refreshCertificates.getLeft());
     }
 
-    private void refreshCertificateThingAttachment(Thing thing) {
-        VerifyThingAttachedToCertificate useCase = useCases.get(VerifyThingAttachedToCertificate.class);
+    private void refreshCertificateAttachments(Thing thing) {
+        AtomicReference<Thing> thingWithCerts = new AtomicReference<>(thing);
```
vaibhavmurkute, line 186: does it _need_ to Atomic? Do we anticipate any race with this reference?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -160,24 +160,44 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
+
+        if (thingWithCerts.get().getSource() != Thing.Source.LOCAL) {
+            thingWithCerts.set(thingRegistry.getThing(thing.getThingName()));
+        }
+
+        if (thingWithCerts.get() == null) {
+            logger.atInfo().kv("thingName", thing.getThingName())
+                    .log("No local version found for thing. Not refreshing thing certificate attachments");
```
vaibhavmurkute, line 194: log at DEBUG?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Yeah it needs to because I am using it within a lambda below

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Good catch

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -40,6 +40,11 @@ public final class Thing implements AttributeProvider, Cloneable {
     // map of certificate ID to the time this certificate was known to be attached to the Thing
     private final Map<String, Instant> attachedCertificateIds;
     private boolean modified = false;
+    private final Source source;
+
+    public enum Source {
+        LOCAL, CLOUD, UNSET;
+    }
```
vaibhavmurkute, line 47: feel free to ignore, just a thought:
Can we live _without_ maintaining this `Source` metadata? I see it is used only to distinguish between things stored locally (in registry) and thing attached to the core in the cloud and they are treated the same way everywhere else. Theoretically, with current use case, we only need thing-names of the thing attached to the core. Maintaining a `Source` metadata just to be able to create `Thing` domain object for these attached thing-names seems bit of an overhead.

(rest of the changes look good to me)

vaibhavmurkute (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -137,21 +142,22 @@ public void run() {
...
-            Stream<Thing> cloudThings = RetryUtils.runWithRetry(
+            Stream<List<AssociatedClientDevice>> cloudAssociatedDevices = RetryUtils.runWithRetry(
                     retryConfig, iotAuthClient::getThingsAssociatedWithCoreDevice,
                             "get-things-associated-with-core-device", logger);
 
-            Stream<Thing> localThings = thingRegistry.getAllThings();
-            ThingAssociations associations =  ThingAssociations.create(cloudThings);
-            associations.setLocalThings(localThings);
-            return Optional.of(associations);
+            Stream<Thing> cloudThings = cloudAssociatedDevices.flatMap(List::stream)
```
jcosentino11, line 204: nit: i think flatMap would fit better in `IotAuthClient#getThingsAssociatedWithCoreDevice`, lets us abstract paging details and easier to reason about the code.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -160,24 +166,44 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
-        certificateRegistry.getAllCertificates()
-                .forEach(certificate -> {
-            if (associations.isAttachedToThing(certificate)) {
-                refreshCertificateValidity(certificate);
-            } else {
-                certificateRegistry.deleteCertificate(certificate);
-            }
-        });
+    private void refresh(Stream<Thing> thingsAttachedToCore) {
+        Pair<Set<String>, Consumer<Thing>> refreshCertificates = refreshCertificateValidity();
```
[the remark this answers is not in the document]
Nelsonochoam: We can but there is catch, which is we wouldn't be processing one page at a time we would basically have to collect all the pages string names before running this loop. I mean we could have the paginated response return an iterator and loop per page lazily and achieve the same with next for loops.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
jcosentino11: if we use a regular for-each loop though, then we don't need AtomicReference

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -215,24 +218,45 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
+
+        Set<String> attachedThingNames = thingsAttachedToCore
+                // Refresh the thing certificate attachments
+                .peek(this::refreshCertificateAttachments)
```
jcosentino11, line 226: in general we should avoid using peek to do real work, as it [exists mainly to support debugging](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#peek-java.util.function.Consumer-), and has some [pitfalls](https://rules.sonarsource.com/java/RSPEC-3864).

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -194,10 +195,10 @@ void GIVEN_storedCertificates_WHEN_refreshEnabled_THEN_storedCertificatesRefresh
        String clientBPem = CertificateHelper.toPem(clientBCrt);
        iotAuthClientFake.activateCert(clientBPem);
        when(networkStateMock.getConnectionStateFromMqtt()).thenReturn(NetworkState.ConnectionState.NETWORK_UP);
-       String thingOne = "ThingOne";
-       iotAuthClientFake.attachCertificateToThing(thingOne, clientAPem);
-       String thingTwo = "ThingTwo";
-       iotAuthClientFake.attachCertificateToThing(thingTwo, clientBPem);
+       Supplier<String> thingOne =  () -> "ThingOne";
```
Nelsonochoam, line 198: This was so we could simulate failures by passing to the fake something that could explode

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -117,43 +123,51 @@ public Optional<Certificate> getIotCertificate(String certificatePem) throws Inv
         }
 
         @Override
-        @SuppressWarnings("PMD.AvoidCatchingGenericException")
         public boolean isThingAttachedToCertificate(Thing thing, Certificate certificate) {
+            if (Objects.isNull(certificate)) {
```
Nelsonochoam, line 127: Override this so we can have a version that just needs the certificateId. In cases we don't have the pem but we do have the id (when loaded from the registry)

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
```diff
@@ -43,23 +43,23 @@ public VerifyThingAttachedToCertificate(IotAuthClient iotAuthClient, ThingRegist
         this.networkState = networkState;
     }
 
-    private boolean verifyLocally(Thing thing, Certificate certificate) {
-        logger.atInfo().kv("thing", thing.getThingName()).kv("certificate", certificate.getCertificateId())
+    private boolean verifyLocally(Thing thing, String certificateId) {
```
Nelsonochoam, line 46: This was so we use the certificateId instead of requiring a full cert.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClientFake.java
```diff
@@ -98,14 +102,84 @@ public Optional<Certificate> getIotCertificate(String certificatePem) throws Inv
...
+    }
+
+    public void detachThingFromCore(Supplier<String> thingName) {
+        thingsAttachedToCore.remove(thingName);
+    }
+
     @Override
-    public Stream<Thing> getThingsAssociatedWithCoreDevice() {
-        return thingsAttachedToCore.stream().map(Thing::of);
+    public Stream<List<AssociatedClientDevice>> getThingsAssociatedWithCoreDevice() {
```
Nelsonochoam, line 129: This code below just simulates what the AWS SDK is doing for pagination. We could have used just a stream that creates lists but I wanted it to work as close as possible to the sdk given it lazily loads each page and we could simulate errors better

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -215,113 +217,80 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
-                useCase.apply(new VerifyThingAttachedToCertificateDTO(thing, certificate));
-            } catch (InvalidCertificateException | RuntimeException e) {
-                logger.atWarn().cause(e).kv("thing", thing.getThingName()).kv("certificate", certificateId)
-                        .log("Failed to verify thing attached to certificate");
+                useCases.get(VerifyThingAttachedToCertificate.class).apply(
+                        new VerifyThingAttachedToCertificateDTO(thingName, certificateId));
+            } catch (RuntimeException e) {
+                logger.atWarn().cause(e).kv("thingName", thing.getThingName()).kv("certificate", certificateId)
+                        .log("Failed to verify thing certificate - certificate pem is invalid");
             }
```
vaibhavmurkute, line 256: Do we anticipate RuntimeException here? How do we know it is because of the invalid PEM?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -215,113 +217,80 @@ private Optional<ThingAssociations> getThingsAssociatedWithCoreDevice() {
...
 
-        VerifyIotCertificate useCase = useCases.get(VerifyIotCertificate.class);
         try {
-            useCase.apply(certPem.get());
+            useCases.get(VerifyIotCertificate.class).apply(certPem.get());
         } catch (RuntimeException e) {
-            logger.atWarn().cause(e).kv("certificateId", certificate.getCertificateId())
-                    .log("Failed to verify certificate validity");
+            logger.atWarn().kv("certificateId", certificateId).cause(e).log("Failed to verify certificate validity");
         }
```
vaibhavmurkute, line 277: same as above, how about catching checked exceptions? Does `VerifyIotCertificate` use case throw RuntimeException?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -117,43 +123,51 @@ public Optional<Certificate> getIotCertificate(String certificatePem) throws Inv
...
+                return false;
+            }
+            return isThingAttachedToCertificate(thing, certificate.getCertificateId());
+        }
+
+        @Override
+        @SuppressWarnings("PMD.AvoidCatchingGenericException")
+        public boolean isThingAttachedToCertificate(Thing thing, String certificateId) {
```
vaibhavmurkute, line 135: make this method `private` ?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: The `VerifyThingAttachedToCertificate` doesn't really requires a certificate, we only need the certificateId so I am passing the primitive to the DTO and the useCase can decide what to do. We won't get an `InvalidCertificateException` given we are not building a certificate. The reason I am catching a generic `RuntimeException` is due to the fact that if someone makes a mistake on the `VerifyThingAttachedToCertificate` and introduces one, then the scheduled thread will die and the refresh would stop.

Right now, the way a runtime exception can happen is if the cloud throws an internal server error or a throttle error and without this we would be on the scenario I described above.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: Same response as above 😂

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
jcosentino11: To vaibhav's point, if RuntimeException happens on ISE or throttle, then we should probably leave out "certificate pem is invalid"

Also, might be worth having a catch-all in the `run` method to provide total assurance the thread won't get killed

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
jbutler: +1 to this, we should do a complete audit of all logs. Can handle that separately.

Also, our exception handling story is not quite good enough for me to feel comfortable with these runtime exceptions. Let's also take a closer look at that. Ideally use cases won't throw them

jbutler (approved):

PR #178, opened 2022-10-26
feat: persist client certs

jbutler:
**Issue #, if available:**

**Description of changes:**

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
@@ -15,7 +15,6 @@ import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.security.SecurityService;
 import com.aws.greengrass.security.exceptions.KeyLoadingException;
 import com.aws.greengrass.security.exceptions.ServiceUnavailableException;
-import com.aws.greengrass.util.Digest;
 import com.aws.greengrass.util.EncryptionUtils;
 import com.aws.greengrass.util.FileSystemPermission;
 import com.aws.greengrass.util.platforms.Platform;
@@ -57,7 +56,6 @@ public class CertificateStore {
     private static final long   DEFAULT_CA_EXPIRY_SECONDS = 60 * 60 * 24 * 365 * 5; // 5 years
     private static final String DEFAULT_CA_CN = "Greengrass Core CA";
     private static final String CA_KEY_ALIAS = "CA";
-    private static final String DEVICE_CERTIFICATE_DIR = "devices";
     private static final String DEFAULT_KEYSTORE_FILENAME = "ca.jks";
     private static final String DEFAULT_CA_CERTIFICATE_FILENAME = "ca.pem";
 
@@ -265,29 +263,6 @@ public class CertificateStore {
         setCaKeyAndCertificateChain(CertificateHelper.ProviderType.DEFAULT, privateKey, certificates);
     }
 
-
-    public String loadDeviceCertificate(String certificateId) throws IOException {
-        return loadCertificatePem(certificateIdToPath(certificateId));
-    }
-
-    /**
-     * Store device certificate if not present.
-     *
-     * @param certificateId  Certificate ID
-     * @param certificatePem Certificate PEM
-     * @throws IOException   if unable to write certificate to disk
-     */
-    public void storeDeviceCertificateIfNotPresent(String certificateId, String certificatePem) throws IOException {
-        Path filePath = certificateIdToPath(certificateId);
-        if (!Files.exists(filePath)) {
-            saveCertificatePem(certificateIdToPath(certificateId), certificatePem);
-        }
-    }
-
-    Path certificateIdToPath(String certificateId) {
-        return workPath.resolve(DEVICE_CERTIFICATE_DIR).resolve(certificateId + ".pem");
-    }
-
     private void createAndStoreDefaultKeyStore(CAType caType) throws KeyStoreException {
         KeyPair kp;
 
@@ -418,10 +393,6 @@ public class CertificateStore {
                 EncryptionUtils.encodeToPem(CertificateHelper.PEM_BOUNDARY_CERTIFICATE,caCert.getEncoded()));
     }
 
-    private String loadCertificatePem(Path filePath) throws IOException {
-        return new String(Files.readAllBytes(filePath));
-    }
-
     private void saveCertificatePem(Path filePath, String certificatePem) throws IOException {
         Files.createDirectories(filePath.getParent());
         try (OutputStream writeStream = Files.newOutputStream(filePath)) {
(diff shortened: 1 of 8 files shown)


src/test/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStoreTest.java
```diff
@@ -165,58 +163,4 @@ public void GIVEN_all_possible_byte_values_WHEN_byteToAsciiCharacter_THEN_values
             assertThat(CertificateStore.byteToAsciiCharacter(c), lessThanOrEqualTo('~'));
         }
     }
-
```
jbutler, line 168: Going to look at these tests and port them over to `ClientCertificateStoreTest`

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateStore.java
```diff
@@ -265,29 +263,6 @@ private void setCaKeyAndCertificateChain(Key privateKey, Certificate... caCertif
         setCaKeyAndCertificateChain(CertificateHelper.ProviderType.DEFAULT, privateKey, certificates);
     }
 
-
```
jbutler, line 268: These methods have more or less just been moved (and re-worked ever so slightly)

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/ClientCertificateStore.java
```diff
@@ -5,86 +5,69 @@
...
 public class ClientCertificateStore {
-    private final KeyStore keyStore;
-
+    private static final String DEVICE_CERTIFICATE_DIR = "devices";
+    private final Path workPath;
+
+    // TODO: ideally we'd remove this, but right now our service init is somewhat fragile
+    //  so directly extracting the plugin work dir from the Kernel is easiest
+    @Inject
+    public ClientCertificateStore(Kernel kernel) throws IOException {
```
MikeDombo, line 30: Inject NucleusPaths instead since that's all you want.

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/ClientCertificateStore.java
```diff
@@ -93,20 +76,32 @@ public void removePem(String certificateId) throws KeyStoreException {
...
-            if (cert == null) {
-                return Optional.empty();
-            }
-
-            return Optional.of(CertificateHelper.toPem(cert));
-        } catch (KeyStoreException | CertificateEncodingException | IOException e) {
+            return Optional.of(loadDeviceCertificate(certificateId));
+        } catch (IOException e) {
+            // TODO: This was existing logic. Empty optional should be reserved for the case
+            //  where no cert is present. This should throw.
```
MikeDombo, line 87: Log?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/ClientCertificateStore.java
```diff
@@ -93,20 +76,32 @@ public void removePem(String certificateId) throws KeyStoreException {
...
             return Optional.empty();
         }
     }
+
+    private void saveCertificatePem(Path filePath, String certificatePem) throws IOException {
+        Files.createDirectories(filePath.getParent());
```
MikeDombo, line 93: Should be OK, but we have Utils.createPaths for this.

MikeDombo (commented):
Seems OK given it is mostly moving existing stuff. Assuming you will be adding new tests.

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateStoreTest.java
```diff
@@ -0,0 +1,115 @@
...
+            long count = stream.filter(Files::isRegularFile).count();
+            assertThat(count, is(certCount));
+        }
+    }
+
+    @Test
+    void GIVEN_workDirectoryWithClientCertificates_WHEN_getPem_THEN_correctCertificateReturned() throws IOException {
+        Function<Integer, String> getId = (i) -> "id-" + i;
+        Function<Integer, String> getPem = (i) -> "pem-" + i;
+        long certCount = 10;
```
Nelsonochoam, line 89: I don't really mind having these functions but it wouldn't hurt to create real certificates with real pems and store that real data instead of fake data, given it better represents the functionality of the class.

We already have helpers for creating a root ca so you could create a method that just creates a bunch of client certificates

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateStoreTest.java
```diff
@@ -0,0 +1,115 @@
...
+            String id = getId.apply(i);
+
+            assertThat(certificateStore.exists(id), is(true));
+            Optional<String> returnedPem = certificateStore.getPem(id);
+            assertThat(returnedPem.isPresent(), is(true));
+            assertThat(returnedPem.get(), is(getPem.apply(i)));
+        }
+
+        // Count files in work dir
+        try (Stream<Path> stream = Files.walk(workDir)) {
```
Nelsonochoam, line 110: Not a bad test, but this is an assertion that is easily bypassed if you just write the same number of random files with random content into the test folder.  A better assertion is to get the file by the certificate id and check the content matches. Or if we are exposing too many details about the implementation to the test we could completely remove this

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateStoreTest.java
```diff
@@ -0,0 +1,115 @@
...
+            long count = stream.filter(Files::isRegularFile).count();
+            assertThat(count, is(certCount));
+        }
+    }
```
Nelsonochoam, line 83: Missing one test for removal which is an important one to make sure we are not leaving files in the fs once a cert gets removed

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateStoreTest.java
jbutler: I started doing that, but then talked myself out of it (time sensitivity + performance reasons - our tests are already quite slow). But I don't disagree

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/ClientCertificateStoreTest.java
jbutler: This is essentially just the above test, but after a reload. The loop above does check retrieved certs against the expected value.

It may actually make sense for this test to be coupled with the implementation (e.g. know details of how certs are stored, and assert that they are written where we expect them to be written). The reason is because we could completely change the file format on disk and these tests would still pass. Whereas that would "soft break" a customer who upgraded.

Let's revisit in a followup

Nelsonochoam (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/ClientCertificateStore.java
```diff
@@ -5,108 +5,99 @@
...
         }
     }
+
+    private Path certificateIdToPath(String certificateId) {
+        return workPath.resolve(CLIENTS_DIRECTORY).resolve(certificateId + ".pem");
+    }
+
+    private String loadDeviceCertificate(String certificateId) throws IOException {
+        Path certPath = certificateIdToPath(certificateId);
+        return new String(Files.readAllBytes(certPath));
```
vaibhavmurkute, line 101: should we specify character set here? The default character-set it uses varies across platform (and users).

vaibhavmurkute (approved):

PR #179, opened 2022-10-27
fix: fix disabled flaky tests

Nelsonochoam:
**Description of changes:**
Some tests had been disabled because they were flaky due to the fact that they where just waiting for a call count to be met and not for the actual configuration changes to have take effect. This PR fixes that by introducing a test utility (which we have in other code bases) to wait for an expectation to be eventually true.

Re enable the flaky tests

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
@@ -16,7 +16,6 @@ import com.aws.greengrass.clientdevices.auth.certificate.CertificateHelper;
 import com.aws.greengrass.clientdevices.auth.certificate.CertificateStore;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
 import com.aws.greengrass.clientdevices.auth.connectivity.CISShadowMonitor;
-import com.aws.greengrass.clientdevices.auth.exception.CertificateChainLoadingException;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
 import com.aws.greengrass.clientdevices.auth.helpers.CertificateTestHelpers;
 import com.aws.greengrass.clientdevices.auth.helpers.TestHelpers;
@@ -29,8 +28,6 @@ import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
 import com.aws.greengrass.security.SecurityService;
-import com.aws.greengrass.security.exceptions.KeyLoadingException;
-import com.aws.greengrass.security.exceptions.ServiceUnavailableException;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
 import com.aws.greengrass.util.GreengrassServiceClientFactory;
 import com.aws.greengrass.util.Pair;
@@ -49,9 +46,7 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import software.amazon.awssdk.services.greengrassv2data.GreengrassV2DataClient;
 import software.amazon.awssdk.services.greengrassv2data.model.PutCertificateAuthoritiesRequest;
 
-import java.io.IOException;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.security.KeyPair;
 import java.security.NoSuchAlgorithmException;
@@ -60,7 +55,6 @@ import java.security.cert.X509Certificate;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
@@ -194,10 +188,7 @@ public class CustomCaConfigurationTest {
 
     @Test
     void Given_CustomCAConfiguration_WHEN_issuingAClientCertificate_THEN_itsSignedByCustomCA() throws
-            CertificateException, URISyntaxException, CertificateGenerationException, ExecutionException,
-            InterruptedException, ServiceLoadException, NoSuchAlgorithmException,
-            OperatorCreationException, IOException, KeyLoadingException, ServiceUnavailableException,
-            CertificateChainLoadingException {
+            Exception {
         Pair<X509Certificate[], KeyPair[]> credentials = givenRootAndIntermediateCA();
         X509Certificate[] chain = credentials.getLeft();
         X509Certificate intermediateCA = chain[0];
@@ -229,10 +220,7 @@ public class CustomCaConfigurationTest {
 
     @Test
     void GIVEN_CustomCAConfiguration_WHEN_whenGeneratingClientCerts_THEN_GGComponentIsVerified() throws
-            NoSuchAlgorithmException, CertificateException, OperatorCreationException, IOException,
-            URISyntaxException, KeyLoadingException, ServiceUnavailableException,
-            CertificateChainLoadingException, CertificateGenerationException, InterruptedException,
-            ServiceLoadException {
+            Exception {
         Pair<X509Certificate[], KeyPair[]> credentials = givenRootAndIntermediateCA();
         X509Certificate[] chain = credentials.getLeft();
         KeyPair[] certificateKeys = credentials.getRight();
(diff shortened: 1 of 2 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -215,31 +210,29 @@ void Given_CustomCAConfiguration_WHEN_issuingAClientCertificate_THEN_itsSignedBy
...
         givenCDAWithCustomCertificateAuthority(privateKeyUri, certificateUri);
 
-        asyncCall.getLeft().get(5, TimeUnit.SECONDS);
-        X509Certificate issuedClientCertificate = eventRef.get().getCertificate();
-        assertTrue(
-            CertificateTestHelpers.wasCertificateIssuedBy(intermediateCA, issuedClientCertificate),
-            String.format(
-                "Certificate not was not issued by intermediate CA %s", CertificateHelper.toPem(issuedClientCertificate)
-        ));
+        TestHelpers.eventuallyTrue(() -> {
```
jcosentino11, line 220: side thought---instead of busy-waiting, could introduce a new "custom ca configured" event emitted by `ConfigureCustomCertificateAuthority` and expose a subscription to the event via `ClientDevicesAuthServiceApi` or something.  would be a decent amount of work just for the sake of tests though 😅

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -196,11 +192,10 @@ private void subscribeToCertificateUpdates(GetCertificateRequest request)
...
     }
 
-    @Disabled("TODO - fix flaky test")
     @Test
     void Given_CustomCAConfiguration_WHEN_issuingAClientCertificate_THEN_itsSignedByCustomCA() throws
             CertificateException, URISyntaxException, CertificateGenerationException, ExecutionException,
-            InterruptedException, TimeoutException, ServiceLoadException, NoSuchAlgorithmException,
+            InterruptedException, ServiceLoadException, NoSuchAlgorithmException,
             OperatorCreationException, IOException, KeyLoadingException, ServiceUnavailableException,
             CertificateChainLoadingException {
```
vaibhavmurkute, line 200: If you're creating new revision, lets throw `Exception` instead.

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
```diff
@@ -0,0 +1,33 @@
...
+ * Helpers for generic assertion tooling
+ */
+public final class TestHelpers {
+    public static final long DEFAULT_GENERIC_POLLING_TIMEOUT_MILLIS =
+            TimeUnit.SECONDS.toMillis(10);
+
+    private TestHelpers() {
+    }
+
+    public static boolean eventuallyTrue(Supplier<Boolean> condition, long... optional) throws InterruptedException {
```
vaibhavmurkute, line 21: Caller of this method won't know what these long optional params correspond to, unless they check its code. I see the first long value corresponds to the timeout in millis, second long value is polling interval in millis. I'd suggest having these as separate parameter values than optionals (`long... optional`).

We can expose two overloaded methods:
```
public eventuallyTrue(Supplier<Boolean> condition, long timeoutMillis, long pollingIntervalMillis);
public eventuallyTrue(Supplier<Boolean> condition);  // uses default values for timeout and pollingInterval
```

jcosentino11, line 21: why not overload instead of varargs?  would still allow for optional args and have the benefits of arg names

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
```diff
@@ -0,0 +1,33 @@
...
+        long timeoutInMillis = optional.length >= 1 ? optional[0] : DEFAULT_GENERIC_POLLING_TIMEOUT_MILLIS;
+        long pollingIntervalInMillis = optional.length >= 2 ? optional[1] : 500;
```
vaibhavmurkute, line 23: nit: have this default `500` as a static final

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
```diff
@@ -0,0 +1,33 @@
...
+        final long startTime = System.currentTimeMillis();
+        while ((System.currentTimeMillis() - startTime) < timeoutInMillis) {
```
vaibhavmurkute, line 25: `Instant` might provide better time checking in this case, than writing our own comparison logic.

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
Nelsonochoam: Yeah I don't like adding stuff that we are only using for tests if we can avoid it. If we eventually have a usecase to expose that then I agree

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
Nelsonochoam: This is a copy past from EvergreenFeatures

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
Nelsonochoam: This is a copy past from EvergreenFeatures exact same code. I agree, I really did not pay to much attention to this literally just reusing what we already have written.

vaibhavmurkute (approved):

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
```diff
@@ -0,0 +1,43 @@
...
+    private TestHelpers() {
+    }
+
+    public static boolean eventuallyTrue(Supplier<Boolean> condition) throws InterruptedException {
+        return eventuallyTrue(condition, DEFAULT_GENERIC_POLLING_TIMEOUT_MILLIS, DEFAULT_POLLING_INTERVAL_MILLIS);
+    }
+
+    public static boolean eventuallyTrue(Supplier<Boolean> condition, long pollingTimeoutMillis,
+                                         long pollingIntervalMillis) throws InterruptedException {
+        final Instant tryUntil = Instant.now().plus(Duration.ofMillis(pollingTimeoutMillis));
```
jcosentino11, line 32:
```suggestion
        final Instant tryUntil = Instant.now().plusMillis(pollingTimeoutMillis);
```

jcosentino11 (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCaConfigurationTest.java
```diff
@@ -321,22 +307,21 @@ void GIVEN_managedCAConfiguration_WHEN_updatedToCustomCAConfiguration_THEN_serve
...
-                "Certificate not was not issued by intermediate CA %s", CertificateHelper.toPem(issuedClientCertificate)
-        ));
+        TestHelpers.eventuallyTrue(() -> {
+            try {
+                X509Certificate intermediateCA =  chain[0];
+                X509Certificate issuedClientCertificate = eventRef.get().getCertificate();
+                return CertificateTestHelpers.wasCertificateIssuedBy(intermediateCA, issuedClientCertificate);
+            } catch (CertificateException e) {
+                return  false;
+            }
```
jbutler, line 324: Do we want to force the supplier to catch exceptions, or can we handle that in the utility method? 

BTW, I love this. Not counting the number of callbacks is a big win

src/test/java/com/aws/greengrass/clientdevices/auth/helpers/TestHelpers.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Helpers for generic assertion tooling
+ */
+public final class TestHelpers {
+    public static final long DEFAULT_GENERIC_POLLING_TIMEOUT_MILLIS =
+            TimeUnit.SECONDS.toMillis(10);
+
+    public static final long DEFAULT_POLLING_INTERVAL_MILLIS = 500;
```
jbutler, line 21: Are these defaults high enough? What if you try to build this on a raspberry pi?

jbutler (approved):

PR #180, opened 2022-10-28
feat: integ test for switching back to managed and using root ca directly

Nelsonochoam:
**Description of changes:**
This PR adds integration tests for changing the configuration from custom ca to managed ca and also tests that custom ca configuration works when the provided ca is not the full chain but just the root ca

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCATest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCATest.java
@@ -27,7 +27,6 @@ import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
-import com.aws.greengrass.testcommons.testutilities.TestUtils;
 import com.aws.greengrass.util.GreengrassServiceClientFactory;
 import com.aws.greengrass.util.Pair;
 import org.apache.commons.lang3.ArrayUtils;
@@ -47,9 +46,9 @@ import software.amazon.awssdk.services.greengrassv2data.model.PutCertificateAuth
 
 import java.io.IOException;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.security.KeyPair;
+import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.cert.CertificateEncodingException;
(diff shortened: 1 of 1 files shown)


vaibhavmurkute (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCATest.java
```diff
@@ -169,6 +175,18 @@ private void givenCDAWithCustomCertificateAuthority(URI privateKeyUri, URI certi
         kernel.getContext().waitForPublishQueueToClear();
     }
 
+    /**
+     * Simulates a customer configuring a managed certificate authority.
+     */
+    private void givenManagedCertificateAuthorityConfiguration() throws ServiceLoadException {
```
jbutler, line 181: I think this is okay, but it's slightly misleading. To me, "given" implies a starting state. This is actually updating the state from whatever the "given" actually is.

I think what I'd prefer for us to do here is have the "givenX" preconditions established by starting nucleus with a given yaml configuration. These methods can then be used to show that the switch during a deployment works.

Thoughts?

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCATest.java
```diff
@@ -185,8 +203,36 @@ private void subscribeToCertificateUpdates(GetCertificateRequest request)
...
+        TestHelpers.eventuallyTrue(() -> {
+            try {
+                String pem = CertificateHelper.toPem(store.getCACertificate());
+                String currentCAPem = CertificateHelper.toPem(currentCA);
+                return !pem.equals(currentCAPem);
+            } catch (KeyStoreException | CertificateEncodingException | IOException e) {
+                return false;
+            }
+        });
+    }
```
jbutler, line 232: Race-y. If the CA changes before this method is called then it will fail. Previous CA needs to be retrieved before the action and then passed in

Nelsonochoam: Yeah it won't cause a race right now because of where I am using it but you are right has the potential for race 🐎

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/CustomCATest.java
Nelsonochoam: I think you are right this should be called just configureManagedCertificateAuthority I used to write tests a few years ago using a build that lazily ran the steps something like

```java
TestRunner.begin()
     .given(configureCustomCertificateAuthority("foo"))
     .when(serviceStarts())
     .then(assertThatXHappened())
     .run();
```

I won't do that know but I like that it highlights the steps and the focus of the test. I will change the name of the method though

jbutler (approved):

PR #181, opened 2022-10-31
fix: return local result when cloud call fails

vaibhavmurkute:
**Issue #, if available:** Network state heuristic is not 100% reliable. In case of false positive Network-UP case, `VerifyThingAttachedToCertificate` usecase does not consider local results when cloud requests fails.

**Description of changes:** Returns local result if `VerifyThingAttachedToCertificate` cloud request fails.

**How was this change tested:** unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
@@ -89,7 +89,7 @@ public class VerifyThingAttachedToCertificate
 
             return verifyLocally(thing, dto.getCertificateId());
         } catch (CloudServiceInteractionException e) {
-            return false;
+            return verifyLocally(thing, dto.getCertificateId());
         }
     }
 }
--- a/src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
+++ b/src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.iot.usecases;
+
+import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
+import com.aws.greengrass.clientdevices.auth.infra.NetworkState;
+import com.aws.greengrass.clientdevices.auth.iot.IotAuthClient;
+import com.aws.greengrass.clientdevices.auth.iot.Thing;
+import com.aws.greengrass.clientdevices.auth.iot.dto.VerifyThingAttachedToCertificateDTO;
+import com.aws.greengrass.clientdevices.auth.iot.infra.ThingRegistry;
+import com.aws.greengrass.testcommons.testutilities.GGExtension;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith({MockitoExtension.class, GGExtension.class})
+class VerifyThingAttachedToCertificateTest {
+    @Mock
+    private IotAuthClient mockIotAuthClient;
+    @Mock
+    private NetworkState mockNetworkState;
+    @Mock
+    private ThingRegistry mockThingRegistry;
+    private VerifyThingAttachedToCertificate verifyThingAttachedToCertificate;
+
+    @BeforeEach
+    void beforeEach() {
+        verifyThingAttachedToCertificate = new VerifyThingAttachedToCertificate(mockIotAuthClient,
+                mockThingRegistry, mockNetworkState);
+    }
+
+    @Test
+    void GIVEN_validDtoAndNetworkUp_WHEN_verifyThingAttachedToCertificate_THEN_returnCloudResult() {
+        Boolean expectedCloudResult = true;
+        Thing mockThing = Thing.of("mock-thing");
+        String mockCertId = "cert-id";
+        VerifyThingAttachedToCertificateDTO dto =
+                new VerifyThingAttachedToCertificateDTO(mockThing.getThingName(), mockCertId);
+
+        when(mockNetworkState.getConnectionStateFromMqtt()).thenReturn(NetworkState.ConnectionState.NETWORK_UP);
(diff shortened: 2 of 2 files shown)


src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
```diff
@@ -0,0 +1,93 @@
...
+        String mockCertId = "cert-id";
+        VerifyThingAttachedToCertificateDTO dto =
+                new VerifyThingAttachedToCertificateDTO(mockThing.getThingName(), mockCertId);
+
+        when(mockNetworkState.getConnectionStateFromMqtt()).thenReturn(NetworkState.ConnectionState.NETWORK_UP);
+        when(mockIotAuthClient.isThingAttachedToCertificate(any(), anyString())).thenReturn(expectedCloudResult);
+        when(mockThingRegistry.getThing(mockThing.getThingName())).thenReturn(mockThing);
+
+        assertThat(verifyThingAttachedToCertificate.apply(dto), is(expectedCloudResult));
+        verify(mockIotAuthClient, times(1)).isThingAttachedToCertificate(mockThing, mockCertId);
```
Nelsonochoam, line 59: I would remove this call count verification and just assert agains expected outputs on the public api. Having this assertion couples this test unnecessarily to the implementation

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
```diff
@@ -0,0 +1,93 @@
...
+    @Test
+    void GIVEN_validDtoAndNetworkUp_WHEN_verifyThingAttachedToCertificate_THEN_returnCloudResult() {
+        Boolean expectedCloudResult = true;
+        Thing mockThing = Thing.of("mock-thing");
+        String mockCertId = "cert-id";
+        VerifyThingAttachedToCertificateDTO dto =
+                new VerifyThingAttachedToCertificateDTO(mockThing.getThingName(), mockCertId);
+
+        when(mockNetworkState.getConnectionStateFromMqtt()).thenReturn(NetworkState.ConnectionState.NETWORK_UP);
+        when(mockIotAuthClient.isThingAttachedToCertificate(any(), anyString())).thenReturn(expectedCloudResult);
```
Nelsonochoam, line 55: Why not use the IotAuthFake?. I know this is a unit test but the less we mock as a general rule the better given our tests become less coupled to implementation while giving us confidence that things are working.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
```diff
@@ -0,0 +1,93 @@
...
+        when(mockThingRegistry.getThing(mockThing.getThingName())).thenReturn(mockThing);
```
Nelsonochoam, line 56: We could do away without mocking any of this and providing an instance of the registry

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
```diff
@@ -0,0 +1,93 @@
...
+    void beforeEach() {
+        verifyThingAttachedToCertificate = new VerifyThingAttachedToCertificate(mockIotAuthClient,
+                mockThingRegistry, mockNetworkState);
+    }
+
+    @Test
+    void GIVEN_validDtoAndNetworkUp_WHEN_verifyThingAttachedToCertificate_THEN_returnCloudResult() {
+        Boolean expectedCloudResult = true;
+        Thing mockThing = Thing.of("mock-thing");
+        String mockCertId = "cert-id";
```
Nelsonochoam, line 50: Even though this is a unit test I would suggest not using fake values that are not representative of what a real id would be. The fact that we can create the dto with just a random string is potentially dangerous

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
```diff
@@ -0,0 +1,93 @@
...
+        mockThing.attachCertificate(mockCertId);
+        VerifyThingAttachedToCertificateDTO dto =
+                new VerifyThingAttachedToCertificateDTO(mockThing.getThingName(), mockCertId);
+
+        when(mockNetworkState.getConnectionStateFromMqtt()).thenReturn(NetworkState.ConnectionState.NETWORK_DOWN);
+        when(mockThingRegistry.getThing(mockThing.getThingName())).thenReturn(mockThing);
+
+        assertThat(verifyThingAttachedToCertificate.apply(dto), is(true));
+        verify(mockIotAuthClient, times(0)).isThingAttachedToCertificate(any(), anyString());
```
Nelsonochoam, line 74: Don't verify calls to specific implementation methods

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
vaibhavmurkute: I think building an instance of the registry would couple registry implementation details to the this test.

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
vaibhavmurkute: removed, thanks

Nelsonochoam (approved):

src/test/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificateTest.java
```diff
@@ -0,0 +1,133 @@
...
+        Thing thing = Thing.of("thing-1");
+        X509Certificate certificate = createTestClientCertificate();
+        String certPem = CertificateHelper.toPem(certificate);
+        Certificate thingCertificate = Certificate.fromPem(certPem);
+        thing.attachCertificate(thingCertificate.getCertificateId());
+        VerifyThingAttachedToCertificateDTO dto =
+                new VerifyThingAttachedToCertificateDTO(thing.getThingName(), thingCertificate.getCertificateId());
+
+        // set up VerifyThingAttachedToCertificate usecase with mock iot auth client
+        // to be able to throw exceptions during cloud calls
```
jbutler, line 108: I'd rather add functionality to the fake instead of using a mock here.

The reason is that we may change the iot auth client in the future, and we will likely forget to update this test since CloudServiceInteractionException is a runtime exception.

But we can do that as a separate PR

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
```diff
@@ -89,7 +89,7 @@ public Boolean apply(VerifyThingAttachedToCertificateDTO dto) {
 
             return verifyLocally(thing, dto.getCertificateId());
         } catch (CloudServiceInteractionException e) {
-            return false;
+            return verifyLocally(thing, dto.getCertificateId());
```
jbutler, line 92: This is better, but we still need to ensure we go to the cloud if we don't have anything in the local registry, regardless of network state.

Refer to my comment in the NetworkState class. We cannot rely on it to be accurate. Right now we have no way to authenticate a client if the `isNetworkUp` never returns true

PR #183, opened 2022-11-01
feat: intro network fake

Nelsonochoam:
**Description of changes:**
This pr refactors the network state and creates and interface so we can implement a network fake to be used


src/main/java/com/aws/greengrass/clientdevices/auth/infra/NetworkStateProvider.java
```diff
@@ -0,0 +1,118 @@
...
+        NETWORK_UP,
+        NETWORK_DOWN,
+    }
+
+    void registerHandler(Consumer<ConnectionState> networkChangeHandler);
+
+    ConnectionState getConnectionState();
+
+
+    class Default implements NetworkStateProvider {
```
Nelsonochoam, line 29: This is just the network state we have always had moved as the default here

src/test/java/com/aws/greengrass/clientdevices/auth/iot/NetworkStateFake.java
```diff
@@ -0,0 +1,26 @@
...
+public class NetworkStateFake implements NetworkStateProvider {
+    private final List<Consumer<ConnectionState>> handlers = new ArrayList<>();
+
+    @Override
+    public void registerHandler(Consumer<ConnectionState> networkChangeHandler) {
+        handlers.add(networkChangeHandler);
+    }
+
+    @Override
+    public ConnectionState getConnectionState() {
```
Nelsonochoam, line 23: This is the fake I have to implement the methods but I will do it when I write the test

src/test/java/com/aws/greengrass/clientdevices/auth/iot/NetworkStateFake.java
```diff
@@ -0,0 +1,26 @@
+/*
```
jbutler, line 1: nit: move to infra?

jbutler (approved):

PR #184, opened 2022-11-01
test: certificate revocation offline/online

Nelsonochoam:
**Description of changes:**
Integration test for the following scenario

Scenario:
1. Client Device connects when GGC is online
2. GGC goes offline
3. Client device certificate is revoked in the cloud
4. Client device is still able to connect to GGC (offline)
5. GGC comes online6. Revoked client device is no longer able to connect to the GGC.

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
@@ -15,12 +15,14 @@ import com.aws.greengrass.clientdevices.auth.certificate.infra.ClientCertificate
 import com.aws.greengrass.clientdevices.auth.certificate.infra.BackgroundCertificateRefresh;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
 import com.aws.greengrass.clientdevices.auth.connectivity.CISShadowMonitor;
+import com.aws.greengrass.clientdevices.auth.exception.AuthenticationException;
 import com.aws.greengrass.clientdevices.auth.helpers.CertificateTestHelpers;
 import com.aws.greengrass.clientdevices.auth.infra.NetworkStateProvider;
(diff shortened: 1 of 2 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -276,4 +280,32 @@ void GIVEN_storedCertificates_WHEN_refreshEnabled_THEN_storedCertificatesRefresh
...
+       assertNotNull(connectToCore(thingOne.get(), clientPem));
+
+       // When
+       network.goOffline();
+       iotAuthClientFake.deactivateCert(clientPem); // Revoked
+       assertNotNull(connectToCore(thingOne.get(), clientPem));
+
+       // Then
+       network.goOnline();
+       assertThrows(AuthenticationException.class, () -> {connectToCore(thingOne.get(), clientPem);});
```
MikeDombo, line 308: this is a race, isn't it? Between the online callback and trying `connectToCore`?

Nelsonochoam: All this is synchronous for the purpose of this test, but there could be a race between the time we record the state as "online" and the time the device connects to core, causing the code to hit the cache instead of going to the cloud.

MikeDombo (dismissed):

vaibhavmurkute (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -182,54 +212,28 @@ void GIVEN_storedCertificates_WHEN_refreshEnabled_THEN_storedCertificatesRefresh
            throws Exception {
         ignoreExceptionOfType(context, java.nio.file.NoSuchFileException.class);
```
jbutler, line 213: import this?

PR #186, opened 2022-11-01
test: offline auth with expired ttl

jbutler:
**Issue #, if available:**

**Description of changes:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
@@ -278,6 +278,36 @@ public class OfflineAuthenticationTest {
    }
 
    @Test
+   void GIVEN_clientConnectsWhileOnline_WHEN_offlineAndTtlExpired_THEN_clientCanNotConnect(ExtensionContext context)
+           throws Exception {
+       ignoreExceptionOfType(context, NoSuchFileException.class);
+
+       // Given
+       network.goOnline();
+       Instant now = Instant.now();
+       mockInstant(now.toEpochMilli());
+
+       // Configure the things attached to the core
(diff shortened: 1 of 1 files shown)


MikeDombo (approved):

Nelsonochoam (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -277,6 +277,36 @@ void GIVEN_clientConnectsWhileOnline_WHEN_offline_THEN_clientCanConnect(Extensio
...
+       iotAuthClientFake.attachCertificateToThing(thingOne.get(), clientPem);
+       iotAuthClientFake.attachThingToCore(thingOne);
+       iotAuthClientFake.activateCert(clientPem);
+
+       runNucleusWithConfig("config.yaml");
+       assertNotNull(connectToCore(thingOne.get(), clientPem));
+
+       // When
+       network.goOffline();
+       Instant twoMinutesLater = now.plusSeconds(2 * 60); // Default expiry is 1 minute
```
vaibhavmurkute, line 303: This test could break if we ever update the default. We could refer to the static default value: `SecurityConfiguration.DEFAULT_CLIENT_DEVICE_TRUST_DURATION_MINUTES` instead of hard-coding.

vaibhavmurkute (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
jbutler: Yes, good call. I left a comment so hopefully it's easy to figure out in the future.

I'd recommend reading from config instead of relying on a constant in the source though

PR #188, opened 2022-11-01
test: corrupt client pem, still authenticates

Nelsonochoam:
**Description of changes:**

Test to ensure corrupting the stored client pem does not break authentication

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
@@ -390,6 +390,7 @@ public class OfflineAuthenticationTest {
         // Configure the things attached to the core
         List<X509Certificate> clientCertificates = createClientCertificates(1);
         String clientPem = CertificateHelper.toPem(clientCertificates.get(0));
+        Certificate certificate = Certificate.fromPem(clientPem);
         Supplier<String> thingOne =  () -> "ThingOne";
         iotAuthClientFake.attachCertificateToThing(thingOne.get(), clientPem);
         iotAuthClientFake.attachThingToCore(thingOne);
@@ -402,6 +403,13 @@ public class OfflineAuthenticationTest {
         corruptStoredClientCertificate(clientPem);
 
         // Then
+
+        // Assert cert pem is corrupted
+        ClientCertificateStore pemStore = kernel.getContext().get(ClientCertificateStore.class);
+        String storeCertificatePem = pemStore.getPem(certificate.getCertificateId()).get();
+        assertThrows(InvalidCertificateException.class, () -> Certificate.fromPem(storeCertificatePem));
+
+        // Assert that authenticating offline or online is not affected
         network.goOffline();
         assertNotNull(connectToCore(thingOne.get(), clientPem));
         network.goOnline();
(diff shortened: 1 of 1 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -154,6 +159,17 @@ private String connectToCore(String thingName, String certificatePem) throws Aut
...
 
+    private void corruptStoredClientCertificate(String pem) throws InvalidCertificateException, IOException {
+        NucleusPaths paths = kernel.getNucleusPaths();
+        Path workPath = paths.workPath(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME);
+        Certificate cert = Certificate.fromPem(pem);
+        Path pemFilePath = workPath.resolve("clients").resolve(cert.getCertificateId() + ".pem");
+
+        try (OutputStream writeStream = Files.newOutputStream(pemFilePath)) {
+            writeStream.write("I am evil :)".getBytes());
+        }
```
jbutler, line 170: How can we ensure that this is actually corrupted the pem? If we are making assumptions here, we should make sure that the tests will fail is those assumptions don't hold true in the future (e.g. cert store changes the storage format).

Right now the test below will continue to pass

jbutler: I think we might need a cert store fake for this

Nelsonochoam: Well we know due to the fact the test fails if I don't ignore the exception. Every time the network goes from up to down and back to up we trigger a background refresh (if we have not run one in the last 24 hours). What this test proofs ids that regardless of what happens to the pem it does not affect the authentication process.

From what we broke on the test scenario
> If Client-Certificate-Store (keystore or the file) gets corrupted (or is unreadable), it does not break normal CDA functioning. (Offline auth should still work, only refresh will be affected).

On thing we can do to improve it is to have some assertions that demonstrate the background refresh failed. We could add an assertion here as well that gets the pem from the registry and tries to build a certificate and fails to check the pem is corrupted

Nelsonochoam: Done https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/188/files#diff-7174340e1ac320caaaa268ce778e4e526258177b9409dc6128987d10d6d6905fR409

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -362,4 +379,41 @@ void GIVEN_clientConnectsWhileOnline_WHEN_offlineAndCertDetachedFromThing_THEN_b
...
+        // Assert cert pem is corrupted
+        ClientCertificateStore pemStore = kernel.getContext().get(ClientCertificateStore.class);
+        String storeCertificatePem = pemStore.getPem(certificate.getCertificateId()).get();
+        assertThrows(InvalidCertificateException.class, () -> Certificate.fromPem(storeCertificatePem));
+
+        // Assert that authenticating offline or online is not affected
+        network.goOffline();
+        assertNotNull(connectToCore(thingOne.get(), clientPem));
+        network.goOnline();
+        assertNotNull(connectToCore(thingOne.get(), clientPem));
```
MikeDombo, line 417: Do we fix the pem on disk by this point? How do we eventually recover from the bad storage? If we do recover by this point, let's verify that the pem on disk is what we want.

Nelsonochoam: Uhh that is a good one. It should get fixed. Let me add it 🧠

jbutler: I agree this is something we should have, but at this point I'm inclined to cut a backlog item to avoid scope creep. We're frozen at this point

jbutler (approved):

jbutler (approved):

vaibhavmurkute (approved):

PR #189, opened 2022-11-02
fix: use Greengrass V2 Data client with no implicit retry

vaibhavmurkute:
**Issue #, if available:** The Nucleus provided GG V2 Data Client has a built in retry policy which cannot be overridden. This results in CDA retrying network calls a number of times, which times out TLS handshakes.

**Description of changes:** IotAuthClient to use new GG V2 Data Client with no implicit retry policy.

**How was this change tested:** Existing unit, integration tests


src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
```diff
@@ -0,0 +1,136 @@
...
+                .credentialsProvider(lazyCredentialProvider)
+                .httpClient(httpClient.build())
+                .overrideConfiguration(ClientOverrideConfiguration.builder()
+                        .retryPolicy(RetryPolicy.none()).build());
+
+        if (Utils.isEmpty(awsRegion)) {
+            return clientBuilder.build();
+        }
+        clientBuilder.region(Region.of(awsRegion));
+        getGreengrassServiceEndpoint(deviceConfiguration, awsRegion)
```
MikeDombo, line 108: Data and control plane endpoints are not the same, they cannot be using the same API to get the endpoints.

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/8a8ff7f53eac7a301103969fcbbb2d567e3ccdf8/src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java#L49

This is the API that must be used for the data endpoint in order to work properly in VPC.

MikeDombo: Also need to ensure the right port is being used, since customers may configure it to use 443, rather than the default of 8443.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
```diff
@@ -0,0 +1,136 @@
...
+                .httpClient(httpClient.build())
+                .credentialsProvider(lazyCredentialProvider)
+                .overrideConfiguration(
+                        ClientOverrideConfiguration.builder().retryPolicy(RetryMode.STANDARD).build());
+
+        if (Utils.isEmpty(awsRegion)) {
+            return clientBuilder.build();
+        }
+        clientBuilder.region(Region.of(awsRegion));
+        getGreengrassServiceEndpoint(deviceConfiguration, awsRegion)
```
MikeDombo, line 79: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/96f636b99ca6db2fdcecd15b56f5057330e4ddcf/src/main/java/com/aws/greengrass/util/RegionUtils.java#L43

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
vaibhavmurkute: thank you, will fix these.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
```diff
@@ -0,0 +1,136 @@
...
+import com.aws.greengrass.util.GreengrassServiceClientFactory;
+import com.aws.greengrass.util.IotSdkClientFactory;
+import com.aws.greengrass.util.RegionUtils;
+import com.aws.greengrass.util.Utils;
+import com.aws.greengrass.util.exceptions.InvalidEnvironmentStageException;
+import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
+import software.amazon.awssdk.core.retry.RetryMode;
+import software.amazon.awssdk.core.retry.RetryPolicy;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
+import software.amazon.awssdk.regions.Region;
```
Nelsonochoam, line 23: We might want to manually test this and make sure the api calls are working as expected

vaibhavmurkute: yep, will do

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
vaibhavmurkute: fixed

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
```diff
@@ -0,0 +1,135 @@
...
+     * Provides a new GG v2 client.
+     * Its upto the caller to maintain the lifecycle of the client.
+     * (TODO: This should ideally exist in Nucleus's GreengrassServiceClientFactory)
+     *
+     * @return GreengrassV2Client
+     */
+    public GreengrassV2Client getGGV2Client()  {
+        DeviceConfiguration deviceConfiguration = getDeviceConfiguration();
+        ApacheHttpClient.Builder httpClient = ClientConfigurationUtils
+                .getConfiguredClientBuilder(deviceConfiguration);
```
MikeDombo, line 66: this method gives you a client which is configured for tls mutual auth. That is not what you want for using the control plane.

Should use `getSdkHttpClientBuilder` or `getSdkHttpClient`

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
```diff
@@ -0,0 +1,135 @@
...
+                .credentialsProvider(lazyCredentialProvider)
+                .overrideConfiguration(
+                        ClientOverrideConfiguration.builder().retryPolicy(RetryMode.STANDARD).build());
+
+        if (Utils.isEmpty(awsRegion)) {
+            return clientBuilder.build();
+        }
+        clientBuilder.region(Region.of(awsRegion));
+        getEnvStage(deviceConfiguration)
+                .ifPresent(stage -> {
```
MikeDombo, line 80: region cannot be empty and neither can environment stage. If either of these are empty, we should not be continuing, these are critical errors

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return GreengrassV2DataClient
+     */
+    public GreengrassV2DataClient getGGV2DataClient()  {
+        DeviceConfiguration deviceConfiguration = getDeviceConfiguration();
+        ApacheHttpClient.Builder httpClient = ClientConfigurationUtils
+                .getConfiguredClientBuilder(deviceConfiguration);
+
+        String awsRegion = Coerce.toString(deviceConfiguration.getAWSRegion());
+        GreengrassV2DataClientBuilder clientBuilder = GreengrassV2DataClient.builder()
+                .credentialsProvider(lazyCredentialProvider)
```
MikeDombo, line 102: data client should have an anonymous credential provider, it does not use sigv4, data client uses mutual auth.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
jbutler: We should essentially just be copy/pasting what Nucleus is doing here, right? Except skip the retry policy

vaibhavmurkute: this is for the control plane client, skipped it from this PR.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassClientFactory.java
vaibhavmurkute: fixed it, thanks.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+        clientBuilder.region(Region.of(awsRegion));
+        clientBuilder.endpointOverride(URI.create(ggServiceEndpoint));
+        return clientBuilder.build();
+    }
+
+    private String getAwsRegion(DeviceConfiguration deviceConfiguration) {
+        String awsRegion = Coerce.toString(deviceConfiguration.getAWSRegion());
+        if (Utils.isEmpty(awsRegion)) {
+            throw new RuntimeException("Failed to retrieve Greengrass service endpoint. "
```
MikeDombo, line 70: why are we throwing unchecked exceptions? Who is handling this, and what are they doing about it?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
```diff
@@ -0,0 +1,76 @@
...
+public class GreengrassV2DataClientFactory {
+    private final GreengrassServiceClientFactory greengrassServiceClientFactory;
+
+    /**
+     * Construct Greengrass Client Factory.
+     *
+     * @param greengrassServiceClientFactory GG Service Client Factory to retrieve the latest device-config
+     */
+    @Inject
+    public GreengrassV2DataClientFactory(GreengrassServiceClientFactory greengrassServiceClientFactory) {
```
MikeDombo, line 38: why are we injecting the factory? That seems pretty weird. You're only using it to get the device configuration, so why aren't we just injecting the `DeviceConfiguration`?

vaibhavmurkute: `GreengrassServiceClientFactory` subscribes to `DeviceConfiguration` updates and also validates those updates. We're injecting this factory here to avoid having to subscribe and validate device config updates.

MikeDombo: Why would you need to do that?

The original client factory will still do its work whether or not you inject it here.

MikeDombo: Let's say that the factory finds the config to be invalid. This code in this class doesn't do anything differently, it doesn't know or care about the validation because that code isn't here.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
vaibhavmurkute: Just replicating Nucleus's behavior since these are similar critical errors and CDA cannot continue with it: [InvalidEnvironmentStageException](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/8a8ff7f53eac7a301103969fcbbb2d567e3ccdf8/src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java#L54-L57), [link](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/dd8e059e802f44a763819a01e2088263ae56f3b3/src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java#L303-L307). Do you have a suggestion on what exception this should be? [InvalidConfigurationException](https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/main/java/com/aws/greengrass/clientdevices/auth/exception/InvalidConfigurationException.java) ?

MikeDombo: Yes invalid configuration would be fine. I'm not too worried about the exception itself, I just really want to make sure it is handled appropriately.

jbutler: We aren't catching runtime exceptions anywhere. Please do not throw them

vaibhavmurkute: fixed it

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
vaibhavmurkute: Makes sense, fixed it now.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -145,8 +149,8 @@ public boolean isThingAttachedToCertificate(Thing thing, String certificateId) {
                     VerifyClientDeviceIoTCertificateAssociationRequest.builder()
                             .clientDeviceThingName(thing.getThingName())
                             .clientDeviceCertificateId(certificateId).build();
-            try {
-                clientFactory.getGreengrassV2DataClient().verifyClientDeviceIoTCertificateAssociation(request);
```
MikeDombo, line 149: handle the `DeviceConfigurationException` nicely in these places. Right now it will fallthrough to the log about checking permissions, which doesn't apply to this error.

vaibhavmurkute: It falls through the catch-all block that also logs the cause (stacktrace) of this exception with a generic failure message.

vaibhavmurkute: Let me add a catch block `DeviceConfigurationException` to make the cause more clear.

MikeDombo (approved):

Nelsonochoam (approved):

PR #192, opened 2022-11-03
fix: ensure we only refresh if the network was previously down

Nelsonochoam:
**Description of changes:**
When the device boots up it will receive a network up event. That causes the background refresh to be immediately triggered. This PR ensures that it only gets triggered if the network was previously down and has not been triggered in the last 24h

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
@@ -24,6 +24,7 @@ import software.amazon.awssdk.services.greengrassv2data.model.InternalServerExce
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
 
 import java.io.IOException;
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
@@ -55,6 +56,7 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
     private final ThingRegistry thingRegistry;
     private final IotAuthClient iotAuthClient;
     private final CertificateRegistry certificateRegistry;
+    private final Clock clock;
     private ScheduledFuture<?> scheduledFuture = null;
     private final ScheduledThreadPoolExecutor scheduler;
     private final AtomicReference<Instant> nextRunDue = new AtomicReference<>();
@@ -70,12 +72,13 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
      * @param networkState        - A network state
      * @param pemStore            -  Store for the client certificates
      * @param useCases            - useCases service
+     * @param clock               - A clock instance
      */
     @Inject
     public BackgroundCertificateRefresh(ScheduledThreadPoolExecutor scheduler, ThingRegistry thingRegistry,
                                         NetworkStateProvider networkState, CertificateRegistry certificateRegistry,
                                         ClientCertificateStore pemStore, IotAuthClient iotAuthClient,
-                                        UseCases useCases) {
+                                        UseCases useCases, Clock clock) {
         this.scheduler = scheduler;
         this.thingRegistry = thingRegistry;
         this.networkState = networkState;
@@ -83,6 +86,7 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
         this.pemStore = pemStore;
         this.iotAuthClient = iotAuthClient;
         this.useCases = useCases;
+        this.clock = clock;
     }
 
     /**
@@ -111,13 +115,6 @@ public class BackgroundCertificateRefresh implements Runnable, Consumer<NetworkS
     }
 
     /**
-     * Return true if the background refresh has started.
-     */
-    public boolean isRunning() {
-        return scheduledFuture != null;
-    }
-
-    /**
      * Returns the next Instant the task is scheduled to run. This will change based on whether the device goes offline
      * or not but the guarantee is that it is scheduled to run 24h after the last successful run.
      */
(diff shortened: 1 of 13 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -171,7 +171,10 @@ private void scheduleNextRun() {
      */
     @Override
     public void accept(NetworkStateProvider.ConnectionState connectionState) {
-        if (connectionState == NetworkStateProvider.ConnectionState.NETWORK_UP) {
+        boolean shouldRun = isNetworkBackOnline();
+        this.previousConnectionState = connectionState;
+
+        if (shouldRun) {
```
jbutler, line 177: Tracking previous connection state doesn't seem right. It's an indirect mechanism which I think achieves the intended goal, but it makes the code harder to reason about and modify.

Can we just update `canRun()`?

```
    private boolean canRun() {
        // if nextScheduledRun is null then we have not started yet
        if (nextScheduledRun.get() == null) {
            return false;
        }
        Instant now = Instant.now();
        return now.equals(nextScheduledRun.get()) || now.isAfter(nextScheduledRun.get());
    }
```

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -346,8 +351,22 @@ void GIVEN_cloudFailure_WHEN_verifyingAThingCertificateAttachment_THEN_refreshTa
         assertEquals(backgroundRefresh.getLastRan(), now);
     }
 
+    @Test
+    void GIVEN_deviceBootUpAndNetworkUp_WHEN_backgroundRefreshStarts_THEN_itRunsOnlyAfterNetworkWasPreviouslyDown() {
```
jbutler, line 355: This isn't testing the behavior we want. If the network flaps right away after CDA starts, we still don't want it to refresh until the 24 hour mark, right?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
Nelsonochoam: I was taking a look at your suggestion but going with that approach it will change a few things in terms of how the refresh works (which are minor). In the scenario we start the service and there is a network up and down in the first 24hours. We wouldn't trigger the refresh, given `nextScheduledRun` wouldn't be null but the seconds condition would return false

jbutler: That's the intended behavior, no?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -123,45 +121,28 @@ public boolean isRunning() {
...
      * or not but the guarantee is that it is scheduled to run 24h after the last successful run.
      */
-    public Instant getNextScheduledRun() {
-        return nextScheduledRun.get();
-    }
-
-    /**
-     * Returns the last Instant the task ran successfully.
-     */
-    public Instant getLastRan() {
```
Nelsonochoam, line 133: We didn't really need this anymore for validating this was working fine

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -123,45 +121,28 @@ public boolean isRunning() {
...
 
     /**
      * Runs verifyIotCertificate useCase for all the registered client certificate PEMs.
      */
     @Override
     public synchronized void run() {
-        if (isNetworkDown()) {
-            logger.debug("Network is down - not refreshing certificates");
-            return;
+        if (canRun()) {
```
Nelsonochoam, line 133: Moved the `isNetworkDown` inside the `canRun`

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -287,24 +340,14 @@ void GIVEN_triggeredByMultipleThreads_WHEN_called_THEN_itShouldOnlyRunOnceForTha
         verify(iotAuthClientFake, times(1)).getThingsAssociatedWithCoreDevice();
     }
 
-    @Test
-    void GIVEN_triggeredInDifferentInstants_WHEN_triggeredExactlyOneDayLater_THEN_itShouldRunAgain() {
```
Nelsonochoam, line 291: This was already covered by other tests

src/test/java/com/aws/greengrass/clientdevices/auth/certificate/BackgroundCertificateRefreshTest.java
```diff
@@ -82,13 +81,12 @@ public class BackgroundCertificateRefreshTest {
     @Mock
     private GreengrassV2DataClient client;
     @Mock
```
Nelsonochoam, line 83: The tests below were changed to reflect the functionality we actually wanted. Trigger the refresh when there is network connectivity after at least 24 hours

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyThingAttachedToCertificate.java
```diff
@@ -33,27 +35,29 @@
...
         this.networkState = networkState;
+        this.clock = clock;
     }
 
     private boolean verifyLocally(Thing thing, String certificateId) {
-        logger.atInfo().kv("thing", thing.getThingName()).kv("certificate", certificateId)
+        logger.atDebug().kv("thing", thing.getThingName()).kv("certificate", certificateId)
                 .log("Network down, verifying thing attached to certificate locally");
-        return thing.isCertificateAttached(certificateId);
+        return thing.isCertificateAttached(certificateId, clock);
```
vaibhavmurkute, line 52: I know this makes testing easy, just something to discuss:
I feel there is a lot of flexibility with time here, with every usecase entity passing their own instance of clock. It makes sense for registries to have a clock, but should usecases (or anything using registries) track their own time (clock)?

PR #195, opened 2022-11-04
fix: change log levels

Nelsonochoam:
**Description of changes:**
Just changing some log levels here to avoid a few unnecessary chatty logs from being printed


jbutler (approved):

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java
```diff
@@ -148,6 +148,8 @@ public synchronized void run() {
             return;
         }
 
+        logger.atInfo("Running background task: Refreshing client certificates");
```
MikeDombo, line 151: this will not log anything. `at<x>` needs to be followed by `.log()`. 

Use `logger.info()` to log at info level.

MikeDombo (approved):

jbutler (approved):

PR #198, opened 2022-11-10
feat: add certificate chain config

Nelsonochoam:
**Issue #, if available:**
TPM modules do not support adding certificate chains into a single key. This renders TPM users unavailable from providing intermediate certificates.

**Description of changes:**
Introduces a new configuration called certificateChainUri. When provided we will read the certificate chain from it. We are keeping the old configuration that would behave the same if the certificateChainUri is not provided. This will allow TPM users to use the following configuration

```
    "certificateAuthority": {
      "certificateChainUri": "file:///home/ec2-user/creds/intermediateCA.pem",
      "certificateUri": "pkcs11:object=CustomerIntermediateCA;type=cert",
      "privateKeyUri": "pkcs11:object=CustomerIntermediateCA;type=private"
    }
```

Which effectively allows TPM users to provide intermediate certificate chains.

Allow TPM users to provide certificate chains

Manual tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
@@ -141,7 +141,7 @@ public final class CAConfiguration {
 
     private static Optional<URI> getCaCertificateChainUriFromConfiguration(Topics certAuthorityTopic)
             throws URISyntaxException {
-        String certificateChainUri = Coerce.toString(certAuthorityTopic.findOrDefault("", CA_CERTIFICATE_CHAIN_URI));
+        String certificateChainUri = Coerce.toString(certAuthorityTopic.find(CA_CERTIFICATE_CHAIN_URI));
 
         if (Utils.isEmpty(certificateChainUri)) {
             return Optional.empty();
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/configuration/CAConfiguration.java
```diff
@@ -132,6 +139,17 @@ private static URI getUri(String rawUri) throws URISyntaxException {
         return uri;
     }
 
+    private static Optional<URI> getCaCertificateChainUriFromConfiguration(Topics certAuthorityTopic)
+            throws URISyntaxException {
+        String certificateChainUri = Coerce.toString(certAuthorityTopic.findOrDefault("", CA_CERTIFICATE_CHAIN_URI));
```
MikeDombo, line 144: using find or default here with an empty string is rather pointless. Coerce to string will give you null if it doesn't exist, which is fine.

MikeDombo (commented):
Please make sure that this PR has a good and understandable description.

Nelsonochoam:
> Please make sure that this PR has a good and understandable description.

Yeah I was going to add it this morning

MikeDombo (dismissed):

src/test/java/com/aws/greengrass/clientdevices/auth/CertificateManagerTest.java
```diff
@@ -159,6 +191,52 @@ void GIVEN_customCAConfiguration_WHEN_configureCustomCA_THEN_returnsCustomCA() t
...
+        KeyPair intermediateKeyPair = CertificateStore.newRSAKeyPair(2048);
+        X509Certificate intermediateCA =
+                CertificateTestHelpers.createIntermediateCertificateAuthority(rootCA, "intermediate",
+                        intermediateKeyPair.getPublic(), rootKeyPair.getPrivate());
+
+
+        X509Certificate[] certificateChain = {intermediateCA, rootCA};
+        storeCAOnFileSystem(intermediateKeyPair.getPrivate(), certificateChain);
+
+        URI certificateChainUri = new File(rootDir.resolve("certificate.pem").toString()).toURI();
```
MikeDombo, line 214: Path has toFile

parrottsquawk (approved):

MikeDombo (approved):

PR #200, opened 2022-11-23
feat: adding infra for Cert.SubscribeError, Config.Invalid, and Cert.Rotate metrics

sameerzuberi:
Adding the infra for 3 extra metrics in ClientDeviceAuthMetrics: Cert.SubscribeError, Config.Invalid, and Cert.Rotate. These changes will make it possible to track and emit the metrics in the future. Each of these metrics are being added as part of phase 1 for the CDA metric ingestion project. Separate tests for each of these metrics will also be created once they are tracked.


jcosentino11 (commented):
looks like ClientDeviceAuth.log was included by accident please remove. Also can you update the PR title to be more specific?

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private ConnectivityInformation mockConnectivityInformation;
+
+    @Mock
+    private CertificateExpiryMonitor mockCertExpiryMonitor;
+
+    @Mock
+    private CISShadowMonitor mockShadowMonitor;
+
+    @Mock
+    private SecurityService securityServiceMock;
```
jcosentino11, line 67: naming nit: consistently use `***Mock` or `mock****`

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -0,0 +1,138 @@
...
+        assertEquals(3L, metrics.getSubscribeSuccess());
+
+        // Checking that the emitter collects the metrics as expected
+        Metric metric = Metric.builder()
+                .namespace("ClientDeviceAuth")
+                .name("Cert.SubscribeSuccess")
+                .unit(TelemetryUnit.Count)
+                .aggregation(TelemetryAggregation.Sum)
+                .value(3L)
+                .timestamp(Instant.now().toEpochMilli())
```
jcosentino11, line 124: since we're comparing timestamps below, there's a race between this and https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/afc52b6e0be2121fdb2620f3d412b7b3a6c64c7b/src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java#L30

you could pass in a clock to `ClientDevicesAuthMetrics` and have instants in that class and in this test method use the same clock

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -189,6 +189,7 @@ public void subscribeToCertificateUpdates(GetCertificateRequest getCertificateRe
             }
             metrics.subscribeSuccess();
         } catch (NoSuchAlgorithmException e) {
+            metrics.subscribeError();
```
jcosentino11, line 192: might be splitting hairs, but it may be good to move usage of the new ClientDeviceAuthMetrics methods to a separate PR.

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.spy;
+
+@ExtendWith({MockitoExtension.class, GGExtension.class})
+public class MetricsTest {
```
jcosentino11, line 50: similar to other comment, i think this test class can be its own PR.

jcosentino11 (commented):
Hopefully this isn't bikeshedding, but I think there's three separate changes that deserve their own PRs
* adding functionality ClientDeviceAuthMetrics
* using the new functionality 
* new MetricsTest

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CertificateRotationHandler.java
```diff
@@ -29,7 +29,6 @@ public class CertificateRotationHandler implements Consumer<CACertificateChainCh
     private final Set<CertificateGenerator> monitoredCertificateGenerators = new CopyOnWriteArraySet<>();
     private final DomainEvents domainEvents;
 
-
```
jcosentino11, line 32: nit: remove

jcosentino11 (approved):
lgtm, please also update the PR description

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -18,6 +18,10 @@
 public class ClientDeviceAuthMetrics {
 
     private final AtomicLong certSubscribeSuccess = new AtomicLong();
+    private final AtomicLong certSubscribeError = new AtomicLong();
```
jbutler, line 21: Take a look at the event emitter infrastructure that we have. We need to be using that in order to keep this code de-coupled.

Application code should emit a metric event containing the event. This code can listen for those events and handle them appropriately.

I also want to re-visit metric names here to make sure we're getting the ones we want. But, the good news is, if we code it up properly, we can make changes easily without needing to make a lot of changes.

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -49,4 +83,52 @@ public void subscribeSuccess() {
...
+    public long getSubscribeSuccess() {
+        return certSubscribeSuccess.get();
+    }
+
+    /**
+     * Increments the Cert.SubscribeError metric.
+     */
+    public void subscribeError() {
+        certSubscribeError.incrementAndGet();
+    }
```
jbutler, line 98: To elaborate more on my comment above, instead of having code that emits metrics call this directly, register this class as a listener for metric events and extract the metric from the event.

jbutler (changes requested):

PR #201, opened 2022-11-23
fix(sdk-clients): use http client builder to avoid leaks

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Using `.httpClient` will create a non-managed SDK HTTP client which means that when the SDK client is closed the HTTP client will NOT close. This is absolutely not what we want in these cases because the SDK client is used and then immediately thrown away. This in turn, causes us to leak HTTP connection managers which are kept alive by the idle connection reaper which will never remove them from tracking because they're never closed.

Using `.httpClientBuilder` instead, will create a managed HTTP client which is properly closed at the same time the SDK client is closed. We are doubtless making this same mistake in the Nucleus and in our other plugins leading to increased memory usage and ultimately leaks.

Also disabling the idle connection reaper for these because we aren't using pooled HTTP clients, so there's no reason to keep the idle connection alive.

Verified by stepping through with a debugger, that closing the SDK client will close the HTTP client when using `httpClientBuilder`. @Nelsonochoam verified in his own testing that the heap stays small when using these changes.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
@@ -53,8 +53,9 @@ public class GreengrassV2DataClientFactory {
 
         GreengrassV2DataClientBuilder clientBuilder =
                 GreengrassV2DataClient.builder().credentialsProvider(AnonymousCredentialsProvider.create())
-                        .httpClient(httpClient.build()).overrideConfiguration(
-                                ClientOverrideConfiguration.builder().retryPolicy(RetryPolicy.none()).build());
+                        .httpClientBuilder(httpClient.useIdleConnectionReaper(false))
+                        .overrideConfiguration(ClientOverrideConfiguration.builder()
+                                .retryPolicy(RetryPolicy.none()).build());
 
         clientBuilder.region(Region.of(awsRegion));
         clientBuilder.endpointOverride(URI.create(ggServiceEndpoint));
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
@@ -202,7 +202,8 @@ public interface IotAuthClient {
         private GreengrassV2Client getGGV2Client() {
             String awsRegion = Coerce.toString(deviceConfiguration.getAWSRegion());
             GreengrassV2ClientBuilder clientBuilder =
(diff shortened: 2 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
```diff
@@ -53,8 +53,9 @@ public GreengrassV2DataClient getClient() throws DeviceConfigurationException {
 
         GreengrassV2DataClientBuilder clientBuilder =
                 GreengrassV2DataClient.builder().credentialsProvider(AnonymousCredentialsProvider.create())
-                        .httpClient(httpClient.build()).overrideConfiguration(
-                                ClientOverrideConfiguration.builder().retryPolicy(RetryPolicy.none()).build());
+                        .httpClientBuilder(httpClient.useIdleConnectionReaper(false))
```
shaguptashaikh, line 56: Why is this set to false?

shaguptashaikh (commented):
Add details on how you tested this?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
MikeDombo: Explained in description. We aren't pooling our clients, so there's no reason why we should use the idle reaper. Setting this to false will also avoid the leak (in addition to using `httpClientBuilder`)

MikeDombo:
> Add details on how you tested this?

Updated description.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/GreengrassV2DataClientFactory.java
Nelsonochoam: I believe that forces the http connection to be closed when the `GreengrassV2DataClient` get closed instead of waiting for it to be closed async which I believe is what we want given if this gets called too frequently we might end up with too many open http resources

Nelsonochoam (approved):

shaguptashaikh (approved):

Nelsonochoam:
For reference I ran a few manual test

This is before this changes I was creating connections every second and over time the go up. I took heap snapshots every few minutes.

```
52M	dump.hprof
52M	total

53M	dump_1.hprof
53M	total

55M	dump_2.hprof
55M	total

62M	dump_3.hprof
62M	total
```

With this changes the heap size starts going down again

```
39M	/tmp/dump.hprof
39M	total

38M	/tmp/dump_1.hprof
38M	total

22M	/tmp/dump_2.hprof
22M	total
```

PR #202, opened 2022-12-02
fix: decoupling infra for metric tracking

sameerzuberi:
Instead of having a dedicated metrics emitter for the CDA, each metric now acts as an event which is routed to `MetricsHandler.java` and is handled according to the specific metric. This change keeps the classes decoupled, and focuses on the first metric, SubscribeToCertificateUpdates.Success. A unit test was made to verify the metric is still being collected correctly.


src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -5,70 +5,59 @@
...
-                .aggregation(TelemetryAggregation.Sum)
-                .value(invalidConfig.getAndSet(0))
-                .timestamp(timestamp)
-                .build();
-        metricsList.add(metric);
-
-        metric = Metric.builder()
-                .namespace(NAMESPACE)
-                .name("Cert.Rotate")
+                .name("SubscribeToCertificateUpdates.Success")
```
MikeDombo, line 57: since this is duplicated, let's get a public constant for it.

telemetry/ClientDeviceAuth.log
```diff
@@ -0,0 +1,5 @@
+{"thread":"main","level":"TRACE","eventType":null,"message":"{\"NS\":\"ClientDeviceAuth\",\"N\":\"SubscribeToCertificateUpdates.Success\",\"U\":\"Count\",\"A\":\"Sum\",\"V\":1,\"TS\":1669939514697}","contexts":{},"loggerName":"Metrics-ClientDeviceAuth","timestamp":1669939514738,"cause":null}
```
MikeDombo, line 1: don't commit these files.

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
sameerzuberi: I would need to inject a `ClientDeviceAuth` instance into `CertificateManager` to use the public constant. Would doing that be ok even though we're decoupling them?

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -5,70 +5,59 @@
...
     /**
      * Builds the CDA metrics.
+     *
      * @return a list of {@link Metric}
      */
     public List<Metric> collectMetrics() {
         List<Metric> metricsList = new ArrayList<>();
 
-        long timestamp = Instant.now().toEpochMilli();
+        long timestamp = clock.instant().toEpochMilli();
```
jcosentino11, line 54: nit:
```suggestion
        long timestamp = Instant.now(clock).toEpochMilli();
```

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -77,58 +69,25 @@ public List<Metric> collectMetrics() {
...
 
     /**
-     * Increments the Cert.SubscribeSuccess metric.
+     * Returns the clock.
+     *
+     * @return Clock
      */
-    public void subscribeSuccess() {
-        certSubscribeSuccess.incrementAndGet();
+    public Clock getClock() {
```
jcosentino11, line 76: not needed anymore

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -187,7 +189,7 @@ public void subscribeToCertificateUpdates(GetCertificateRequest getCertificateRe
                 };
                 subscribeToClientCertificateUpdatesNoCSR(getCertificateRequest, keyPair.getPublic(), consumer);
             }
-            metrics.subscribeSuccess();
+            domainEvent.emit(new MetricEvent(ClientDeviceAuthMetrics.METRIC_SUBSCRIBE_TO_CERTIFICATE_UPDATES_SUCCESS));
```
jbutler, line 192: This can work, but it also introduces coupling between the certificate and metric domains.

Ideally this code doesn't know anything about metrics at all. It should emit a certificate event and let an external listener translate that into a metric.

Perhaps you could add a CertificateSubscription event that includes details about the type of certificate and whether it succeeded or failed? Then we have flexibility to slice and dice the metrics however we want without needing to change this code.

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -0,0 +1,134 @@
...
+        metricsHandler.listen();
+        certificateStore = new CertificateStore(tmpPath, domainEvents, mockSecurityService);
+        certRotationMonitor = new CertificateRotationHandler(mockConnectivityInformation, domainEvents);
+
+        certificateManager = new CertificateManager(certificateStore, mockConnectivityInformation,
+                mockCertExpiryMonitor, mockShadowMonitor, Clock.systemUTC(), mockClientFactory, mockSecurityService,
+                certRotationMonitor, domainEvents);
+
+        CertificatesConfig certificatesConfig =
+                new CertificatesConfig(Topics.of(new Context(), CONFIGURATION_CONFIG_KEY, null));
```
jcosentino11, line 91: Context needs to be closed after each

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
sameerzuberi: Is the type of certificate needed? Do we want to distinguish between server/client subscriptions or is one metric for both alright? I could always have the distinction there and only increment one metric in case we want to change it in the future though.

jbutler: IMO they should be separate metrics. Initially we may only want to emit metrics for server certificates since we don't expose client cert subscriptions to customers. But in the future we may, and tracking those separately is definitely something we would want

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
jbutler: Ideally this _isn't_ duplicated. Only the the thing sending this to the telemetry emitter needs to know it

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/events/CertificateSubscriptionEvent.java
```diff
@@ -0,0 +1,19 @@
...
+import com.aws.greengrass.clientdevices.auth.api.GetCertificateRequestOptions;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+public class CertificateSubscriptionEvent implements DomainEvent {
+    @Getter
+    private GetCertificateRequestOptions.CertificateType certificateType;
+    @Getter
+    private boolean successfulSubscription;
```
MikeDombo, line 18: prefer enum over boolean. We may have additional states here, like _why_ it failed (ie. network error, permissions error).

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -24,7 +25,7 @@
 import com.aws.greengrass.clientdevices.auth.exception.CloudServiceInteractionException;
 import com.aws.greengrass.clientdevices.auth.exception.InvalidCertificateAuthorityException;
 import com.aws.greengrass.clientdevices.auth.exception.InvalidConfigurationException;
-import com.aws.greengrass.clientdevices.auth.metrics.ClientDeviceAuthMetrics;
+import com.aws.greengrass.clientdevices.auth.metrics.events.CertificateSubscriptionEvent;
```
jbutler, line 28: This is a certificate domain event - not a metric. Should be moved into `certificates/events`

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -5,70 +5,62 @@
...
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
+import javax.inject.Inject;
 
 public class ClientDeviceAuthMetrics {
+    private final AtomicLong subscribeToCertificateUpdatesSuccess = new AtomicLong();
+    private final MetricFactory mf = new MetricFactory(NAMESPACE);
+    private final Clock clock;
+    private static final String NAMESPACE = "ClientDeviceAuth";
```
jbutler, line 24: `ClientDevicesAuth` (missing "s"). Is this the namespace we're using or should it be`aws.greengrass.clientdevices.Auth` ?

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/handlers/SubscribeToCertificateUpdatesHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
+import com.aws.greengrass.clientdevices.auth.api.GetCertificateRequestOptions;
+import com.aws.greengrass.clientdevices.auth.metrics.ClientDeviceAuthMetrics;
+import com.aws.greengrass.clientdevices.auth.metrics.events.CertificateSubscriptionEvent;
+
+import java.util.function.Consumer;
+import javax.inject.Inject;
+
+public class SubscribeToCertificateUpdatesHandler implements Consumer<CertificateSubscriptionEvent> {
```
jbutler, line 16: nit: CertificateSubscriptionHandler

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/handlers/SubscribeToCertificateUpdatesHandler.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Trigger metric update.
+     *
+     * @param event Update metric event
+     */
+    @Override
+    public void accept(CertificateSubscriptionEvent event) {
+        if (event.getCertificateType().equals(GetCertificateRequestOptions.CertificateType.SERVER)
+                && event.isSuccessfulSubscription()) {
+            metrics.subscribeSuccess();
```
jbutler, line 48: Add a TODO if we're also tracking failures

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        GetCertificateRequestOptions requestOptions = new GetCertificateRequestOptions();
+        requestOptions.setCertificateType(GetCertificateRequestOptions.CertificateType.SERVER);
+        GetCertificateRequest certificateRequest =
+                new GetCertificateRequest("testService", requestOptions, con.getRight());
+
+        // Subscribe multiple times to ensure the Cert.SubscribeSuccess metric increments correctly
+        certificateManager.subscribeToCertificateUpdates(certificateRequest);
+        certificateManager.subscribeToCertificateUpdates(certificateRequest);
+        certificateManager.subscribeToCertificateUpdates(certificateRequest);
```
jbutler, line 122: Ok, everything else looks good, but this can be improved still :) 

The beautiful thing about using events is that you don't need to deal with the certificate manager at all to test it. All you need to do is emit a CertificateSubscriptionEvent. You can get rid of all of the mocks you're using above that you need to instantiate the cert manager. Right now this test is fragile. If we make changes in the certificate manager, we wouldn't want metric tests to break. 

Later we can add an integration test to make sure that the end to end flow works as expected. But that test doesn't need to be aware of CDA internals. They rely only on external interfaces, so they won't break if we change domain internals

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -0,0 +1,75 @@
...
+                .namespace("aws.greengrass.clientdevices.Auth")
+                .name(ClientDeviceAuthMetrics.METRIC_SUBSCRIBE_TO_CERTIFICATE_UPDATES_SUCCESS)
+                .unit(TelemetryUnit.Count)
+                .aggregation(TelemetryAggregation.Sum)
+                .value(3L)
+                .timestamp(Instant.now(clock).toEpochMilli())
+                .build();
+
+        List<Metric> collectedMetrics = metrics.collectMetrics();
+        Metric subscribeSuccess = collectedMetrics.get(0);
```
jbutler, line 65: This seems a bit fragile, but can be addressed separately

jbutler (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/events/CertificateSubscriptionEvent.java
```diff
@@ -0,0 +1,28 @@
...
+    private GetCertificateRequestOptions.CertificateType certificateType;
+    @Getter
+    private SubscriptionStatus status;
+
+    public enum SubscriptionStatus {
+        SUCCESS,
+        FAIL
+    }
+
+    public CertificateSubscriptionEvent(GetCertificateRequestOptions.CertificateType certificateType,
```
MikeDombo, line 23: nit: you can use Lombok to generate this using `@AllArgsConstructor`

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -5,70 +5,62 @@
...
+    @Inject
+    public ClientDeviceAuthMetrics(Clock clock) {
+        this.clock = clock;
+    }
 
-    private static final String NAMESPACE = "ClientDeviceAuth";
+    /**
+     * Emit metrics using Metric Factory.
+     */
+    public void emitMetrics() {
```
MikeDombo, line 41: Next PR this emit needs to be called on a timer like we do in the Nucleus for the builtin telemetry producers

MikeDombo (approved):

jbutler (approved):

PR #204, opened 2022-12-05
feat: emit metrics on a periodic schedule

sameerzuberi:
Created a new class which will emit metrics on a periodic schedule, currently set at once per day. Without this change metrics cannot be automatically collected.


src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -14,10 +14,12 @@
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.inject.Inject;
 
 public class ClientDeviceAuthMetrics {
+    protected ScheduledFuture<?> future;
```
MikeDombo, line 22: why is this here?

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,36 @@
...
+    private ClientDeviceAuthMetrics metrics;
+
+    /**
+     * Constructor for the MetricsEmitter.
+     *
+     * @param ses
+     * @param metrics
+     */
+    @Inject
+    public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metrics) {
```
MikeDombo, line 24: Follow the pattern that I linked you in order to read the emission rate from the configuration.

This should emit once an hour by default. It is then aggregated (not by you) once a day and sent to the cloud.

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
sameerzuberi: I added it to follow the format you sent earlier. I use it here: [https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/204/files/4bd913d6d4157163b9b5e6a4eb5352cf6cf16061#diff-d3f582a458b026f2b5e5ec7d26a2671a2908c74f9d069e7930f2e90737795843R33](url)

MikeDombo: This doesn't belong here, this future has nothing to do with this class. It should just be in the MetricsEmitter

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,38 @@
...
+        this.ses = ses;
+        this.metrics = metrics;
+    }
+
+    /**
+     * Emit metrics once an hour.
+     */
+    public void emit() {
+        future = ses.scheduleWithFixedDelay(metrics::emitMetrics, 0, PERIODIC_AGGREGATE_INTERVAL,
```
jcosentino11, line 35: we should protect against multiple backing tasks running at the same time--typically we cancel the existing task

jcosentino11: Do we want initial delay of zero? Will we have any metrics to emit then?

sameerzuberi: We were following this format: [https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/e720757c1b0920ee771a61b73a2bcb1554bcbc33/src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java#L190](url) , which is why I went with the initial delay of 0

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,47 @@
...
+        this.ses = ses;
+        this.metrics = metrics;
+    }
+
+    /**
+     * Emit metrics once an hour.
+     */
+    public void emit() {
+        // Cancel previously running task
+        synchronized (emitMetricsLock) {
```
jcosentino11, line 37: also include the `future = ses.scheduleWithFixedDelay(...` line here, otherwise we're not actually protecting from concurrency issues :)

sameerzuberi: haha don't know how I missed that, thanks

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,47 @@
...
+    public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metrics) {
+        this.ses = ses;
+        this.metrics = metrics;
+    }
+
+    /**
+     * Emit metrics once an hour.
+     */
+    public void emit() {
+        // Cancel previously running task
```
jcosentino11, line 36: nit: move this before the cancel call, otherwise it reads like this comment applies to the whole synchronized block

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,47 @@
...
+        synchronized (emitMetricsLock) {
+            if (future != null) {
+                future.cancel(false);
+            }
+
+            // Start emitting metrics with no delay
```
jcosentino11, line 42:
```suggestion
            // Start emitting metrics with no initial delay
```

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param metrics {@link ClientDeviceAuthMetrics}
+     */
+    @Inject
+    public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metrics) {
+        this.ses = ses;
+        this.metrics = metrics;
+    }
+
+    /**
+     * Emit metrics once an hour.
```
jcosentino11, line 33: nit: `once and hour` is a bit too specific and may change in the future, or be configurable.  maybe something like "Emits metrics at an interval"

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
jcosentino11: @sameerzuberi is the config stuff being addressed in this PR?

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.aws.greengrass.clientdevices.auth.metrics;
+
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+
+public class MetricsEmitter {
+    protected ScheduledFuture<?> future;
```
jcosentino11, line 14: nit: why protected? can we lower to private?

sameerzuberi: Had it in a separate file earlier but it can be private now. I'm not sure what you mean by the config stuff though

sameerzuberi: ah I misunderstood it the first time around, I'll go address the config stuff here

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,48 @@
...
+    @Inject
+    public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metrics) {
+        this.ses = ses;
+        this.metrics = metrics;
+    }
+
+    /**
+     * Emit metrics at an interval.
+     */
+    public void emit() {
```
jbutler, line 35: What calls this?

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,48 @@
...
+        synchronized (emitMetricsLock) {
+            // Cancel previously running task
+            if (future != null) {
+                future.cancel(false);
```
vaibhavmurkute, line 39: Wouldn't we want to interrupt the task if it is running (`future.cancel(true)`)?
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#cancel(boolean)

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
sameerzuberi: How could I go about emitting on a timer without the executor service? For context, I was following the format in the TelemetryAgent here: [https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/221c28bde96ee7c9aec853ed81fffeee0376ae43/src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java#L190](url)

jbutler: Ok I see that this is separate from the main metrics class. So that's good. But I actually don't think this is needed at all. It's just another layer that doesn't really provide value.

Right now we're kinda gluing pieces together in the service entry class. It's a bit of a dumping ground, but I'd honestly rather concentrate how everything gets hooked up in that one central place.

E.g. add a `initializeMetrics` startup method or something.

https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/8682bcb81b23022fe43a95f18fe67229d8acbad8/src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java#L117

Also take a look at https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/main/java/com/aws/greengrass/clientdevices/auth/certificate/infra/BackgroundCertificateRefresh.java

There are some commonalities between these two use cases. Ideally we'd actually just have a scheduler as part of infra, and can just ask it to periodically do things rather than re-creating the wheel

sameerzuberi: I'll take a look into this and adjust, thank you!

jbutler: Chatted offline. There are some complexities with the CertRefresh logic that aren't present here, so we'll keep it simple for now

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
jbutler: Your call if we can plumb it through or not. This may not be the interface we want. `start()`/`stop()` maybe?

This will probably be called from ClientDeviceAuthService, yes

MikeDombo (approved):

jcosentino11 (approved):

PR #205, opened 2022-12-06
feat: add metrics for Verify Client Device Identity API

sameerzuberi:
Added metrics to track the success/failure of the Verify Client Device Identity API. Events are emitted when a device is either successfully or unsuccessfully verified, and handled accordingly by incrementing the respective metric. Unit tests were created to ensure the metrics are correctly updated after events are emitted. 

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
@@ -54,8 +54,8 @@ public class ClientDevicesAuthServiceApi {
     public boolean verifyClientDeviceIdentity(String certificatePem) {
         // Allow internal clients to verify their identities
         if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
-            domainEvents.emit(new VerifyClientDeviceIdentityEvent
-                    (VerifyClientDeviceIdentityEvent.VerificationStatus.SUCCESS));
+            domainEvents.emit(new VerifyClientDeviceIdentityEvent(
+                    VerifyClientDeviceIdentityEvent.VerificationStatus.SUCCESS));
             return true;
         } else {
             return useCases.get(VerifyIotCertificate.class).apply(certificatePem);
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
@@ -12,7 +12,6 @@ import com.aws.greengrass.clientdevices.auth.exception.AuthenticationException;
 import com.aws.greengrass.clientdevices.auth.exception.AuthorizationException;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
 import com.aws.greengrass.clientdevices.auth.iot.usecases.VerifyIotCertificate;
-import com.aws.greengrass.clientdevices.auth.metrics.VerifyClientDeviceIdentityEvent;
 import com.aws.greengrass.clientdevices.auth.session.SessionManager;
 
 import java.util.Map;
@@ -58,8 +57,6 @@ public class ClientDevicesAuthServiceApi {
                     VerifyClientDeviceIdentityEvent.VerificationStatus.SUCCESS));
             return true;
         } else {
-            domainEvents.emit(new VerifyClientDeviceIdentityEvent(
-                    VerifyClientDeviceIdentityEvent.VerificationStatus.FAIL));
             return useCases.get(VerifyIotCertificate.class).apply(certificatePem);
         }
     }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
@@ -60,8 +60,8 @@ public class ClientDevicesAuthServiceApi {
             success = useCases.get(VerifyIotCertificate.class).apply(certificatePem);
         }
 
-        domainEvents.emit(new VerifyClientDeviceIdentityEvent(success ?
-                VerifyClientDeviceIdentityEvent.VerificationStatus.SUCCESS :
+        domainEvents.emit(new VerifyClientDeviceIdentityEvent(success
+                ? VerifyClientDeviceIdentityEvent.VerificationStatus.SUCCESS :
                 VerifyClientDeviceIdentityEvent.VerificationStatus.FAIL));
 
         return success;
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -79,6 +85,26 @@ public List<Metric> collectMetrics() {
                 .build();
         metricsList.add(metric);
 
+        metric = Metric.builder()
+                .namespace(NAMESPACE)
+                .name(METRIC_VERIFY_CLIENT_DEVICE_IDENTITY_SUCCESS)
+                .unit(TelemetryUnit.Count)
+                .aggregation(TelemetryAggregation.Sum)
+                .value(verifyClientDeviceIdentitySuccess.getAndSet(0))
```
vaibhavmurkute, line 93: super nit: `0` --> `0L`. (One less cast for the compiler!)

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -49,8 +54,12 @@ public ClientDevicesAuthServiceApi(SessionManager sessionManager, DeviceAuthClie
     public boolean verifyClientDeviceIdentity(String certificatePem) {
         // Allow internal clients to verify their identities
         if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
+            domainEvents.emit(new VerifyClientDeviceIdentityEvent(
+                    VerifyClientDeviceIdentityEvent.VerificationStatus.SUCCESS));
             return true;
         } else {
+            domainEvents.emit(new VerifyClientDeviceIdentityEvent(
```
jcosentino11, line 61: i'm not clear on why we determine success/failure based on `deviceAuthClient.isGreengrassComponent(certificatePem)`, what's the reason?  i would assume we'd want to determine success based on the overall result of this method, but i might be missing something

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/VerifyClientDeviceIdentityEvent.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.metrics;
```
jcosentino11, line 6: sounds like this should be in API domain, so `com.aws.greengrass.clientdevices.auth.api.events`?

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -49,8 +54,12 @@ public ClientDevicesAuthServiceApi(SessionManager sessionManager, DeviceAuthClie
     public boolean verifyClientDeviceIdentity(String certificatePem) {
         // Allow internal clients to verify their identities
         if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
+            domainEvents.emit(new VerifyClientDeviceIdentityEvent(
```
jcosentino11, line 57: thinking out loud, would it be worth making this a generic `ApiOperationCompleteEvent`? that way we wouldn't need a `VerifyClientDeviceIdentityEvent.VerificationStatus` for each operation?  thoughts @jbutler @sameerzuberi

sameerzuberi: I did consider doing that, the only reason I created a specific event and handler for the API is in case we decide we want to track other API specific things in the future. That way it'd be easy to just add a field in the enum for the event. But for right now since we're only focused on successes and failures it could work if we don't see ourselves making changes anytime soon

jcosentino11: yeah i agree having one for each would be more flexible---ignore my thought for now because we can always refactor later

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
jcosentino11: > and then depending on if it returns true or false

what i'm saying is that we're returning success based on `deviceAuthClient.isGreengrassComponent(certificatePem)` rather than the overall result of `verifyClientDeviceIdentity`, doesn't seem like that matches the above explanation?

sameerzuberi: Updated it to track metrics from VerifyIotCertificate.java as well

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -93,14 +98,70 @@ void GIVEN_certificateSubscriptionEvents_WHEN_eventsEmitted_THEN_subscribeFailur
...
+                .namespace("aws.greengrass.clientdevices.Auth")
+                .name(ClientDeviceAuthMetrics.METRIC_VERIFY_CLIENT_DEVICE_IDENTITY_FAILURE)
+                .unit(TelemetryUnit.Count)
+                .aggregation(TelemetryAggregation.Sum)
+                .value(3L)
+                .timestamp(Instant.now(clock).toEpochMilli())
+                .build();
+
+        List<Metric> collectedMetrics = metrics.collectMetrics();
+        Metric verifyDeviceIdentityFail = collectedMetrics.get(3);
```
jcosentino11, line 157: index is pretty fragile, can we search for the metric by name?  same question for other tests that do this

jcosentino11 (dismissed):

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -66,8 +65,12 @@ void GIVEN_certificateSubscriptionEvents_WHEN_eventsEmitted_THEN_subscribeSucces
                 .timestamp(Instant.now(clock).toEpochMilli())
                 .build();
 
-        List<Metric> collectedMetrics = metrics.collectMetrics();
-        Metric subscribeSuccess = collectedMetrics.get(0);
+        Metric subscribeSuccess = new Metric();
```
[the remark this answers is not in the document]
sameerzuberi: Interesting, wasn't familiar with this approach. Thanks! Updated the tests to follow it

jcosentino11 (approved):

vaibhavmurkute (approved):

PR #206, opened 2022-12-08
feat: add metrics for the Authorize Client Device Actions API

sameerzuberi:
Created the AuthorizeClientDeviceActions.Success and AuthorizeClientDeviceActions.Failure metrics. When a client device action is successfully or unsuccessfully authorized, an event is emitted and handled by incrementing its respective metric. Both the event and handler are specific to the Authorize Client Device Actions API. 

This change is necessary to allow for tracking the operational status of the Authorize Client Device Actions API. The change was tested by writing unit tests which ensure that the metric is correctly incremented upon events being emitted.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/AuthorizeClientDeviceActionEvent.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/AuthorizeClientDeviceActionEvent.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.api;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+public class AuthorizeClientDeviceActionEvent implements DomainEvent {
+    @Getter
+    private AuthorizationStatus status;
+
+    public enum AuthorizationStatus {
+        SUCCESS,
+        FAIL
+    }
+}
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
@@ -102,13 +102,10 @@ public class ClientDevicesAuthServiceApi {
             throws AuthorizationException {
         boolean isAuthorized = deviceAuthClient.canDevicePerform(authorizationRequest);
 
-        if (isAuthorized) {
-            domainEvents.emit(new AuthorizeClientDeviceActionEvent(AuthorizeClientDeviceActionEvent
-                    .AuthorizationStatus.SUCCESS));
-        } else {
-            domainEvents.emit(new AuthorizeClientDeviceActionEvent(AuthorizeClientDeviceActionEvent
-                    .AuthorizationStatus.FAIL));
-        }
+        domainEvents.emit(new AuthorizeClientDeviceActionEvent(isAuthorized
+                ? AuthorizeClientDeviceActionEvent.AuthorizationStatus.SUCCESS :
+                AuthorizeClientDeviceActionEvent.AuthorizationStatus.FAIL));
+
         return isAuthorized;
     }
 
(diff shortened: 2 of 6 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -100,7 +100,16 @@ public void closeClientDeviceAuthSession(String authToken) {
      */
     public boolean authorizeClientDeviceAction(AuthorizationRequest authorizationRequest)
             throws AuthorizationException {
-        return deviceAuthClient.canDevicePerform(authorizationRequest);
+        boolean isAuthorized = deviceAuthClient.canDevicePerform(authorizationRequest);
+
+        if (isAuthorized) {
+            domainEvents.emit(new AuthorizeClientDeviceActionEvent(AuthorizeClientDeviceActionEvent
```
jcosentino11, line 106: nit: for consistency w/ other one

```

        domainEvents.emit(new AuthorizeClientDeviceActionEvent(isAuthorized
                ? AuthorizeClientDeviceActionEvent.AuthorizationStatus.SUCCESS :
                AuthorizeClientDeviceActionEvent.AuthorizationStatus.FAIL));

File: src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java [hunk 2]
```diff
@@ -63,49 +69,35 @@ public void emitMetrics() {
...
-                .namespace(NAMESPACE)
-                .name(METRIC_VERIFY_CLIENT_DEVICE_IDENTITY_SUCCESS)
-                .unit(TelemetryUnit.Count)
-                .aggregation(TelemetryAggregation.Sum)
-                .value(verifyClientDeviceIdentitySuccess.getAndSet(0L))
-                .timestamp(timestamp)
-                .build();
-        metricsList.add(metric);
+    private Metric buildMetric(String metricName, AtomicLong metricValue) {
+        long timestamp = Instant.now(clock).toEpochMilli();
```
jcosentino11, line 89: there's a subtle change here, instead of all metrics having the same timestamp, they may be different now.  would advise reverting to old behavior, will be consistent w/ other emitters (KernelMetricsEmitter/SystemMetricsEmitter)

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/api/AuthorizeClientDeviceActionEvent.java
```diff
@@ -0,0 +1,20 @@
...
+
+@AllArgsConstructor
+public class AuthorizeClientDeviceActionEvent implements DomainEvent {
+    @Getter
+    private AuthorizationStatus status;
+
+    public enum AuthorizationStatus {
+        SUCCESS,
+        FAIL
+    }
```
jbutler, line 19: Just got an idea from this.

If we added more details to this event then we could add a listener to this which logs authZ events. This would normally be extremely noisy most of the time. It would be nice to control this via CDA config instead of regular log levels (enabling nucleus debug logs == RIP)

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -33,6 +35,10 @@ public class ClientDeviceAuthMetrics {
             "VerifyClientDeviceIdentity.Success";
     public static final String METRIC_VERIFY_CLIENT_DEVICE_IDENTITY_FAILURE =
             "VerifyClientDeviceIdentity.Failure";
+    public static final String METRIC_AUTHORIZE_CLIENT_DEVICE_ACTIONS_SUCCESS =
```
jbutler, line 38: Can we make these private? That would be a good litmus test to see if we actually have the encapsulation that I currently think we do

jbutler (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
sameerzuberi: I have them as public so we can access them in the tests as well, but that's the only other place they're used

jbutler: One downside of re-using string constants in tests, is that you can't test that the string itself is what you want. If someone accidentally fat fingered this string, then we'd start emitting a different metric but the tests would still fail. Something to consider.

For things that are only used in tests, consider making them package private

sameerzuberi: Interesting, hadn't thought of that. Would it be better to have them private and write out the string in the tests to ensure they're what we're looking for instead of having them package private?

PR #207, opened 2022-12-09
feat: add metrics for the Get Client Device Auth Token API

sameerzuberi:
Created the GetClientDeviceAuthToken.Success and GetClientDeviceAuthToken.Failure metrics. When an auth token is successfully or unsuccessfully retrieved, an event is emitted and handled by incrementing its respective metric. Both the event and handler are specific to the Get Client Device Auth Token API.

This change is necessary to allow for tracking the operational status of the Get Client Device Auth Token API. The change was tested by writing unit tests which ensure that the metric is correctly incremented upon events being emitted.


src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -130,6 +136,26 @@ public List<Metric> collectMetrics() {
...
+        metric = Metric.builder()
+                .namespace(NAMESPACE)
+                .name(METRIC_GET_CLIENT_DEVICE_AUTH_TOKEN_FAILURE)
+                .unit(TelemetryUnit.Count)
+                .aggregation(TelemetryAggregation.Sum)
+                .value(getClientDeviceAuthTokenFailure.getAndSet(0L))
+                .timestamp(timestamp)
+                .build();
+        metricsList.add(metric);
+
```
jbutler, line 158: Do we want to remove metrics without counts?

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
```diff
@@ -97,8 +102,12 @@ public GetClientDeviceAuthTokenResponse handleRequest(GetClientDeviceAuthTokenRe
                 String sessionId =
                         clientDevicesAuthServiceApi.getClientDeviceAuthToken(MQTT_CREDENTIAL_TYPE, credentialMap);
                 GetClientDeviceAuthTokenResponse response = new GetClientDeviceAuthTokenResponse();
+                domainEvents.emit(new GetClientDeviceAuthTokenEvent(GetClientDeviceAuthTokenEvent
```
jbutler, line 105: Following the logic from above, we'd then want to emit this at a lower layer. Moquette actually isn't using IPC. It's calling the service API directly. Ideally this should be coming from the session domain

src/main/java/com/aws/greengrass/clientdevices/auth/api/GetClientDeviceAuthTokenEvent.java
```diff
@@ -0,0 +1,20 @@
+/*
```
jbutler, line 1: This should probably belong in the session domain. Follow the directory structure we're using for the other domain events (take the term quite literally -- a domain event is an event that a software domain emits)

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/handlers/GetClientDeviceAuthTokenMetricHandler.java
```diff
@@ -0,0 +1,46 @@
+/*
```
jbutler, line 1: Similarly, this isn't listening for `GetClientDeviceAuthTokenEvents`, it's listening to session creation events

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
sameerzuberi: I can check if a metric has a positive value before building it to begin with

jbutler:
I'm not concerned with performance. Using a stream and filter below is
probably easier to read. However you want to do it

src/main/java/com/aws/greengrass/ipc/GetClientDeviceAuthTokenOperationHandler.java
jbutler: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/8682bcb81b23022fe43a95f18fe67229d8acbad8/src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionManager.java#L66

You could catch it, emit the metric, then rethrow

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/ClientDeviceAuthMetrics.java
```diff
@@ -130,6 +137,30 @@ public List<Metric> collectMetrics() {
...
+        metricsList = metricsList.stream()
```
jcosentino11, line 160: lets add a test for this

src/main/java/com/aws/greengrass/clientdevices/auth/session/handlers/SessionCreationHandler.java
```diff
@@ -0,0 +1,46 @@
...
+package com.aws.greengrass.clientdevices.auth.session.handlers;
+
+import com.aws.greengrass.clientdevices.auth.api.DomainEvents;
+import com.aws.greengrass.clientdevices.auth.metrics.ClientDeviceAuthMetrics;
+import com.aws.greengrass.clientdevices.auth.session.events.SessionCreationEvent;
+
+import java.util.function.Consumer;
+import javax.inject.Inject;
+
+public class SessionCreationHandler implements Consumer<SessionCreationEvent> {
```
vaibhavmurkute, line 15: Consider renaming this handler to reflect that its an _event handler_ and not Session-Creation handler (we actually have another class for that `SessionCreator`). 

Same goes for other _event_ handlers:
- `VerifyClientDeviceIdentityHandler`
- `CertificateSubscriptionHandler`

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -11,6 +11,7 @@
 import com.aws.greengrass.clientdevices.auth.exception.AuthenticationException;
 import com.aws.greengrass.clientdevices.auth.exception.AuthorizationException;
 import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
+import com.aws.greengrass.clientdevices.auth.iot.events.VerifyClientDeviceIdentityEvent;
```
jbutler, line 14: ?? shouldn't this cause a pmd violation? is it being used?

src/main/java/com/aws/greengrass/clientdevices/auth/certificate/handlers/CertificateSubscriptionEventHandler.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package com.aws.greengrass.clientdevices.auth.metrics.handlers;
+package com.aws.greengrass.clientdevices.auth.certificate.handlers;
```
jbutler, line 6: Sorry I think I confused you. The handler should still be part of the metrics domain. This is the thing that knows about metrics.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/handlers/VerifyClientDeviceIdentityEventHandler.java
```diff
@@ -3,16 +3,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package com.aws.greengrass.clientdevices.auth.metrics.handlers;
+package com.aws.greengrass.clientdevices.auth.iot.handlers;
```
jbutler, line 6: Same thing. Can keep this in metrics

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
sameerzuberi: I'm using it down in `verifyClientDeviceIdentity()` to emit the event

jbutler (approved):

jcosentino11 (approved):

PR #208, opened 2022-12-12
feat: adding service error metric

sameerzuberi:
Created the ServiceError metric. This metric is not API specific, but tracks the number of unhandled exceptions across the CDA, and logs the event with an error message. Currently the metric is tracking runtime exceptions encountered during the execution of the APIs. These exceptions are caught, an event is emitted to increment the metric, and then the error is logged to help track these service errors. The change was tested by writing a unit test which ensures that the metric is correctly incremented upon a `ServiceErrorEvent` being emitted.


src/main/java/com/aws/greengrass/clientdevices/auth/metrics/handlers/ServiceErrorEventHandler.java
```diff
@@ -0,0 +1,46 @@
...
+     * Listen for service error events.
+     */
+    public void listen() {
+        domainEvents.registerListener(this, ServiceErrorEvent.class);
+    }
+
+    @Override
+    public void accept(ServiceErrorEvent event) {
+        metrics.incrementServiceError();
+        logger.atError().log(event.getErrorMessage());
```
jcosentino11, line 44: erroring code already throws exceptions or (should) log, so this wouldn't be providing new information, i think we can skip this log

sameerzuberi: An issue I noticed is that if I don't want to log anything in the handler, then the `ServiceErrorEvent` class wouldn't need any fields since it's only used to call `metrics.incrementServiceError()`. And at that point I don't think we need an event or handler since it would make more sense to just call the increment function directly, so would it be better to use that approach or stay consistent with the other metrics?

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -375,6 +376,7 @@ public void uploadCoreDeviceCAs(String thingName, X509Certificate certificate)
             // by the next catch block
             throw e;
         } catch (Exception e) {
+            domainEvent.emit(new ServiceErrorEvent());
```
MikeDombo, line 379: this isn't a service error. This is most likely a network or permissions error.

We need to distinguish between these types of expected and unexpected errors. For example, if the exception is an SDKException, then it is not a service error.

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -51,21 +51,26 @@ public ClientDevicesAuthServiceApi(SessionManager sessionManager, DeviceAuthClie
...
+                success = true;
+            } else {
+                success = useCases.get(VerifyIotCertificate.class).apply(certificatePem);
+            }
+
+            domainEvents.emit(new VerifyClientDeviceIdentityEvent(success ? VerifyClientDeviceIdentityEvent
+                    .VerificationStatus.SUCCESS : VerifyClientDeviceIdentityEvent.VerificationStatus.FAIL));
+
+            return success;
+        } catch (Exception e) {
```
MikeDombo, line 70: We don't want to catch only exception. In fact, runtime exceptions are the most important to catch as a service error because they aren't otherwise handled.

sameerzuberi: Do we want to handle runtime exceptions separately then? I thought that catching exception would also catch runtime exceptions since it inherits from exception

sameerzuberi: I have another question branching off of that, in the IPC handlers a new ServiceError is thrown after they catch an exception, which is why I considered any exception as one. Is that ServiceError different from what we're tracking?

MikeDombo: Yes, because we're not only tracking IPC invocations, but any invocation including from Moquette.

sameerzuberi: Ah ok, thank you

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -165,6 +165,7 @@ public String getCaPassPhrase() {
      * @param getCertificateRequest get certificate request
      * @throws CertificateGenerationException if unable to generate certificate
      */
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
```
MikeDombo, line 168: is this needed here? There are no other changes in this file.

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -51,21 +51,26 @@ public ClientDevicesAuthServiceApi(SessionManager sessionManager, DeviceAuthClie
...
+                success = useCases.get(VerifyIotCertificate.class).apply(certificatePem);
+            }
+
+            domainEvents.emit(new VerifyClientDeviceIdentityEvent(success ? VerifyClientDeviceIdentityEvent
+                    .VerificationStatus.SUCCESS : VerifyClientDeviceIdentityEvent.VerificationStatus.FAIL));
+
+            return success;
+        } catch (RuntimeException e) {
+            domainEvents.emit(new ServiceErrorEvent());
+            throw e;
```
MikeDombo, line 72: is this being logged anywhere? We're going to have a very hard time investigating errors if they're never being logged.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
sameerzuberi: My mistake, I was trying something and forgot to remove it

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
sameerzuberi: It's logged in the IPC handler, but since those aren't called in every instance another approach I had was to log in the event handler for each service error and remove the other logs to avoid duplicates

sameerzuberi: I had an approach earlier to log each service error event in the handler with the exception and an error message. Removed it after this conversation: [https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/208#discussion_r1046070557](url)

I think now it would be good to add it back since the existing logs don't apply to runtime exceptions

jcosentino11: it's being logged at the ipc level, but forgot that this could be called directly--i take back my original comment :)

sameerzuberi: Yup I'd forgotten that too. Changed it to just log when the exception happens instead of having to go through the handler this time

src/main/java/com/aws/greengrass/clientdevices/auth/api/ClientDevicesAuthServiceApi.java
```diff
@@ -52,22 +55,28 @@ public ClientDevicesAuthServiceApi(SessionManager sessionManager, DeviceAuthClie
...
     public boolean verifyClientDeviceIdentity(String certificatePem) {
-        boolean success;
-
-        // Allow internal clients to verify their identities
-        if (deviceAuthClient.isGreengrassComponent(certificatePem)) {
-            success = true;
-        } else {
-            success = useCases.get(VerifyCertificateValidityPeriod.class).apply(certificatePem) && useCases.get(
+        try {
+            boolean success;
```
jbutler, line 61: nit: I know this was the existing variable name, but how about `isVerified`?

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionManager.java
```diff
@@ -63,6 +64,7 @@ public Session findSession(String sessionId) {
      * @return session id
      * @throws AuthenticationException if device credentials were not able to be validated
      */
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
```
jbutler, line 67: Is there a reason you added this here instead of in the API class with the other service errors? I don't have a super strong preference, but it may make things easier to find if we co-locate

sameerzuberi: It was the first one I did so I ended up going a layer deeper, but I agree it'd be better to have it in the API class. I'll make that change

jcosentino11 (approved):
LGTM, assuming https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/208#discussion_r1047447614 is being addressed in a follow-up

sameerzuberi:
> LGTM, assuming [#208 (comment)](https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/208#discussion_r1047447614) is being addressed in a follow-up

I don't think we're looking to track the network/permissions errors yet in phase 1 since they don't fall under the service error umbrella

jbutler (approved):

PR #209, opened 2022-12-14
feat: add a new section for metrics to the cda config

sameerzuberi:
This PR adds a new Metrics Configuration to the CDA config. Customers will now have the option to enable or disable metrics, and to set the frequency at which metrics are emitted. By default, metrics will be emitted once an hour. These changes were tested by creating unit tests which ensure that the metrics configuration is read correctly and returns the expected values.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
@@ -55,8 +55,8 @@ public final class MetricsConfiguration {
      * @return true if changed, else false
      */
     public boolean hasChanged(MetricsConfiguration config) {
-        return config.getEnableMetrics() != getEnableMetrics() ||
-                config.getEmittingFrequency() != getEmittingFrequency();
+        return config.getEnableMetrics() != getEnableMetrics()
+                || config.getEmittingFrequency() != getEmittingFrequency();
     }
 
     private static Optional<Boolean> getEnableMetricsFlagFromConfiguration(Topics metricsTopic) {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
@@ -44,11 +44,11 @@ public class MetricsConfiguration {
     public static MetricsConfiguration from(Topics configurationTopics) {
         Topics metricsTopic = configurationTopics.lookupTopics(METRICS_TOPIC);
 
-        return new MetricsConfiguration(getEmitMetricsFlagFromConfiguration(metricsTopic),
+        return new MetricsConfiguration(getEnableMetricsFlagFromConfiguration(metricsTopic),
                 getEmittingFrequencyFromConfiguration(metricsTopic));
     }
 
-    private static Optional<Boolean> getEmitMetricsFlagFromConfiguration(Topics metricsTopic) {
+    private static Optional<Boolean> getEnableMetricsFlagFromConfiguration(Topics metricsTopic) {
         boolean emit = Coerce.toBoolean(metricsTopic.find(ENABLE_METRICS));
 
         if (emit) {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.configuration;
+
+import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.util.Coerce;
+import lombok.Getter;
+
+import java.util.Optional;
+
+/**
+ * Represents the metrics part of the component configuration. Acts as an adapter from the GG topics to the domain.
+ * <p>
+ * |---- configuration
+ * |    |---- metrics:
+ * |          |---- emitMetrics
+ * |          |---- emittingFrequency
+ * </p>
+ */
+
+public class MetricsConfiguration {
+    public static final String METRICS_TOPIC = "metrics";
+    public static final String ENABLE_METRICS = "enableMetrics";
+    public static final String EMITTING_FREQUENCY = "emittingFrequency";
+    @Getter
+    private Optional<Boolean> enableMetrics;
+    @Getter
+    private Optional<Integer> emittingFrequency;
+
+    private MetricsConfiguration(Optional<Boolean> enableMetrics, Optional<Integer> emittingFrequency) {
+        this.enableMetrics = enableMetrics;
+        this.emittingFrequency = emittingFrequency;
+    }
+
+    /**
+     * Create a MetricsConfiguration from the service configuration.
+     *
+     * @param configurationTopics the configuration key of the service configuration
+     * @return MetricsConfiguration
+     */
+    public static MetricsConfiguration from(Topics configurationTopics) {
+        Topics metricsTopic = configurationTopics.lookupTopics(METRICS_TOPIC);
+
+        return new MetricsConfiguration(getEmitMetricsFlagFromConfiguration(metricsTopic),
+                getEmittingFrequencyFromConfiguration(metricsTopic));
+    }
+
+    private static Optional<Boolean> getEmitMetricsFlagFromConfiguration(Topics metricsTopic) {
+        boolean emit = Coerce.toBoolean(metricsTopic.find(ENABLE_METRICS));
+
+        if (emit) {
+            return Optional.of(true);
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<Integer> getEmittingFrequencyFromConfiguration(Topics metricsTopic) {
+        int frequency = Coerce.toInt(metricsTopic.find(EMITTING_FREQUENCY));
+
+        if (frequency > 0) {
+            return Optional.of(frequency);
+        } else {
+            return Optional.empty();
+        }
+    }
+}
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
@@ -9,30 +9,29 @@ import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.util.Coerce;
 import lombok.Getter;
 
-import java.util.Optional;
-
 /**
  * Represents the metrics part of the component configuration. Acts as an adapter from the GG topics to the domain.
  * <p>
  * |---- configuration
  * |    |---- metrics:
- * |          |---- emitMetrics
- * |          |---- emittingFrequency
+ * |          |---- disableMetrics
+ * |          |---- aggregatePeriod
  * </p>
  */
 
 public final class MetricsConfiguration {
     public static final String METRICS_TOPIC = "metrics";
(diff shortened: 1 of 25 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @param config Metric Configuration
+     * @return true if changed, else false
+     */
+    public boolean hasChanged(MetricsConfiguration config) {
+        return config.getEnableMetrics() != getEnableMetrics()
+                || config.getEmittingFrequency() != getEmittingFrequency();
+    }
+
+    private static Optional<Boolean> getEnableMetricsFlagFromConfiguration(Topics metricsTopic) {
```
jbutler, line 62: This is not an appropriate use of Optional. Default to true. Keep that business logic contained within this class, otherwise callers need to decide how to interpret getting `Optional.empty()`

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+        boolean emit = Coerce.toBoolean(metricsTopic.find(ENABLE_METRICS));
+
+        if (emit) {
+            return Optional.of(true);
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<Integer> getEmittingFrequencyFromConfiguration(Topics metricsTopic) {
```
jbutler, line 72: Same comment here

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -15,40 +17,63 @@ public class MetricsEmitter {
     private final ScheduledExecutorService ses;
     private final Object emitMetricsLock = new Object();
     private static final int DEFAULT_PERIODIC_AGGREGATE_INTERVAL_SEC = 3_600;
+    private int periodicAggregateIntervalSec;
+    private boolean metricsEnabled = true;
```
jbutler, line 21: This should all come from the config class

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Optional;
+
+/**
+ * Represents the metrics part of the component configuration. Acts as an adapter from the GG topics to the domain.
+ * <p>
+ * |---- configuration
+ * |    |---- metrics:
+ * |          |---- emitMetrics
+ * |          |---- emittingFrequency
```
jbutler, line 20: The convention that we (Greengrass) have adopted for boolean values is that the "default" value should be false (think, the "zero-value" for a boolean variable). In this case, that would make it `disableMetrics`.

`period` is probably a more appropriate term than `frequency` in this context. `aggregationPeriod` perhaps?

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
jbutler: More context on `Optional`, from one of the Java language designers: https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

In this case, you could argue that there is "no result" and that this is therefore appropriate. But configuration needs to have a default value, and it makes sense to encapsulate that here.

sameerzuberi: Makes sense, my line of thinking was that customers have the option to provide a config for metrics so in case they didn't we'd just set it as empty, but simply setting the default as false is easier to follow

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
```diff
@@ -55,27 +54,21 @@ public static MetricsConfiguration from(Topics configurationTopics) {
...
     }
 
-    private static Optional<Integer> getEmittingFrequencyFromConfiguration(Topics metricsTopic) {
-        int frequency = Coerce.toInt(metricsTopic.find(EMITTING_FREQUENCY));
+    private static int getAggregatePeriodFromConfiguration(Topics metricsTopic) {
+        int aggregatePeriod = Coerce.toInt(metricsTopic.find(AGGREGATE_PERIOD));
 
-        if (frequency > 0) {
-            return Optional.of(frequency);
+        if (aggregatePeriod == 0) {
```
jbutler, line 68: What if it's negative?

src/test/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfigurationTest.java
```diff
@@ -39,32 +40,32 @@ void afterEach() throws IOException {
...
-    public void GIVEN_cdaConfiguration_WHEN_metricsEnabledAndFrequencyProvided_THEN_getEmittingFrequencyReturnsInt() {
+    public void GIVEN_cdaConfiguration_WHEN_metricsEnabledAndFrequencyProvided_THEN_getAggregatePeriodReturnsInt() {
         MetricsConfiguration metricsConfiguration = MetricsConfiguration.from(configurationTopics);
-        assertFalse(metricsConfiguration.getEmittingFrequency().isPresent());
-        configurationTopics.lookup(METRICS_TOPIC, EMITTING_FREQUENCY).withValue(10_000);
+        assertEquals(DEFAULT_PERIODIC_AGGREGATE_INTERVAL_SEC, metricsConfiguration.getAggregatePeriod());
+        configurationTopics.lookup(METRICS_TOPIC, AGGREGATE_PERIOD).withValue(10_000);
         metricsConfiguration = MetricsConfiguration.from(configurationTopics);
-        assertEquals(10_000, metricsConfiguration.getEmittingFrequency().get());
+        assertEquals(10_000, metricsConfiguration.getAggregatePeriod());
```
jbutler, line 69: Possibly add a quick test for bounds checks

jbutler (approved):
Overall LGTM. Couple comments about bounds

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Compares 2 Metric Configurations and returns true if there's been a change.
+     *
+     * @param config Metric Configuration
+     * @return true if changed, else false
+     */
+    public boolean hasChanged(MetricsConfiguration config) {
+        return config.disableMetrics != disableMetrics
```
jcosentino11, line 57: nit: use getters here

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/MetricsConfiguration.java
jcosentino11: +1 we don't want to break on this

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -30,15 +29,31 @@ public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metr
...
-     * Start emitting metrics with no initial delay.
+     * Restart the metrics emitter and apply configurations.
+     *
+     * @param disableMetrics               Enable metrics flag
+     * @param periodicAggregateIntervalSec Periodic aggregate interval in seconds
      */
-    public void start() {
+    public void restart(boolean disableMetrics, int periodicAggregateIntervalSec) {
+        if (disableMetrics) {
+            return;
```
jcosentino11, line 39: i think we're never stopping metrics emitter if we disable it in config?

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -30,15 +29,31 @@ public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metr
...
+        }
+
+        stop();
+        start(periodicAggregateIntervalSec);
+    }
+
+    /**
+     * Start emitting metrics with no initial delay unless specified otherwise by the Metrics Configuration.
```
jcosentino11, line 47: > unless specified otherwise by the Metrics Configuration.

this isn't true though? initial delay is always zero

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/handlers/MetricsConfigurationChangedHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Restart the Metrics Emitter to apply new configuration.
+     *
+     * @param event Metric configuration changed event
+     */
+    @Override
+    public void accept(MetricsConfigurationChanged event) {
+        metricsEmitter.restart(event.getConfiguration().isDisableMetrics(),
```
jcosentino11, line 45: related to other comment about stopping, maybe it's clearer to do something like

```
if (disableMetrics) {
  emitter.stop()
} else {
  emitter.restart()
}
```

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
sameerzuberi: Forgot to edit the comment, but the otherwise part was related to starting it, not the initial delay. Will go back and fix it

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
sameerzuberi: Oh that's true, I need to stop it before returning

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -29,16 +28,27 @@ public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metr
         this.metrics = metrics;
     }
 
+    /**
+     * Restart the metrics emitter and apply configurations.
+     *
+     * @param periodicAggregateIntervalSec Periodic aggregate interval in seconds
+     */
+    public void restart(int periodicAggregateIntervalSec) {
```
jcosentino11, line 36: nit: this is the same as start, so we can remove this---just add a comment in start that mentions it'll kill previously running tasks.

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
sameerzuberi: updated it to just use start()

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -29,22 +29,7 @@ public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metr
...
-            return;
-        }
-
-        stop();
-        start(periodicAggregateIntervalSec);
-    }
-
-    /**
-     * Start emitting metrics with no initial delay unless specified otherwise by the Metrics Configuration.
+     * Kill previous task and start emitting metrics with no initial delay.
```
jbutler, line 32: nit: fyi amazon style guidelines want you to avoid words like "kill"

jbutler (approved):

PR #210, opened 2022-12-16
fix: reject expired client certs

jbutler:
**Issue #, if available:**

**Description of changes:**
Adds a new step to the VerifyClientDeviceIdentity workflow which ensures that the certificate is not expired.

Without this change, it is possible for us to continue to trust certificate even after they are expired, if the TTL for that metadata expires after the certificate.

Unit / integ tests. Ensured tests fail prior to this change.

This matches IoT Core behavior which will reject expired client certs, regardless of their registration state.


src/test/java/com/aws/greengrass/clientdevices/auth/helpers/CertificateTestHelpers.java
```diff
@@ -176,7 +176,8 @@ private static BigInteger getSerialNumber() {
 
     private static Pair<Date, Date> getValidityDateRange() {
         Instant now = Instant.now();
-        Date notBefore = Date.from(now);
+        // TODO: caller should pass a clock or date range in instead
```
jbutler, line 179: Will clean this up a bit more in the next PR

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyCertificateValidityPeriod.java
```diff
@@ -0,0 +1,60 @@
...
+        try {
+            cf = CertificateFactory.getInstance("X.509");
+        } catch (CertificateException e) {
+            logger.atError().cause(e).log("Unable to create X.509 certificate factory");
+            return false;
+        }
+
+        try (InputStream is = new ByteArrayInputStream(certificatePem.getBytes(StandardCharsets.UTF_8))) {
+            X509Certificate cert = (X509Certificate) cf.generateCertificate(is);
+            Date now = Date.from(clock.instant());
```
MikeDombo, line 53: is this handling timezone properly?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyCertificateValidityPeriod.java
```diff
@@ -0,0 +1,60 @@
...
+            return !cert.getNotBefore().after(now) && cert.getNotAfter().after(now);
```
MikeDombo, line 54: log when the cert is expired to make it clear why we rejected the cert?

MikeDombo: How about just using `checkValidity(Date)`? https://docs.oracle.com/javase/8/docs/api/javax/security/cert/X509Certificate.html#checkValidity-java.util.Date-

jbutler: Yes, that's much better. Will use that

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyCertificateValidityPeriod.java
jbutler: Are you asking if the time in the cert has TZ information encoded? It should

MikeDombo: Cert rfc specifies GMT. The clock you use has a timezone which is typically UTC but isn't guaranteed to be UTC. Instant loses the timezone info, as does Date I believe. I don't think that there's actually a problem here, just pointing out that Java Date is unfortunately a bad type because timezone handling isn't clear. But you have no choice because Date is all that the x509 Java type works with.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyCertificateValidityPeriod.java
```diff
@@ -51,10 +54,18 @@ public Boolean apply(String certificatePem) {
         try (InputStream is = new ByteArrayInputStream(certificatePem.getBytes(StandardCharsets.UTF_8))) {
             X509Certificate cert = (X509Certificate) cf.generateCertificate(is);
             Date now = Date.from(clock.instant());
-            return !cert.getNotBefore().after(now) && cert.getNotAfter().after(now);
+            try {
+                cert.checkValidity(now);
+                isValid = true;
```
MikeDombo, line 59: Just return true in here and return false at the very end of the method?

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyCertificateValidityPeriod.java
jbutler: https://docs.oracle.com/javase/8/docs/api/java/util/Date.html
https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html

I'm not getting the impression that timezone info is "lost". Date appears to use UTC, and Instant uses some special java time that seems to be, for this use case, "good enough".

MikeDombo (approved):

jcosentino11 (approved):

PR #212, opened 2023-01-04
feat: publish metrics with ipc

sameerzuberi:
This PR changes how the CDA publishes metrics. We now use the IPC to publish in order to avoid coupling in the metrics domain. The MetricsEmitter will aggregate metrics once every hour (unless the customer configures otherwise), and will send a request to the Nucleus' `PutComponentMetric` IPC operation. If the CDA is unauthorized, then the emitter will stop and will not continue to send requests to the IPC.

In order to publish using the IPC, the minimum Nucleus version had to be bumped to 2.8. The metrics are now built using the metrics class in the AWS SDK, and then converted to the telemetry metric class when being published. 

An integration test was created testing that the MetricsEmitter correctly schedules tasks and that a telemetry message is received after metrics are published. 


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -246,6 +247,8 @@ public void postInject() {
                         cloudCallThreadPool));
         greengrassCoreIPCService.setAuthorizeClientDeviceActionHandler(
                 context -> new AuthorizeClientDeviceActionOperationHandler(context, serviceApi, authorizationHandler));
+        greengrassCoreIPCService.setPutComponentMetricHandler(
```
MikeDombo, line 250: why are you doing this? `putComponentMetric` isn't implemented in CDA, it is implemented in the Nucleus.

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -37,8 +43,10 @@ public void start(int periodicAggregateIntervalSec) {
         synchronized (emitMetricsLock) {
             // Cancel previously running task
             stop();
-            future = ses.scheduleWithFixedDelay(metrics::emitMetrics, 0, periodicAggregateIntervalSec,
-                    TimeUnit.SECONDS);
+            PutComponentMetricRequest request = new PutComponentMetricRequest();
+            request.setMetrics(metrics.collectMetrics());
+            future = ses.scheduleWithFixedDelay((Runnable) putComponentMetricOperationHandler.handleRequest(request),
```
MikeDombo, line 48: this is just going to push the same metrics over and over again. You need to collect the metrics and create a new request object every time this scheduled task executes.

src/main/java/com/aws/greengrass/ipc/PutComponentMetricOperationHandler.java
```diff
@@ -0,0 +1,151 @@
+/*
```
MikeDombo, line 1: this file must not be here. This already exists in the Nucleus and we need to use the code that is in the Nucleus.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
sameerzuberi: I talked to Nikkhil last week and he mentioned I should write it without the authorization portion of `putComponentMetric`, so I made the class without those bits

MikeDombo: This is unworkable. We need a change in the Nucleus to support our use case. Copying the implementation is not the right thing to do.

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -43,9 +51,7 @@ public void start(int periodicAggregateIntervalSec) {
         synchronized (emitMetricsLock) {
             // Cancel previously running task
             stop();
-            PutComponentMetricRequest request = new PutComponentMetricRequest();
-            request.setMetrics(metrics.collectMetrics());
-            future = ses.scheduleWithFixedDelay((Runnable) putComponentMetricOperationHandler.handleRequest(request),
+            future = ses.scheduleWithFixedDelay((Runnable) emitMetrics(),
```
MikeDombo, line 55: you can't just cast `emitMetrics()` to a runnable. What is this actually doing?

It is executing `emitMetrics()` immediately, then casting the response from `emitMetrics` which is `PutComponentMetricResponse` to a runnable, which will not succeed.

What you want is `this::emitMetrics`

src/integrationtests/resources/com/aws/greengrass/integrationtests/ipc/metricsConfig.yaml
```diff
@@ -0,0 +1,26 @@
...
+        definitions:
+          devices:
+            selectionRule: "*"
+            policyName: "allowPublishMetrics"
+        policies:
+          allowPublishMetrics:
+            policyStatement1:
+              statementDescription: "publish metrics"
+              operations: "*"
+              resources: "*"
```
MikeDombo, line 23: this isn't correct.

You're confusing CDA policy with IPC's access controls. 
https://docs.aws.amazon.com/greengrass/v2/developerguide/interprocess-communication.html#ipc-authorization-policies

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -5,27 +5,45 @@
...
+     * @param ses                     {@link ScheduledExecutorService}
+     * @param metrics                 {@link ClientDeviceAuthMetrics}
+     * @param componentMetricIPCAgent Component Metric IPC Event Stream Agent
+     * @param context                 Operation continuation handler
      */
     @Inject
-    public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metrics) {
+    public MetricsEmitter(ScheduledExecutorService ses, ClientDeviceAuthMetrics metrics,
+                          ComponentMetricIPCEventStreamAgent componentMetricIPCAgent,
+                          OperationContinuationHandlerContext context) {
```
MikeDombo, line 42: you cannot inject an `OperationContinuationHandler`. You must make one for _every_ request.

src/main/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsEmitter.java
```diff
@@ -52,4 +70,30 @@ public void stop() {
...
         }
     }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private PutComponentMetricResponse emitMetrics() {
+        if (unauthorized) {
+            //Avoid trying to repeatedly publish when unauthorized
+            stop();
+            return new PutComponentMetricResponse();
+        }
```
MikeDombo, line 80: when we last talked we said not to do this. Don't do any special checking, just handle the authorization error gracefully.

PR #213, opened 2023-01-10
feat: add integration test for metrics emitter

sameerzuberi:
This PR is adding integration tests to check the functionality of the `MetricsEmitter`. The tests  ensure that metrics are properly emitted during CDA flows. There are tests for the success metrics of each CDA API. Each test reuses logic from their respective IPC integ tests, with the addition of checking that the correct metric was emitted after the process. A new `metricsConfig` was created which specifies the aggregate interval to 1 second instead of the default one hour to simplify testing.


src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
```diff
@@ -367,4 +367,11 @@ void GIVEN_serviceErrorEvent_WHEN_eventsEmitted_THEN_serviceErrorMetricCorrectly
         assertEquals(metric.getUnit(), serviceError.getUnit());
         assertEquals(metric.getNamespace(), serviceError.getNamespace());
     }
+
+    @Test
+    void placeholder() {
```
MikeDombo, line 372: placeholder? still needed?

If needed, give it a better name.

src/integrationtests/resources/com/aws/greengrass/integrationtests/metrics/metricsConfig.yaml
```diff
@@ -0,0 +1,26 @@
...
+            policyName: "allowPublishMetrics"
+        policies:
+          allowPublishMetrics:
+            policyStatement1:
+              statementDescription: "publish metrics"
+              operations: "*"
+              resources: "*"
+      metrics:
+        disableMetrics: "false"
+        aggregatePeriod: "2"
```
MikeDombo, line 26: what is the unit of the aggregation period? The unit should be in the name: `aggregationPeriodSeconds`

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+            try {
+                MetricsPayload mp = new ObjectMapper().readValue(request.getPayload(), MetricsPayload.class);
+                assertEquals(QualityOfService.AT_LEAST_ONCE, request.getQos());
+                assertEquals("2022-06-30", mp.getSchema());
+                telemetryMessageVerified = true;
+                break;
+            } catch (IOException e) {
+                fail("The message received at this topic is not of MetricsPayload type.", e);
+            }
```
MikeDombo, line 202: this isn't useful. Just allow the test to throw the exception, you don't need try-catch.

src/test/java/com/aws/greengrass/clientdevices/auth/metrics/MetricsTest.java
sameerzuberi: Ah, nope, not needed - I was just using it to try some things out and forgot to remove it

src/integrationtests/resources/com/aws/greengrass/integrationtests/metrics/metricsConfig.yaml
sameerzuberi: I'll update this in the config

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,207 @@
...
+        assertThat(authServiceRunning.await(TEST_TIME_OUT_SEC, TimeUnit.SECONDS), is(true));
+    }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private void mockInstant(long expected) {
+        this.clockMock.ifPresent(ScopedMock::close);
+        Clock spyClock = spy(Clock.class);
+        MockedStatic<Clock> clockMock;
+        clockMock = mockStatic(Clock.class);
+        clockMock.when(Clock::systemUTC).thenReturn(spyClock);
```
MikeDombo, line 129: why are you mocking a static?

Clock should be injected via Context so that it is very easy to replace if you want to replace it.

Use `kernel.getContext().put(mockClock, Clock.class)` before calling `kernel.launch`.

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,207 @@
...
+        long delay = kernel.getContext().get(MetricsEmitter.class).getFuture().getDelay(TimeUnit.SECONDS);
+        assertEquals(delay, periodicAggregateIntervalSec);
+
+        //move clock one day ahead to trigger publish
+        Instant aDayLater = Instant.now().plusSeconds(DEFAULT_PUBLISH_INTERVAL);
+        mockInstant(aDayLater.toEpochMilli());
+
+        //check for telemetry logs in ~root/telemetry
+        assertEquals(kernel.getNucleusPaths().rootPath().resolve("telemetry"),
+                TelemetryConfig.getTelemetryDirectory());
```
MikeDombo, line 171: this isn't useful. You aren't actually checking that any telemetry exists. Anyway, it is an implementation detail which you should not be validating. 

In this test you should just validate that the metrics you emit from CDA are emitting in an aggregate form over MQTT.

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,207 @@
...
+            return CompletableFuture.completedFuture(0);
+        });
+        assertTrue(metricsPublished.await(30, TimeUnit.SECONDS), "Metrics not published");
+
+        for (PublishRequest request : requests) {
+            if (!telemetryPublishTopic.equals(request.getTopic())) {
+                continue;
+            }
+            try {
+                MetricsPayload mp = new ObjectMapper().readValue(request.getPayload(), MetricsPayload.class);
```
MikeDombo, line 195: you need to validate the actual metrics. It is good that you're validating _something_ was published, but you need to make sure it contains all the metrics which you tried to push (and that they're aggregated).

sameerzuberi: Just to double check, will the mqtt publish message have the same structure as the messages in the telemetry log in here: [https://github.com/aws-greengrass/aws-greengrass-nucleus/tree/main/src/main/java/com/aws/greengrass/telemetry](url)

MikeDombo: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java#L262. It pushes using the metric payload type.

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,207 @@
...
+    private Optional<MockedStatic<Clock>> clockMock;
+    private Kernel kernel;
+    private DomainEvents domainEvents;
+    @Mock
+    private MqttClient mqttClient;
+    @TempDir
+    Path rootDir;
+
+    @BeforeEach
+    void beforeEach() {
```
MikeDombo, line 87:
```
// Set this property for kernel to scan its own classpath to find plugins
System.setProperty("aws.greengrass.scanSelfClasspath", "true");
```

https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/cd64f652effad0de98f88dd2204ddc9e4e59e6f0/src/test/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthServiceTest.java#L123

src/integrationtests/resources/com/aws/greengrass/integrationtests/metrics/metricsConfig.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - BrokerWithAuthorizeClientDeviceActionPermission
+  BrokerSubscribingToCertUpdates:
+    dependencies:
+      - aws.greengrass.clientdevices.Auth
+    lifecycle:
+      run:
+        windows:
+          powershell -command sleep 1
+        posix:
+          sleep 1
```
MikeDombo, line 20: no lifecycle should be needed

src/integrationtests/resources/com/aws/greengrass/integrationtests/metrics/metricsConfig.yaml
```diff
@@ -0,0 +1,104 @@
...
+              - 'aws.greengrass#AuthorizeClientDeviceAction'
+            resources:
+              - '*'
+  aws.greengrass.Nucleus:
+    configuration:
+      runWithDefault:
+        posixUser: nobody
+        windowsUser: integ-tester
+      logging:
+        level: "DEBUG"
```
MikeDombo, line 79: do you need debug logs? if not,  just let it be info.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -126,7 +125,6 @@ private void initializeInfrastructure() {
         networkState.registerHandler(context.get(CISShadowMonitor.class));
         networkState.registerHandler(context.get(BackgroundCertificateRefresh.class));
         context.get(BackgroundCertificateRefresh.class).start();
-        context.get(MetricsEmitter.class).start(MetricsConfiguration.DEFAULT_PERIODIC_AGGREGATE_INTERVAL_SEC);
```
MikeDombo, line 129: how is the emitter being started now?

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,192 @@
...
+        expectedMetrics.add(buildMetric(METRIC_VERIFY_CLIENT_DEVICE_IDENTITY_SUCCESS));
+        expectedMetrics.add(buildMetric(METRIC_GET_CLIENT_DEVICE_AUTH_TOKEN_SUCCESS));
+        expectedMetrics.add(buildMetric(METRIC_SERVICE_ERROR));
+
+        //Capture the emitted metrics when the emitter runs
+        ResultCaptor<List<Metric>> resultMetrics = new ResultCaptor<>();
+        doAnswer(resultMetrics).when(metricSpy).collectMetrics();
+
+        //Wait for emitter to run
+        Thread.sleep(2000);
```
MikeDombo, line 179: avoid sleeping, instead, trigger off of an event if possible

src/integrationtests/resources/com/aws/greengrass/integrationtests/metrics/metricsConfig.yaml
```diff
@@ -0,0 +1,104 @@
...
+            resources:
+              - '*'
+        aws.greengrass.ipc.pubsub:
+          policyId2:
+            policyDescription: access to pubsub topics for ServiceName
+            operations:
+              - '*'
+            resources:
+              - '*'
+  BrokerWithGetClientDeviceAuthTokenPermission:
```
MikeDombo, line 37: what are all these components for? They don't appear to be used at all. Your test is manually emitting metrics.

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,192 @@
...
+                .name(name)
+                .unit(TelemetryUnit.Count)
+                .aggregation(TelemetryAggregation.Sum)
+                .value(1L)
+                .timestamp(Instant.now(clock).toEpochMilli())
+                .build();
+    }
+
+    @Test
+    void GIVEN_kernelRunningWithMetricsConfig_WHEN_launched_THEN_metricsCorrectlyEmittedAtAggregationInterval()
```
MikeDombo, line 150: this doesn't really seem to be an integration test. You're manually emitting the events, these events aren't being emitting through the normal code path which a customer would execute.

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
sameerzuberi: When auth service subscribes to config changes it reads in the metrics config and starts based off of what's specified in the config. Starting it here was actually redundant since it would restart after the subscribeToConfigChanges anyways

src/integrationtests/resources/com/aws/greengrass/integrationtests/metrics/metricsConfig.yaml
sameerzuberi: These are for tests that trigger the different APIs as well, I left them out of the PR right now to keep it short, but I'll add those tests in next

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
sameerzuberi: My goal with this test was to verify that the emitter schedules correctly and runs as expected, but I can move it to the unit tests if needed

MikeDombo: Just testing scheduling seems to not provide value. It sounds like you do have a plan for additional testing, let's get that in.

sameerzuberi: I went ahead and added in the tests for SubscribeToCertificateUpdates, VerifyClientDeviceIdentity, and AuthorizeClientDeviceAction. Still working on the GetClientDeviceAuthToken test

jcosentino11 (commented):
let's also get rid of that `.log` file, we don't want to commit that

src/integrationtests/java/com/aws/greengrass/integrationtests/metrics/MetricsEmitterTest.java
```diff
@@ -0,0 +1,381 @@
...
+                ipcClient.authorizeClientDeviceAction(request, Optional.empty());
+        AuthorizeClientDeviceActionResponse response = handler.getResponse().get(1, TimeUnit.SECONDS);
+        consumer.accept(response);
+    }
+
+    @Test
+    void GIVEN_kernelRunningWithMetricsConfig_WHEN_launched_THEN_metricsCorrectlyEmittedAtAggregationInterval()
+            throws InterruptedException {
+        startNucleusWithConfig("metricsConfig.yaml");
+        assertNotNull(kernel.getContext().get(MetricsEmitter.class).getFuture(),
```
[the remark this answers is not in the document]
sameerzuberi: That makes sense, removed it

MikeDombo (approved):

jcosentino11 (dismissed):

MikeDombo (approved):

jcosentino11 (approved):

PR #216, opened 2023-01-19
feat: handle IPC requests in CDA to deliver client certs

robcmann:
**Issue #, if available:**

**Description of changes:**


src/main/java/com/aws/greengrass/ipc/SubscribeToCertificateUpdatesOperationHandler.java
```diff
@@ -197,6 +199,11 @@ public void afterHandleRequest() {
 
     @Override
     protected void onStreamClosed() {
-        certificateManager.unsubscribeFromCertificateUpdates(getCertificateRequest);
+        if (getCertificateRequest == null) {
+            logger.atInfo().log("Certificate already released");
```
jbutler, line 203: nit: I'm glad you added the null check here, but the log probably isn't useful. We'll already see in the event stream logs if it gets closed immediately. you can remove this

jbutler (approved):

sameerzuberi (approved):

PR #223, opened 2023-03-02
Ggmq 11 implemenet connect

bgklika:
**Issue #, if available:**
GGMQ-11, GGMQ-12, GGMQ-19

**Description of changes:**

- Update protocol to pass connection timeout
- Extend gRPC part to handle shutdownAgent/createMqttConnection/closeMqttConnection requests
- Added MqttLibImpl and MqttConnectionImpl and implement methods required for connect and disconnect
- Update README with for build and run commands
- Tune pom.xml
- Use newMqttBuilder without TLS settings
- Switch to log4j2

MQTT client should be able to connect MQTT broker and disconnect from it

At the moment manually, later will covered by unit tests

### DIFF
--- a/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
+++ b/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
@@ -5,11 +5,22 @@
 
 package com.aws.greengrass.testing.mqtt5.client.grpc;
 
-import com.aws.greengrass.testing.mqtt.client.MqttAgentDiscoveryGrpc;
+import com.aws.greengrass.testing.mqtt.client.MqttClientControlGrpc;
+import com.aws.greengrass.testing.mqtt.client.MqttCloseRequest;
+import com.aws.greengrass.testing.mqtt.client.MqttConnectRequest;
+import com.aws.greengrass.testing.mqtt.client.MqttConnectionId;
+import com.aws.greengrass.testing.mqtt.client.MqttProtoVersion;
+import com.aws.greengrass.testing.mqtt.client.ShutdownRequest;
+import com.aws.greengrass.testing.mqtt.client.TLSSettings;
+import com.aws.greengrass.testing.mqtt5.client.MqttConnection;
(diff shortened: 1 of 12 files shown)


uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
```diff
@@ -18,26 +29,216 @@
...
-        return host + ":" + port;
+    class MqttClientControlImpl extends MqttClientControlGrpc.MqttClientControlImplBase {
+
+        @Override
+        public void shutdownAgent(ShutdownRequest request, StreamObserver<Empty> responseObserver) {
+            // save reason
+            shutdownReason = request.getReason();
+
+            // log an event
+            logger.log(Level.INFO, "shutdownAgent: {0}", new Object[]{shutdownReason});
```
MikeDombo, line 67: OTF uses log4j2, let's use that.

bgklika: I suggess that code will not a part of OTF because it actually standalone (but controlled) MQTT client. We going to have at least 5 clients. Because clients can be written on C, C++, Python etc no visible reasons to integrate with OTF. That also allows to use clients in other test scenarios for example to test backend. I will check log4j2 usage in OTF and in auth plugin too.

MikeDombo: Our plugins use logback, but OTF does use log4j2, which I'd recommend for our usage here. The builtin java logger isn't very good.

MikeDombo:

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
bgklika: switched to log4j2 in 74bafa1ae9b89fcdc50ac6a2007308157046f2d2

MikeDombo (approved):

PR #224, opened 2023-03-02
fix: store hostAddresses properly

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Fixes how hostAddresses runtime config is stored in topics


MikeDombo (commented):
How do we know this works? Your testing in this repo somehow didn't find the issue. Please make sure we have a test here which can positively identify the issue.

MikeDombo (approved):

MikeDombo:
Without your change, what does the tlog tell us from the integ test?
This doesn't seem to explain why the UAT tlog had no host addresses in it, it would only explain an odd format.

robcmann (approved):

jcosentino11:
> Without your change, what does the tlog tell us from the integ test? This doesn't seem to explain why the UAT tlog had no host addresses in it, it would only explain an odd format.

From our offline convo:  UAT failed because it was using unwanted version of CDA that was manually created

PR #225, opened 2023-03-02
fix: prevent hostAddresses from being cleared

jcosentino11:
**Issue #, if available:**

**Description of changes:**

If CIS call fails for some non-retryable exception (ValidationException/ResourceNotFoundException), the hostAddresses cache will be cleared, this change prevents that from happening

unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -241,9 +241,13 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
             }
 
             if (!connectivityInfo.isPresent()) {
+                // CIS call failed for either ValidationException or ResourceNotFoundException.
+                // We won't retry in this case, but we will update the CIS shadow reported state
+                // to signal that we have fully processed this version.
                 try {
                     updateCISShadowReportedState(desiredState);
                 } finally {
+                    // Don't process the same version again
                     lastVersion = version;
                 }
                 return;
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -26,10 +26,9 @@ import software.amazon.awssdk.iot.iotshadow.model.ShadowDeltaUpdatedEvent;
 import software.amazon.awssdk.iot.iotshadow.model.ShadowDeltaUpdatedSubscriptionRequest;
 import software.amazon.awssdk.iot.iotshadow.model.ShadowState;
 import software.amazon.awssdk.iot.iotshadow.model.UpdateShadowRequest;
+import software.amazon.awssdk.services.greengrassv2data.model.ConnectivityInfo;
 import software.amazon.awssdk.services.greengrassv2data.model.InternalServerException;
-import software.amazon.awssdk.services.greengrassv2data.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
-import software.amazon.awssdk.services.greengrassv2data.model.ValidationException;
 
 import java.time.Duration;
 import java.util.Arrays;
@@ -37,6 +36,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -226,6 +228,14 @@ private synchronized void processCISShadow(String version, Map<String, Object> d
                         .log("Retry workflow for getting connectivity info interrupted");
                 Thread.currentThread().interrupt();
                 return;
+            } catch (ValidationException | ResourceNotFoundException e) {
```
MikeDombo, line 231: why is this changing? Can't we just change: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/2eb6d62da250522efd48a18aa7b67a0ed1b757a8/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/ConnectivityInformation.java#L96 to only cache when successful?

jcosentino11: trying to prevent unnecessary rotation if `ValidationException/ResourceNotFoundException` is thrown by CIS.  if that doesn't bubble up, then CISShadowMonitor will see empty list and rotate the certs

MikeDombo: updating the reported state here seems unusual. Why is it the right thing to do?

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/ConnectivityInformation.java
```diff
@@ -66,22 +64,20 @@ public List<String> getCachedHostAddresses() {
...
         List<ConnectivityInfo> connectivityInfoList = Collections.emptyList();
 
-        try {
-            GetConnectivityInfoResponse getConnectivityInfoResponse =
-                    clientFactory.getGreengrassV2DataClient().getConnectivityInfo(getConnectivityInfoRequest);
-            if (getConnectivityInfoResponse.hasConnectivityInfo()) {
-                // Filter out port and metadata since it is not needed
-                connectivityInfoList = getConnectivityInfoResponse.connectivityInfo();
-            }
-        } catch (ValidationException | ResourceNotFoundException e) {
```
MikeDombo, line 83: and why are we now removing error handling from here?

jcosentino11: reason is on other comment, moved the handling to CISShadowMonitor

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: it doesn't have an effect on processing, i think it's more of a way to tell that the shadow was processed/dealt with just by looking at the shadow contents itself?   so for debugging purposes, and to keep with the convention that CISShadowMontior updates the shadow when it's finishing processing a shadow.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/ConnectivityInformation.java
jcosentino11: maybe it'd be better to make the return type an optional, so the error handling stays here

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -226,6 +228,14 @@ private synchronized void processCISShadow(String version, Map<String, Object> d
...
+                LOGGER.atWarn().cause(e).log("Connectivity info doesn't exist");
+                try {
+                    updateCISShadowReportedState(desiredState);
```
MikeDombo, line 234: this is actually publishing to the cloud. Why would we publish anything if our request was invalid or the shadow wasn't found.

jcosentino11: yep, was considering those situations to be non-retyable, so updating as a means of signaling that we are done processing that particular shadow version

just my initial thinking, not hard committed to this if you think otherwise

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/ConnectivityInformation.java
jcosentino11: adjusted in latest commit

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
MikeDombo: add a comment in the code to explain this.

jcosentino11: added

MikeDombo (approved):

robcmann (approved):

robcmann (approved):

PR #226, opened 2023-03-02
feat: skip cert rotation if connectivity info hasn't changed

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Before rotating certificates, check that CIS info has actually changed.

To prevent unnecessary rotations

Unit test

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -31,13 +31,16 @@ import software.amazon.awssdk.services.greengrassv2data.model.InternalServerExce
 import software.amazon.awssdk.services.greengrassv2data.model.ThrottlingException;
 
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -245,6 +248,24 @@ private synchronized void processCISShadow(String version, Map<String, Object> d
...
+                } finally {
+                    // Don't process the same version again
+                    lastVersion = version;
+                }
+                return;
+            }
+
+            // skip cert rotation if connectivity info hasn't changed
+            List<String> cachedHostAddresses = connectivityInformation.getCachedHostAddresses();
+            if (Objects.equals(prevCachedHostAddresses, cachedHostAddresses)) {
```
MikeDombo, line 263: convert to sets before comparing, otherwise duplicates or changed order will mess things up

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/ConnectivityInformation.java
```diff
@@ -55,7 +58,7 @@ public ConnectivityInformation(DeviceConfiguration deviceConfiguration,
     }
 
     /**
-     * Get cached connectivity info.
+     * Get cached connectivity info. Items in this list are unique.
```
MikeDombo, line 61: how is this ensured?

jcosentino11: ConnectivityInfoCache returns a set, so the conversions in CISShadowMonitor would be extra.  Using set in CISShadowMonitor would show clearer intent,  wanted to have a small change w/o refactoring every usage to be set

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -255,8 +276,7 @@ private synchronized void processCISShadow(String version, Map<String, Object> d
 
             try {
                 for (CertificateGenerator cg : monitoredCertificateGenerators) {
-                    cg.generateCertificate(connectivityInformation::getCachedHostAddresses,
-                            "connectivity info was updated");
+                    cg.generateCertificate(() -> cachedHostAddresses, "connectivity info was updated");
```
MikeDombo, line 279: why is this changing? calling the method should have the same effect right?

jcosentino11: should have the same effect, just being defensive if a future change ends up messing w/ cache concurrently or something

jcosentino11: esp since generateCertificate can take a while

MikeDombo (approved):

MikeDombo (changes requested):
Do convert to sets before comparing.

MikeDombo (approved):

sameerzuberi (approved):

PR #227, opened 2023-03-03
fix: synchronize CIS shadow processing

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Adds synchronization when processing CIS shadows, as we may receive concurrent requests for processing (e.g. getting shadow deltas).

The approach is as follows:
* "queue" incoming requests using an atomic reference
* we replace the queued item based on most recent shadow version, as we can ignore outdated information
* a separate thread polls the queue and submits a task for processing the shadow

❗ This change also adds in a retry mechanism in the case of certificate generation failures, or cloud call fails.

The code for thread waiting is heavily inspired from shadow manager's RequestBlockingQueue.  This approach is a lot of code, but it avoids busy-waiting entirely.

Shadow processing is not synchronized.  There's a chance that connectivity info could be different between concurrent shadow processing tasks, and certificate generation could interleave, causing certs to be generated with inconsistent connectivity information.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -47,10 +47,11 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -195,35 +217,98 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
+            this.shadowVersion = shadowVersion;
+            this.desiredState = desiredState;
+        }
+
+        @Override
+        @SuppressWarnings("PMD.AvoidCatchingGenericException")
+        public void run() {
+            try {
+                processCISShadow();
+            } catch (InterruptedException ignored) {
```
MikeDombo, line 300: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

PR #232, opened 2023-03-08
feat(uat): implement passing received MQTT message to gRPC server

bgklika:
**Issue #, if available:**
Implementation of  forwardning information from received MQTT message to gRPC server.

**Description of changes:**

- gRPC protocol has been updated to receive message by control side
- Implement passing received message from MQTT connection layer to gRPC client via BiConsumer
- Updated README
- Added GRPCClient and MqttReceivedMessage interfaces to decouple classes.
- Arguments of publish now collected in Message object
- Tune error messages in GRPCControlServer
- Tune code by adding javadoc, replace deprecated calls

Controlled MQTT client should be able to deliver information from receive MQTT message to contol side.

At the moment manually, later will cover code by unit tests.

### DIFF
--- a/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
+++ b/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
@@ -14,8 +14,17 @@ import lombok.Data;
 @Data
 @Builder
 public class MqttReceivedMessage {
+    /** QoS value. */
     int qos;
+
+    /** Retain flag. */
     boolean retain;
+
+    /** Topic of message. */
     String topic;
+
+    /** Payload of message. */
     byte[] payload;
+
+    // TODO: add user's properties and so one
 }
--- a/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
+++ b/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.testing.mqtt5.client;
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * Contains information about received MQTT v5.0 message.
+ */
+@Data
+@Builder
+public class MqttReceivedMessage {
+    int qos;
+    boolean retain;
+    String topic;
+    byte[] payload;
+}
--- a/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
+++ b/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
@@ -297,7 +297,7 @@ class GRPCControlServer {
             if (qos < QOS_MIN || qos > QOS_MAX) {
                 logger.atWarn().log("invalid QoS {}, must be in range [{},{}]", qos, QOS_MIN, QOS_MAX);
                 responseObserver.onError(Status.INVALID_ARGUMENT
-                                            .withDescription("invalie QoS, must be in range [0,3]")
+                                            .withDescription("invalid QoS, must be in range [0,3]")
                                             .asRuntimeException());
                 return;
             }
--- a/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
+++ b/uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
@@ -83,6 +83,12 @@ class GRPCControlServer {
      */
     class MqttClientControlImpl extends MqttClientControlGrpc.MqttClientControlImplBase {
 
+        /**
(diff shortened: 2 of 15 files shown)


uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
```diff
@@ -0,0 +1,30 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * Contains information about received MQTT v5.0 message.
+ */
+@Data
+@Builder
+public class MqttReceivedMessage {
```
MikeDombo, line 16: should these classes be generated by protobuf so that they're the same in every language?

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
```diff
@@ -131,26 +144,27 @@ public void createMqttConnection(MqttConnectRequest request,
 
             MqttProtoVersion version = request.getProtocolVersion();
             if (version != MqttProtoVersion.MQTT_PROTOCOL_V50) {
-                logger.atWarn().log("MQTT_PROTOCOL_V50 is only supported but {} requested", version);
+                logger.atWarn().log("invalid protocolVersion {}, {} is only supported",
+                                        version, MqttProtoVersion.MQTT_PROTOCOL_V50);
                 responseObserver.onError(Status.INVALID_ARGUMENT
-                                .withDescription("invalid protocolVersion, only MQTT_PROTOCOL_V50 is supported")
+                                .withDescription("invalid protocolVersion, only MQTT_PROTOCOL_V50 supported")
```
MikeDombo, line 150: we will need to be able to use mqtt 3 as well in the near future.

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
```diff
@@ -262,33 +282,38 @@ public void closeMqttConnection(MqttCloseRequest request, StreamObserver<Empty>
...
-            logger.atDebug().log("publishMqtt");
+            Mqtt5Message message = request.getMsg();
 
             int qos = message.getQosValue();
             if (qos < QOS_MIN || qos > QOS_MAX) {
-                logger.atWarn().log("qos is invalid can be in range [0,3] but is {}", qos);
+                logger.atWarn().log("invalid QoS {}, must be in range [{},{}]", qos, QOS_MIN, QOS_MAX);
                 responseObserver.onError(Status.INVALID_ARGUMENT
-                                            .withDescription("qos is invalid can be in range [0,3]")
+                                            .withDescription("invalie QoS, must be in range [0,3]")
```
MikeDombo, line 300: invalid

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
bgklika: I decoupled gRPC and MQTT parts by using intermediate objects. It allows replacing or adding another MQTT part much more simply.

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
bgklika: Its is possible but was not evaluated for phases 1 and 2. I added in gRPC protocol that ability but not in that MQTT5 client. To implement MQTT3.11 client we need add another implementations of MqttConnection and extend MqttLibImpl due to not only another client is used but also all underlying objects like options, arguments, packets and so one.

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCControlServer.java
bgklika: Fixed in 1db6093b6edb3f292620f48e0392d713d1900a5d

uat/src/main/java/com/aws/greengrass/testing/mqtt5/client/MqttReceivedMessage.java
MikeDombo: I'm not sure I see the benefit here, transforming between data types with so many fields opens up opportunities to do the mapping incorrectly. Right now you need to transform from AWS CRT type to intermediate type to gRPC type and all the transforms the other way around too.

"adding another mqtt part" -- another mqtt part of what? If you use the gRPC type as our main data type, then don't you have the same benefit as creating your own intermediate type? I understand not using the AWS CRT type, but the gRPC type is within our control in this package, so that would seem to be OK to me.

bgklika: Another MQTT can be CRT MQTT 3.11 or another MQTT5 implementation on Java.
Idea is do not poison the MQTT code with gRPC types. 
Currently it works in both directions. Sure we have double conversion here.
Benefits: 
1. Code has visible borders.
2. Changes in gRPC protocol do not touch MQTT code (in future multiple implementations). At the moment gRPC proto is the most frequently changed part due to limited time to deep design.
3. Easy to write unit test based on API only objects (not need to check generated code of gRPC)

MikeDombo (approved):

PR #235, opened 2023-03-10
fix: inject domain events in SessionManager

sameerzuberi:
**Description of changes:**
Injecting Domain Events into the Session Manager instead of creating a new domain events object with the class.

By creating a new domain events object, the session manager was overriding all of the existing handlers that had been registered in domain events. This caused an issue with the GetClientDeviceAuthToken metrics, since they could not be emitted when domain events was being cleared out.

Updated unit tests.


MikeDombo (commented):
Don't commit that telemetry file.

Also make sure this same mistake isn't anywhere else.

MikeDombo (approved):

jcosentino11 (approved):

PR #238, opened 2023-03-14
fix: log details of cert attachment use case

jcosentino11:
**Issue #, if available:**

**Description of changes:**

The goal of the change is to clearly log the results of cert attachment, such that a user is able to debug why attachment may have failed, possibly due to local attachment expiring, for example.

Also added missing test coverage for verifying expired local cert attachment

We recently had a UAT failure likely due to failed attachment verification, due to expiry. This was determined by examining timestamps, so this change would remove any guesswork.


MikeDombo (commented):
This feels like a lot to change just to get a few more logs. Any room to simplify this?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -113,11 +113,7 @@ public Map<String, Instant> getAttachedCertificateIds() {
      * @param certificateId - A certificateId
      */
     public Optional<Instant> certificateLastAttachedOn(String certificateId) {
-        if (!attachedCertificateIds.containsKey(certificateId)) {
-            return Optional.empty();
-        }
-
-        return Optional.of(attachedCertificateIds.get(certificateId));
+        return Optional.ofNullable(attachedCertificateIds.get(certificateId));
```
MikeDombo, line 116: good call, that was a race before.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/CreateIoTThingSession.java
```diff
@@ -64,16 +67,23 @@ public Session apply(CreateSessionDTO dto) throws AuthenticationException {
...
 
-            if (thingAttachedResult) {
+            logger.atDebug()
+                    .kv("thingIsAttached", result.isThingAttachedToCertificate())
+                    .kv("lastAttachedOn", result.getLastAttached())
+                    .kv("attachmentExpiration", result.getAttachmentExpiration())
+                    .kv("source", result.getVerificationSource())
+                    .log("Attachment verification result");
+
+            if (result.isThingAttachedToCertificate()) {
```
MikeDombo, line 80: check for expiration here too, not just that it is attached?

jcosentino11: ah yeah that was a miss, fixing

jcosentino11: spoke too soon, it's covered here: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/238/files#diff-d564d0cc489527df3742905afa00509b6cfb6ebdb07597c35138150a4b2c1a45R56

probably needs a wording

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
jcosentino11: :)

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/CreateIoTThingSession.java
jcosentino11: changed wording from `thingAttachedToCertificate` to `thingHasValidAttachmentToCertificate`,  hopefully more clear

MikeDombo (approved):

MikeDombo:
Updated title. This is a fix, not a feature. Feature would mean that we need to bump the minor version and that's definitely no the case for this change.

robcmann (approved):

PR #247, opened 2023-03-27
feat(uat): added ggad scenario for otf

esmirnov-klika-tech:
**Description of changes:**
- modified GGAD-1-T1 scenario
- added placeholder for cucumber step definitions
- fixed pom to use client jar as dependency


uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
```diff
@@ -0,0 +1,26 @@
...
+      port: 47619
+Manifests:
+  - Artifacts:
+      - URI: classpath:/greengrass/artifacts/aws.greengrass.client.Mqtt5JavaSdkClient.jar
+        Permission:
+          Read: ALL
+          Execute: ALL
+    Lifecycle:
+      Run: |
+        java -jar {artifacts:path}/aws.greengrass.client.Mqtt5JavaSdkClient.jar {configuration:/agentId} {configuration:/broker/address} "{configuration:/broker/port}"
```
MikeDombo, line 26: Let's use `Startup` and use Greengrass IPC to report that the component is in the running state. https://docs.aws.amazon.com/greengrass/v2/developerguide/ipc-component-lifecycle.html#ipc-operation-updatestate

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
+@GGAD
+Feature: GGMQ-1
+
+  As a developer, I can configure my client agents to connect and use MQTT
+
+  Background:
+    Given my device is registered as a Thing
+    And my device is running Greengrass
+    And my MQTT Client Control is running
```
MikeDombo, line 9: what is this step going to do exactly? This isn't clear. This isn't something that an end-customer would understand. The point of using cucumber is that these steps are directly understandable by non-techincal readers.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+
+  Scenario: GGMQ-1-T1: As a customer, I can connect, subscribe and publish using client application to MQTT topic
```
MikeDombo, line 11: this scenario is missing critical steps.

The client device auth component must be configured in order to allow the client device to connect, publish, subscribe, etc.

What I want is to actually configure the auth component in the same way that a customer would configure it, by writing the device group selectors and policies for those selectors.

Something like:

```json
{
  "deviceGroups": {
    "formatVersion": "2021-03-05",
    "definitions": {
      "MyDeviceGroup": {
        "selectionRule": "thingName: ${clientDevice1}",
        "policyName": "MyClientDevicePolicy"
      }
    },
    "policies": {
      "MyClientDevicePolicy": {
        "AllowConnect": {
          "statementDescription": "Allow client devices to connect.",
          "operations": [
            "mqtt:connect"
          ],
          "resources": [
            "*"
          ]
        },
        "AllowPublish": {
          "statementDescription": "Allow client devices to publish to all topics.",
          "operations": [
            "mqtt:publish"
          ],
          "resources": [
            "*"
          ]
        },
        "AllowSubscribe": {
          "statementDescription": "Allow client devices to subscribe to all topics.",
          "operations": [
            "mqtt:subscribe"
          ],
          "resources": [
            "*"
          ]
        }
      }
    }
  }
}
```

Where, `${clientDevice1}` would be replaced by the client device's thing name.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+    When I create a Greengrass deployment with components
+      | aws.greengrass.clientdevices.Auth        | LATEST                                                        |
+      | aws.greengrass.clientdevices.mqtt.EMQX   | LATEST                                                        |
+      | aws.greengrass.clientdevices.IPDetector  | LATEST                                                        |
+      | aws.greengrass.client.Mqtt5JavaSdkClient | classpath:/greengrass/components/recipes/client_java_sdk.yaml |
+    Then the Greengrass deployment is COMPLETED on the device after 300 seconds
+    And agent "agent1" is connected to MQTT Client Control
+    When I create client device "test" with the following policy
```
MikeDombo, line 19: this implies an IoT Policy to me, but it isn't clear. Make it clear what this policy is exactly.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+    And my MQTT Client Control is running
+
+  Scenario: GGMQ-1-T1: As a customer, I can connect, subscribe and publish using client application to MQTT topic
+    When I create a Greengrass deployment with components
+      | aws.greengrass.clientdevices.Auth        | LATEST                                                        |
+      | aws.greengrass.clientdevices.mqtt.EMQX   | LATEST                                                        |
+      | aws.greengrass.clientdevices.IPDetector  | LATEST                                                        |
+      | aws.greengrass.client.Mqtt5JavaSdkClient | classpath:/greengrass/components/recipes/client_java_sdk.yaml |
+    Then the Greengrass deployment is COMPLETED on the device after 300 seconds
+    And agent "agent1" is connected to MQTT Client Control
```
MikeDombo, line 18: this step again isn't very clear what it means and it isn't something a customer would do.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+      | subscribe | "iot_data_0" |
+      | publish   | "iot_data_0" |
+    And I associate "test" with ggc
+    And I connect device "test" on "agent1" to "default_broker" as "agent1-test-default_broker"
+    Then connection "agent1-test-default_broker" is successfully established within 3 seconds
+    When I subscribe "agent1-test-default_broker" to "iot_data_0" with qos 0
+    Then subscription to "iot_data_0" is successfull on "agent1-test-default_broker"
+    When I publish "agent1-test-default_broker" to "iot_data_0" with qos 0 and message "Test message"
+    Then publish message "test" to "iot_data_0" is successfully on "agent1-test-default_broker"
+    And message "Test message" received on "agent1-test-default_broker"
```
MikeDombo, line 31: validate we receive it on the expected topic

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+      | connect   | "*"          |
+      | subscribe | "iot_data_0" |
+      | publish   | "iot_data_0" |
+    And I associate "test" with ggc
+    And I connect device "test" on "agent1" to "default_broker" as "agent1-test-default_broker"
+    Then connection "agent1-test-default_broker" is successfully established within 3 seconds
+    When I subscribe "agent1-test-default_broker" to "iot_data_0" with qos 0
+    Then subscription to "iot_data_0" is successfull on "agent1-test-default_broker"
+    When I publish "agent1-test-default_broker" to "iot_data_0" with qos 0 and message "Test message"
+    Then publish message "test" to "iot_data_0" is successfully on "agent1-test-default_broker"
```
MikeDombo, line 30: Let's simplify this: `Then the message is published with reason code 0`, for example.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+      | aws.greengrass.client.Mqtt5JavaSdkClient | classpath:/greengrass/components/recipes/client_java_sdk.yaml |
+    Then the Greengrass deployment is COMPLETED on the device after 300 seconds
+    And agent "agent1" is connected to MQTT Client Control
+    When I create client device "test" with the following policy
+      | operation | resource     |
+      | connect   | "*"          |
+      | subscribe | "iot_data_0" |
+      | publish   | "iot_data_0" |
+    And I associate "test" with ggc
+    And I connect device "test" on "agent1" to "default_broker" as "agent1-test-default_broker"
```
MikeDombo, line 25: What is "default_broker"? Where is that coming from?

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,32 @@
...
+    Then connection "agent1-test-default_broker" is successfully established within 3 seconds
```
MikeDombo, line 26: Everything is very "agent" based here. Let's make this client-based as client is something well understood.

`Then client "test" is successfully connected within 3 seconds`

Make a reasonable agent name automatically and store the association between client and agent so that we don't need to keep referencing the agent here in the test scenario.

MikeDombo (changes requested):

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
bgklika: We have a little more complicated architecture than in GGAT examples.
In GGAT we have a (maybe real) IoT device which is up once will infinite try to make connection with IoT or Greengrass device. In terms of GGAT client it is the same as MQTT connection. 
But it is not true for our architecture due to the fact that we have a client fully controlled. Initiatially when started it has no MQTT connection and just establishes bidirectional  gRPC link with Control. After that control can send requests to establish one or more MQTT connections. That means our 'client' is not equal to one MQTT connection. This way to avoid misunderstanding I offer a term agent. Which can have more than one MQTT connection. With that architecture we need to specify an agentId and connectionId pair to do connection-related operations like publish or subscribe. To simplify schema for end users I already implement logical connection names and how we can use any unique string to identify a particular connection despide of agent.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
bgklika: It is possible by using default logical connection name, but currently we want to explain architecture and possibilities in details.

With that architecture we can start several agents (identified by agentId string) and can create several MQTT connection on each.

in "agent1-test-default_broker" means logical connection id which can by just "connection1" for example or "last_connection" by default.

GADD examples have have just one MQTT connection. With that we can't test new MQTT5 features like group subscription.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
bgklika: And in limited amount of time due to messages are received asynchronous.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
MikeDombo: This isn't a step that a user would do, so let's hide it. In any step that needs the control to be running, verify that it is running, or start it if it isn't running. Then this step doesn't need to exist.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Since we have not implemented building blocks yet, this is only a placeholder.
I think we should have a test method with ability to specify which broker to use. Also we should have an overloaded version of this method with some defaults configured.

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
esmirnov-klika-tech: Changed to use Startup 369a951b8f6c1ab252d97cacf06c211091282565

esmirnov-klika-tech: Could you clarify about IPC usage, please? Would you like to change "Then the Greengrass deployment is COMPLETED on the device after 180 seconds" step implementation?

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Fixed c6275de8e27218f059bdf629c98e144c6891a718

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Fixed b6f305abbeb79765805f08540493ea08711e51e5

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
MikeDombo: No, I do not want to change the implementation of the deployment. I'm suggesting that the implementation Mqtt5JavaSdkClient.jar should use Greengrass IPC to report that it is in the running state when it is running.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
MikeDombo: That's the point of this pull request, I want us to have the test nearly exactly how I want it to look; then you go and implement the logic.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
MikeDombo: That's the point of this pull request, I want us to have the test nearly exactly how I want it to look; then you go and implement the logic.

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
```diff
@@ -0,0 +1,26 @@
...
+      address: 127.0.0.1
+      port: 47619
+Manifests:
+  - Artifacts:
+      - URI: classpath:/greengrass/artifacts/aws.greengrass.client.Mqtt5JavaSdkClient.jar
+        Permission:
+          Read: ALL
+          Execute: ALL
+    Lifecycle:
+      Startup: |
```
MikeDombo, line 25: just changing this to startup without changing the code is not correct. Leave it as run until the code is updated.

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,31 @@
...
+      | aws.greengrass.clientdevices.mqtt.EMQX   | LATEST                                                        |
+      | aws.greengrass.clientdevices.IPDetector  | LATEST                                                        |
+      | aws.greengrass.client.Mqtt5JavaSdkClient | classpath:/greengrass/components/recipes/client_java_sdk.yaml |
+    Then the Greengrass deployment is COMPLETED on the device after 300 seconds
+    And agent "agent1" is connected to MQTT Client Control
+    When I create client device "test" with the following AWS IoT policy
+      | operation | resource     |
+      | connect   | "*"          |
+      | subscribe | "iot_data_0" |
+      | publish   | "iot_data_0" |
```
MikeDombo, line 22: you copied this from our existing test which described _local_ MQTT policies applied to the client device auth component. The client device should NOT have these permissions in the iot policy.

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
esmirnov-klika-tech: Fixed d2ca07c1c3a20331a116f61e1e820d4553066829

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Changed to separate Auth component configuration step 66c017923d2748f670f6b81ae3c9d7d80dbde216

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -15,12 +15,35 @@ Feature: GGMQ-1
...
+    When I create client device "test"
     And I associate "test" with ggc
+    And I configure aws.greengrass.clientdevices.Auth component
+    """
+{
+  "deviceGroups": {
+    "formatVersion": "2021-03-05",
+    "definitions": {
+      "MyPermissiveDeviceGroup": {
+        "selectionRule": "thingName: test",
```
MikeDombo, line 27: the thing name should not actually be "test" - this will cause problems when the test is run in parallel on multiple devices. Thing names and anything else which is shared must have uniquely generated names which are replaced at runtime. This is why I suggested `${test}` or similar, as a string replacement.

esmirnov-klika-tech: I see. Is it ok, if we change  thing creation step to something like:
"When I create client device 'test' with random prefix in thing name" ?
Name mapping will be stored in the test context.
Then we will be able to use short name in test scenarios and full thing name when it is necessary.
For example:
"""
"selectionRule": "thingName: ${testContext.getThingNameForDevice('test')}",
"""
What do you think?

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Added 66c017923d2748f670f6b81ae3c9d7d80dbde216

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
MikeDombo: Right, like that. I don't think you need to say "with random prefix in thing name", we can make the assumption that it will need to always be random.

esmirnov-klika-tech: Fixed ed88cd4b6ed567a4a912fc9635fa36c4ca271d5b

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Fixed ced07c0b79ca159d1f64195e44e23cdb61bbb7cd
I've added step to discover a broker using IPDetector. We have the assumption that we'll need some additional steps with direct configuration. 
Also let me clarify, Is it ok to run discovery once on the test application? Or you'd like us to implement discovery in test clients code?

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -44,11 +44,11 @@ Feature: GGMQ-1
...
     """
-    And I connect device "test" on "agent1" to "default_broker" as "agent1-test-default_broker"
-    Then connection "agent1-test-default_broker" is successfully established within 3 seconds
-    When I subscribe "agent1-test-default_broker" to "iot_data_0" with qos 0
-    Then subscription to "iot_data_0" is successfull on "agent1-test-default_broker"
-    When I publish "agent1-test-default_broker" to "iot_data_0" with qos 0 and message "Test message"
-    Then publish message "test" to "iot_data_0" is successfully on "agent1-test-default_broker"
-    And message "Test message" received on "agent1-test-default_broker" from "iot_data_0" topic within 1 second
+    And I connect device "clientDeviceTest" on "agent1" to "default_broker" as "connection1"
+    Then connection "connection1" is successfully established within 3 seconds
```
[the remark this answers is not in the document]
bgklika: With this assumption we will miss all test cases when requires more than one connection for thingName. For example we can connnect one Thing to Iot Core directly and to local Greengrass. Another case create second connection to IoT core to check first connection will be broken. It is possible as test scenario. More production case: connect the same Thing to Greengrass1 and Greengrass2 as backup link,  I do not know allow IoT core how association of Thing with more that one Greengrass but can imagine such production scenaries. We know AWS has a millions of consumers with different needs, so we design control to be flexible as posible. Maybe we can assign Thing name to connectionName by default and simplify most of scenarios.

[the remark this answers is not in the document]
MikeDombo: You cannot have more than one connection per thing name. Thing name must be exactly equal to the client ID and client IDs must be unique. It is impossible to have more than one connection per thing.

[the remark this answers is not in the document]
bgklika: I know it is limitation of MQTT broker of IoT Core, but it may be not true for all brokers or for set of broker. At least we should be able to test that limitation of IoT core, right ?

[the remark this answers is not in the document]
MikeDombo: This test is for Greengrass, this is requirement of the Greengrass broker.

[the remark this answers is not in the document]
esmirnov-klika-tech: The current implementation of test clients is flexible enough to establish several connections with different thing certificates to different brokers from single instance called "agent".
Do you want to hide this connection based logic or to make it optional? 
For example, after setup  "And I connect device "clientDeviceTest" on "agent1" to "default_broker" as "connection1"
we may execute same action using different syntax:
1. When I subscribe connection "connection1" ...
2. When I subscribe device "clientDeviceTest" ...
3. When I subscribe agent1 "agent1" ...
if there is only one connection then this steps will give us the same result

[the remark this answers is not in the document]
bgklika: @esmirnov-klika-tech I thing we can use thing name (exactly the same as clientId in terms of MQTT protocol) as connection name by default. It makes most of scenarios simplest. At the same time by design we will be able to write extended test scenarios by assign non-default connection name.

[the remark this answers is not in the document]
esmirnov-klika-tech: Fixed 44986341adb2fd0b5bfbd49371c60b6ae4edc2f9

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
esmirnov-klika-tech: Fixed 83977311c0ea2d24e15855137dfc88e34854cc4e

uat/testing-features/src/main/resources/greengrass/features/ggmq-1.feature
```diff
@@ -0,0 +1,55 @@
...
+      | aws.greengrass.clientdevices.Auth        | LATEST                                                        |
+      | aws.greengrass.clientdevices.mqtt.EMQX   | LATEST                                                        |
+      | aws.greengrass.clientdevices.IPDetector  | LATEST                                                        |
+      | aws.greengrass.client.Mqtt5JavaSdkClient | classpath:/greengrass/components/recipes/client_java_sdk.yaml |
+    Then the Greengrass deployment is COMPLETED on the device after 300 seconds
+    And agent "agent1" is connected to MQTT Client Control
+    When I create client device "clientDeviceTest"
+    And I associate "clientDeviceTest" with ggc
+    And I update my Greengrass deployment configuration, setting the component aws.greengrass.clientdevices.Auth configuration to:
+    """
```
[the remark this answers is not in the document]
esmirnov-klika-tech: Done 322131a0465986bccd5449dc0925764897e3851b

MikeDombo (approved):

bgklika (approved):

PR #249, opened 2023-03-28
feat(uat): create event storage in uat

bgklika:
**Issue #, if available:**
Create event storage

**Description of changes:**
- Added API of EventStorage, Event and EventFilter with builder
- Added implementation of EventStorageImpl and MqttMessageEvent
 
Received events like MQTT messages, disconnect (and partially connect) events can't be handled asynchronously and we need to store it. In addition we need to implement a counter of received messages matched to filter in future. We decided to implement EventStore with the ability to clear, put events, get events by filter and wait for some time for future events if currently missing.

That code is not yet used, not able to run manual tests. No unit tests has been created.

### DIFF
--- a/uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/ConnectionControl.java
+++ b/uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/ConnectionControl.java
@@ -16,6 +16,14 @@ import lombok.NonNull;
  * Control of single MQTT connection.
  */
 public interface ConnectionControl {
+
+    /**
+     * Gets agent control for this connection control.
+     *
+     * @return the agent control
+     */
+    AgentControl getAgentControl();
+
     /**
      * Gets id of the connection.
      *
--- a/uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/AgentControlImpl.java
+++ b/uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/AgentControlImpl.java
@@ -50,6 +50,7 @@ public class AgentControlImpl implements AgentControl {
     // That lock prevent cases when messages are received but connection not yet registered.
     private final Lock connectLock = new ReentrantLock();
 
+    private final EngineControlImpl engineControl;
     private final String agentId;
     private final String address;
     private final int port;
@@ -69,12 +70,15 @@ public class AgentControlImpl implements AgentControl {
     /**
      * Creates instanse of AgentControlImpl.
      *
-     * @param agentId id of agent
-     * @param address address of gRPC server of agent (MQTT client)
-     * @param port port of  gRPC server of agent (MQTT client)
+     * @param engineControl the back reference to whole engineControl
+     * @param agentId the id of agent
+     * @param address the address of gRPC server of agent (MQTT client)
+     * @param port the port of gRPC server of agent (MQTT client)
      */
-    public AgentControlImpl(@NonNull String agentId, @NonNull String address, int port) {
+    public AgentControlImpl(@NonNull EngineControlImpl engineControl, @NonNull String agentId, @NonNull String address,
+                                int port) {
         super();
+        this.engineControl = engineControl;
         this.agentId = agentId;
         this.address = address;
         this.port = port;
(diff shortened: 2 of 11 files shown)


uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/EventFilter.java
```diff
@@ -0,0 +1,74 @@
...
+        this.connectionName = builder.connectionName;
+        this.topic = builder.topic;
+        this.topicFilter = builder.topicFilter;
+        this.content = builder.content;
+    }
+
+    /**
+     * Builder of event filter.
+     */
+    public static class Builder {
```
MikeDombo, line 34: we can use Lombok to get rid of all the boilerplate code

bgklika: We have overloaded withContent() with String and byte[] with lombok we will have only one withContent(byte[])

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/EventStorage.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    void addEvent(@NonNull Event event);
+
+    /**
+     * Gets matched events from storage immediately.
+     *
+     * @param filter the filter of events
+     * @return the list of matched events
+     */
+    List<Event> getEvents(@NonNull EventFilter filter);
```
MikeDombo, line 37: I think that all this filtering should be done in our UAT itself and not in the client control. 

The agent should expose an API for callbacks on connect, disconnect, publish, subscribe and anything else that's needed.
The UAT will then listen for those events and can handle it however it wants.

bgklika: Yes, actually whole EventStorage is an add-on to Control.
Due to building blocks is not yet ready I place it to Control

MikeDombo: Let's get the work on the UAT then. This change seems really large for what it is trying to accomplish and I believe that doing this in the UAT will be much more reasonable.

bgklika: It not yet fully finished due to collect only MQTT messages. We are going to collect all events like agent registration/unregistration, MQTT connection/disconnection, and sure MQTT messages.

Ok, will rework to keep it separately from Control on UAT level.

bgklika: Moved to testing-features in 07f5113bf5eb1e3c7c7f99ea87e915082be7d665

bgklika:
You noted the simplest scenario. It does not cover all new MQTT5 features.
For example with group subscription we should verify exactly one message is received even on the several subscribers. We can extend the case to use 2 groups and part of subscribers who subscribed on both. That not simple anymore?

With that EventStorage we can just wait N seconds, make getEvent() with a suitable filter and  call List.size() to check if the group subscription works or not.

Actually we should be able to test any functional behaviour of the whole broker-bridge-IoTCore chain for any existing and future customer, not only a simple MQTT client.

MikeDombo:
> You noted the simplest scenario. It does not cover all new MQTT5 features. For example with group subscription we should verify exactly one message is received even on the several subscribers. We can extend the case to use 2 groups and part of subscribers who subscribed on both. That not simple anymore?

You can do that using the feature set that I described.

> With that EventStorage we can just wait N seconds, make getEvent() with a suitable filter and call List.size() to check if the group subscription works or not.

Yes, you can do that using what you have implemented, but that's not the only way. 600 lines of code added just to see that an event has happened is too much.

> Actually we should be able to test any functional behaviour of the whole broker-bridge-IoTCore chain for any existing and future customer, not only a simple MQTT client.

We can always add more capability, but I really want us to push forward on implementing the feature set we need right now. This is premature abstraction with the assumption that we'll need these features in the future.



Now that we have the basic scenario outline feature implemented, I would like to see the next PRs implement functionality which is required to make that test pass. Each PR should implement one more of the steps and we should be able to see that the test has more and more steps passing.

PR #259, opened 2023-04-14
feat(uat): implement subscribe publish message_is_received steps

bgklika:
**Issue #, if available:**
Implement "message is received" step
Implement subscribe step
Implement publish step
Implement event storage

**Description of changes:**
- Add API of EventStore, Event and EventFilter with builder
- Add implementation of EventStorageImpl and MqttMessageEvent
- Update readme files
- Make timeouts in control stackable
- Add factory for EventStorage
- Use agentId equals to ComponentName
- Remove check steps from scenario
- Rework MqttControlSteps.java to implement publlish and subscribe
- Update readme files to be actual
- Inject EventStorage to steps implementation class
- Make AgentControlImpl constructors chained
- Fixed scenario
- Cleanup outdated tests
 
We should implement building blocks for subscribe, publish and receive MQTT messages.

Part of code is covered by unit tests, some - part of Cucumber's scenario steps.
Unit test run on CI, uat not yet.

**Test results:**
```
Given my device is registered as a Thing....................................passed
And my device is running Greengrass.........................................passed
When I create a Greengrass deployment with components.......................passed
And I create client device "clientDeviceTest"...............................passed
When I associate "clientDeviceTest" with ggc................................passed
And I update my Greengrass deployment configuration, setting the component aws.greengrass.clientdevices.Auth configuration to:.passed
And I update my Greengrass deployment configuration, setting the component aws.greengrass.client.Mqtt5JavaSdkClient configuration to:.passed
And I deploy the Greengrass deployment configuration........................passed
Then the Greengrass deployment is COMPLETED on the device after 300 seconds.passed
And I discover core device broker as "default_broker" from "clientDeviceTest".passed
And I connect device "clientDeviceTest" on aws.greengrass.client.Mqtt5JavaSdkClient to "default_broker".passed
When I subscribe "clientDeviceTest" to "iot_data_0" with qos 0..............passed
When I publish from "clientDeviceTest" to "iot_data_0" with qos 0 and message "Test message".passed
And message "Test message" received on "clientDeviceTest" from "iot_data_0" topic within 5 seconds.passed
```


bgklika:
> We identified new issues on unchanged lines of code. Navigate to the Amazon CodeGuru Reviewer console to view the recommendations to fix them.

@MikeDombo Could you provide more details?

MikeDombo:
No. That comment is not from me, it is automated.

Ignore it.

uat/README.md
```diff
@@ -6,6 +6,17 @@ the `aws.greengrass.clientdevices.Auth` component.
 
 ## Running UATs locally
 
+### Requirements
+Some scenarios require to install and run the EMQX, Auth, IPDetector Greengrass components.
+
+#### Docker
+Because EMQX compoment running in a docker containter docker should be installed on workstation where scenario is running.
+
+#### Rrivileges
```
MikeDombo, line 15: typo

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/addon/EventFilter.java
```diff
@@ -0,0 +1,190 @@
...
+import lombok.Getter;
+import lombok.NonNull;
+
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Filter of events in event storage.
+ */
+@Getter
+public final class EventFilter {
```
MikeDombo, line 18: I still really want to greatly simplify this.

What I'd like is for the agent to emit all events to the UAT - CONNACK, disconnect, message received. We will then handle these events within the UAT itself; not in the agent. The agent should only forward the events and have all the logic inside the UAT.

uat/README.md
bgklika: Fixed in 9ab68b0a061e41903c9de947922073777de93463

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/addon/EventFilter.java
bgklika: All businesses  logic are in UAT.
EventStorage is a utility to store events received by control currently only received MQTT messages.
In future we can extend for other types of events including unsolicited disconnections and so one.

EventStorage is required in any case when we expecting more that one event i.e. MQTT messages.
EventFilter just a filter in EventStorage used to skip not related events during waiting.
That code developed separately from UAT part because it is a addon for Control and can be developed outside UAT in parallel with building blocks. That utility move complexity from UAT to other place provides clean border between part via API and allow create unit test before UAT part has been developed.

It is really required if we are expecting more that one MQTT message during the time we waiting for broker's delivery.
I have no idea how this can be simplified by moving directly to UAT without implementing a lot of custom error prone cases.


CONNACK, SUBACK, UNSUBACK and so one are passed from agent to control and provided to users of control's API as before.

MikeDombo: > That code developed separately from UAT part because it is a addon for Control and can be developed outside UAT in parallel with building blocks.

The purpose of all of this code is to support the UAT, we aren't trying to solve abstract problems. We have very concrete problems that need to be implemented.

The UAT should be receiving these events and handling them how it chooses. I do not want an additional unnecessarily complex storage and filtering layer and API.

bgklika: That code works and solve concrete problems. I have no idea how I can simplify that, sorry.

MikeDombo: Simplify by deleting this code.

What I am suggesting is that all events should be forwarded to the UAT. The UAT can then choose how it wants to handle those events. Storing them in a list, executing callback when an event happens, whatever we want to do. I want the logic to be in the UAT itself, not here.

bgklika: We just move that code to uat, by removing class borders and drop ability to use unit test.
It will  produce solution worst in reading, testing and finally quality.
And nothing will be added.

bgklika: I have several concerns here:
1. Technical. The above example of code functionally not the same as implemented in EventStorageImpl and EventFilter
 a. That code misses filtering by clientId/thingName or connection. That means in "message {string} received on {string} " clientDeviceId is unused. That building block will not be suitable when we have more than a single device. But MQTT5 features like group subscription require that. In other words that implementation is not scalable. It resulting in bigger code in feature and error prone misuse of that building block.
 b.  I have not found documentation on eventuallyEval() but from the name expecting code will check the list after the whole timeout, even when a message is already in the list or received early. It will slow down the tests. Not so critical but disadvantageous.
2. Organizational.
 a. It is hard to do coding on the same file with several programmers, as well as implement receive when no connection or publish is yet implemented. To provide the ability to work in parallel I develop from bottom and other devs from top of the project. This way we must define interfaces as a midway road milestone where we should meet. When other developers did one part of the job I was able to implement functionally isolated classes and test it. Just because we have no other chance to test it before and do not waste time when all untested code dropped to a single pool.
 b. Two developers always implement the same requirement in different ways. It is usual. It depends on their experience and strong and weak  sides. For example if you have good knowledge of modern steaming API lambdas and FI  but I'm not. I prefer the procedural approach because my preferred programming language is still C. As I said it requires time to learn these things but we have a tight schedule.

Don't worry about the number of lines. Complexity of a project depends not on the number of lines but on good (or bad) separation to parts like classes and other black boxes. Number of lines does not affect the cost of the project. But number of lines it mirrors the nature of more complex and flexible architecture with multiple clients and control. Useless to reproduce your already done testing tool, we need build something better and that way defined in project goals and requirements.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/EventStorageImpl.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+
+/**
+ * Implementation storage of the events.
+ */
+public class EventStorageImpl implements EventStorage {
+
+    private final List<Event> events;
```
MikeDombo, line 24: Why not store the events on the Agent (or Connection) itself? Then we would not need to filter the events by which agent it came from, only need to lookup the Agent and get the list of events from it.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/MqttMessageEvent.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+/**
+ * Implements received MQTT message event.
+ */
+public class MqttMessageEvent extends EventImpl {
+
+    private final ConnectionControl connectionControl;
+    private final Mqtt5Message message;
```
MikeDombo, line 23: add a getter for the message so we can inspect it outside of the event if we want.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/MqttMessageEvent.java
```diff
@@ -0,0 +1,114 @@
...
+
+        ByteString byteStringPayload = message.getPayload();
+        if (byteStringPayload == null) {
+            return false;
+        } else {
+            return Arrays.equals(expected, byteStringPayload.toByteArray());
+        }
+    }
+
+    private static boolean isTopicMatched(@NonNull String topic, @NonNull String topicFilter) {
```
MikeDombo, line 110: You can steal the implementation from the Nucleus:
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/94a91fb3344a9069462d7bb8828d1376304ec67b/src/main/java/com/aws/greengrass/mqttclient/MqttTopic.java#L44

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/addon/EventStorage.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.NonNull;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Interface to storage of the events.
+ */
+public interface EventStorage {
```
MikeDombo, line 17: this does not need to be an interface. This can be concrete.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/EventStorageImpl.java
MikeDombo: The goal of the project does not include testing MQTT broker compliance, testing shared subscriptions is not needed. Even if it is needed, it is easy to validate using the procedural implementation that I described earlier (not using your object oriented event filter approach).

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/addon/EventFilter.java
MikeDombo: 1.a. of course I just gave a simple example which can be extended with a single line of code to filter for anything else.

1. b. https://github.com/grantwest/eventually-matchers. This does not wait for the timeout before checking, that would be wasteful as you said. It will poll the method on a default or specified interval until the condition succeeds or times out.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/MqttMessageEvent.java
bgklika: Added in be1c3a7d3605ffadf4a972c3b5c0b21ac0669512

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/MqttMessageEvent.java
MikeDombo: Yes please copy the implementation into a utility class for the mqtt topics, then use the implementation here.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/EventStorageImpl.java
bgklika: We obligated to do that test in Phase 2 of the project. Please check actual SoW document here https://docs.google.com/document/d/1ILcnkuCpK_SoUclyOF4GAjStoyJOUqGD  5.4  point 5. x.  about shared subscriptions

MikeDombo: You do not need to validate the broker implements shared subscriptions properly, only that you are able to publish to a shared subscription topic and receive the message in a subscriber to the shared subscription topic.

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/implementation/addon/MqttMessageEvent.java
bgklika: Added in fed30933e0c81aa468fdcd2bd25fb21bdd08cfc0

uat/mqtt-client-control/src/main/java/com/aws/greengrass/testing/mqtt/client/control/api/addon/EventStorage.java
bgklika: Interface removed in 514910c3347e6c98ebf847a9ef2be1514e3aa1a5

MikeDombo (approved):

PR #262, opened 2023-04-19
feat(uat): add support of remote clients and multi-homed control

bgklika:
**Issue #, if available:**
Multi-homed control is not supported

**Description of changes:**
- Add detection of control addresses and pass to client
- Allow SDK-base MQTT client to receive multiple IP/host of control

Currently agent got hard coded 127.0.0.1 as address of control.
That limit clients to be run only locally.
Implementation is changed to get actual list of IP addresses how control can be reached and pass it to the agent arguments.

Unit test added to new method of control.
Other parts can be verified during scenario run.

### DIFF
--- a/uat/custom-components/client-java-sdk/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCLibImpl.java
+++ b/uat/custom-components/client-java-sdk/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCLibImpl.java
@@ -17,7 +17,7 @@ public class GRPCLibImpl implements GRPCLib {
     private final LinkFactory linkFactory;
 
     interface LinkFactory {
-        GRPCLink newLink(@NonNull String agentId, @NonNull String host, int port) throws GRPCException;
+        GRPCLink newLink(@NonNull String agentId, @NonNull String[] hosts, int port) throws GRPCException;
     }
 
     /**
@@ -26,8 +26,8 @@ public class GRPCLibImpl implements GRPCLib {
     public GRPCLibImpl() {
         this(new LinkFactory() {
             @Override
-            public GRPCLink newLink(@NonNull String agentId, @NonNull String host, int port) throws GRPCException {
-                return new GRPCLinkImpl(agentId, host, port);
+            public GRPCLink newLink(@NonNull String agentId, @NonNull String[] hosts, int port) throws GRPCException {
+                return new GRPCLinkImpl(agentId, hosts, port);
             }
         });
     }
@@ -43,7 +43,7 @@ public class GRPCLibImpl implements GRPCLib {
     }
 
     @Override
-    public GRPCLink makeLink(@NonNull String agentId, @NonNull String host, int port) throws GRPCException {
-        return linkFactory.newLink(agentId, host, port);
+    public GRPCLink makeLink(@NonNull String agentId, @NonNull String[] hosts, int port) throws GRPCException {
+        return linkFactory.newLink(agentId, hosts, port);
     }
 }
--- a/uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
+++ b/uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
@@ -13,7 +13,7 @@ ComponentConfiguration:
   DefaultConfiguration:
     # agentId should be the same as ComponentName
     agentId: aws.greengrass.client.Mqtt5JavaSdkClient
-    controlAddress: 127.0.0.1
+    controlAddresses: 127.0.0.1
     controlPort: 47619
 Manifests:
   - Artifacts:
@@ -23,4 +23,4 @@ Manifests:
           Execute: ALL
     Lifecycle:
       Run: |
-        java -jar {artifacts:path}/aws.greengrass.client.Mqtt5JavaSdkClient.jar "{configuration:/agentId}" "{configuration:/controlPort}" "{configuration:/controlAddress}"
+        java -jar {artifacts:path}/aws.greengrass.client.Mqtt5JavaSdkClient.jar "{configuration:/agentId}" "{configuration:/controlPort}" {configuration:/controlAddresses}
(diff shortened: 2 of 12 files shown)


uat/custom-components/client-java-sdk/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCLibImpl.java
```diff
@@ -17,7 +17,7 @@ public class GRPCLibImpl implements GRPCLib {
     private final LinkFactory linkFactory;
 
     interface LinkFactory {
-        GRPCLink newLink(@NonNull String agentId, @NonNull String host, int port) throws GRPCException;
+        GRPCLink newLink(@NonNull String agentId, @NonNull String[] hosts, int port) throws GRPCException;
```
MikeDombo, line 20: use `List<String>` instead of `String[]` if possible. Dealing with lists is much simpler than builtin arrays.

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
```diff
@@ -23,4 +23,4 @@ Manifests:
           Execute: ALL
     Lifecycle:
       Run: |
-        java -jar {artifacts:path}/aws.greengrass.client.Mqtt5JavaSdkClient.jar {configuration:/agentId} {configuration:/controlAddress} "{configuration:/controlPort}"
+        java -jar {artifacts:path}/aws.greengrass.client.Mqtt5JavaSdkClient.jar "{configuration:/agentId}" "{configuration:/controlPort}" {configuration:/controlAddresses}
```
MikeDombo, line 26: I think you'll want to JSON encode the control address array and pass it as stringified JSON, this will make parsing on the other end simpler and more extensible.

bgklika: IP addresses are passed as shell style program arguments.
Double quote prevents one but with spaces argument like "first argument" to be accepted as two separated arguments by the program. At the same time list of IP passed without quotas because passed as separated arguments. 
JSON useless here.

{configuration:/controlPort} is just reference to Greengrass component configuration value.
From scenario side these are settings of deployment passed as JSON doc.

uat/custom-components/client-java-sdk/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCLibImpl.java
bgklika: Will replace.

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
MikeDombo: If you need to add any other arguments, then that becomes difficult since you are trying to handle the last _n_ items as IP address information.

Using named environment variables and/or JSON to pass the configuration would be less errorprone and require less manual work to parse the inputs.

bgklika: That is usual impediments for arguments of the CLI programs.
It is solved by using option signs like "-a IP -a IP -a IP"  or "--address IP --address IP" or  end-options-mark like "-- IP IP IP IP ".
But that structure of CLI arguments should be the same over all MQTT clients. When using JSON we need to add a JSON library to each client. For example C/C++ completely absent global packages repo like maven central that means with each new dependency you more and more limit your software to be build and run only on selected versions of OSes. Has no sence to inject JSON library only to pass list of string to the program.

Windows missed the default library to parse command line short or long options, that was a reason why arguments were extremely simplified.

MikeDombo: Using environment variables would also simplify this.

`CLIENT_ID=client id`, `IP_ADDRESSES="<list of ip space separated>"`, ... other config options.

bgklika: Environment variables are the implicit dependency, arguments - explicit.
In that case we need explicit dependency.
Complexity of passing value via environment is the same as aruments.
No pros.

MikeDombo (approved):

uat/testing-features/src/main/resources/greengrass/components/recipes/client_java_sdk.yaml
bgklika: shell already did parsing

MikeDombo: Shell didn't parse anything. You are parsing in the `main` function. You are assigning meaning to the inputs and validating them.

bgklika: main() and parseArgs()  just move input arguments from the array to the output list https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/uat-dev/uat/custom-components/client-java-sdk/src/main/java/com/aws/greengrass/testing/mqtt5/client/Main.java#L80

to `void main(String[] args)` IP addresses came as args[2], args[3], args[4] separately.
Double quote is magic here.

uat/custom-components/client-java-sdk/src/main/java/com/aws/greengrass/testing/mqtt5/client/grpc/GRPCLibImpl.java
bgklika: Because PR already merged changes made in separate branch GGMQ-84-multi-homed-control-is-not-supported.ADDON

bgklika: See new PR https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/263

PR #278, opened 2023-05-05
feat(uat): docker and local build of mosquitto client

bgklika:
**Issue #, if available:**
Docker build of Mosquitto client

**Description of changes:**
- Updated cmake files of the CPP project
- Added docker file
- Replaced Empty in proto and update code due to google proto files not found
- Updated README to actual build commands

We need stable local and docker build of mosquitto client

At the moment manually, later can add github CI build or docker build in Code Build.


uat/custom-components/client-mosquitto-c/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:22.04 AS builder
```
MikeDombo, line 1: please use an image from ECR public gallery and not dockerhub: https://gallery.ecr.aws/

jcosentino11 (approved):
Lgtm (aside from michaels comment)

uat/custom-components/client-mosquitto-c/Dockerfile
bgklika: When replaced to public.ecr.aws/lts/ubuntu:22.04_stable I got an error in git clone:
```
Cloning into 'grpc-src'...
fatal: unable to access 'https://github.com/grpc/grpc/': server certificate verification failed. CAfile: none CRLfile: none
```
Looks like that image lack the trusted CA list.
Do we need install CAs manually ?

MikeDombo: Try using amazonlinux.

Sounds like that image may have outdated CA certs perhaps.

bgklika: Image ubuntu:22.04_stable  completely missing root CAs  I thing due to package was not installed at all. Solution is reinstall ca-certificates package. Now testing that solution, build takes a time due to gRPC + protobuf + all requirements added as submodules are building.

amazonlinux2 is centOS based distro, it potentially can cause a problem because github uses ubuntu 22.04 on default runners

bgklika: Resolved in 8505bb92b8d5fabed2f7ba05324d44540b95af8d

MikeDombo (approved):

PR #293, opened 2023-05-23
fix(uat): fix assetion server shutdown

esmirnov-klika-tech:
**Description of changes:**
- add null validation
- rename method

- to prevent uat test failtures


uat/testing-features/src/main/java/com/aws/greengrass/steps/AssertionSteps.java
```diff
@@ -13,6 +13,7 @@
 import io.cucumber.java.en.When;
 import lombok.extern.log4j.Log4j2;
 
+import java.util.Optional;
```
MikeDombo, line 16: Remove import.

MikeDombo (approved):

jcosentino11 (approved):

PR #339, opened 2023-06-23
feat: utility method for policy variable support

cyyan88:
**Issue #, if available:**

**Description of changes:**

- Add updateResource method to allow Permissions to be updated with policy variable value
- Error support for updateResource: add to Session + SessionImplementation to allow check if attribute/namespace exists
- Unit test for updateResource

Required for implementation of CDA policy variables.

- 3 unit tests for updateResource method
- Passes current integration tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -165,7 +165,6 @@ public final class PermissionEvaluationUtils {
      * @param permission permission to parse
      * @return updated permission
      */
-
     static Permission updateResource(@NonNull Session session, @NonNull Permission permission) {
 
         String resource = permission.getResource();
@@ -180,15 +179,13 @@ public final class PermissionEvaluationUtils {
             String attributeName = vars[2];
 
             // this supports the ThingName attribute only
-            if (THING_NAME_VARIABLE.equals(policyVariable)
-                    && session.containsSessionAttribute(attributeNamespace, attributeName)) {
-
+            if (THING_NAME_VARIABLE.equals(policyVariable)) {
                 String policyVariableValue =
                         Coerce.toString(session.getSessionAttribute(attributeNamespace, attributeName));
 
                 if (policyVariableValue == null) {
                     logger.atWarn().kv("attributeName", attributeName)
-                            .log("No attribute found for current session.");
+                            .log("No attribute found for current session");
                 } else {
                     resource = matcher.replaceFirst(policyVariableValue);
 
@@ -197,14 +194,11 @@ public final class PermissionEvaluationUtils {
                             .operation(permission.getOperation())
                             .resource(resource).build();
                 }
-
-
             } else {
                 logger.atWarn().kv("policyVariable", policyVariable)
-                        .log("Policy variable detected but could not be parsed.");
+                        .log("Policy variable detected but could not be parsed");
             }
         }
-
         return permission;
     }
 
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Utils;
 import lombok.Builder;
+import lombok.NonNull;
 import lombok.Value;
 
 import java.util.Map;
@@ -167,7 +168,7 @@ public final class PermissionEvaluationUtils {
      * @return updated permission
      */
 
-    public static Permission updateResource(Session session, Permission permission) {
+    static Permission updateResource(@NonNull Session session, @NonNull Permission permission) {
 
         String resource = permission.getResource();
 
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -35,9 +35,10 @@ public final class PermissionEvaluationUtils {
             String.format(SERVICE_RESOURCE_FORMAT, SERVICE_PATTERN_STRING, SERVICE_RESOURCE_TYPE_PATTERN_STRING,
                     SERVICE_RESOURCE_NAME_PATTERN_STRING), Pattern.UNICODE_CHARACTER_CLASS);
 
-    private static final String POLICY_VARIABLE_PATTERN_STRING = "\\$\\{iot:(\\w*+.\\w*+.\\w*+)}";
+    private static final String POLICY_VARIABLE_PATTERN = "\\$\\{iot:(Connection.Thing.ThingName)}";
 
-    private static final String PROVIDER_STRING = "Connection";
+
+    private static final String PROVIDER = "Connection";
 
     private static final Logger LOGGER = LogManager.getLogger(SessionConfig.class);
 
@@ -165,11 +166,12 @@ public final class PermissionEvaluationUtils {
      * @param permission              permission to parse
      * @return updated permission
      */
+
     public static Permission updateResource(Session session, Permission permission) {
 
         String resource = permission.getResource();
 
-        Pattern pattern = Pattern.compile(POLICY_VARIABLE_PATTERN_STRING);
+        Pattern pattern = Pattern.compile(POLICY_VARIABLE_PATTERN);
         Matcher matcher = pattern.matcher(resource);
 
         while (matcher.find()) { 
@@ -179,7 +181,7 @@ public final class PermissionEvaluationUtils {
             String attributeName = vars[2];
 
             // this currently supports the ThingName attribute
-            if (vars[0].equals(PROVIDER_STRING)
+            if (vars[0].equals(PROVIDER)
                     && session.containsSessionAttribute(attributeNamespace, attributeName)) {
 
                 String policyVariableValue = session.getSessionAttribute(attributeNamespace, attributeName).toString();
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -6,6 +6,8 @@
 package com.aws.greengrass.clientdevices.auth;
 
 import com.aws.greengrass.clientdevices.auth.configuration.Permission;
+import com.aws.greengrass.clientdevices.auth.session.Session;
+import com.aws.greengrass.clientdevices.auth.session.SessionConfig;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Utils;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -139,14 +148,57 @@ private static Resource parseResource(String resourceStr) {
...
+    /**
+     * utility method to permission with variable value.
+     *
+     * @param session             current device session
+     * @param permission              permission to parse
+     * @return updated permission
+     */
+    public static Permission updateResource(Session session, Permission permission) {
+
+        String resource = permission.getResource();
```
MikeDombo, line 170: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem:**
 This line of code lacks validation when processing input data through the following parameter: &#39;permission&#39; (index: 1 | type: Permission). The parameter is exposed to external callers, because its enclosing class and method are publicly accessible. This means that upstream validation, if it exists, can be bypassed. Other validated parameters: &#39;session&#39;. Malicious, malformed, or unbounded inputs can cause unexpected runtime behavior or crashes, and can slow performance.

**Fix:**
 Add checks to ensure the validity of the parameter&#39;s value, such as testing it for nullness (for example, using the @NonNull annotation as [described in the Lombok library](https://projectlombok.org/features/NonNull)), emptiness, or equality. Or to prevent direct calls to it, [reduce the method&#39;s visibility](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html).

 Learn more about potential threats and guidance from the [Common Weakness Enumeration website](https://cwe.mitre.org/data/definitions/693.html) and the [OWASP Cheat Sheet series](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html).

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionImpl.java
```diff
@@ -1,48 +1,78 @@
...
 
-package com.aws.greengrass.clientdevices.auth.session;
+ package com.aws.greengrass.clientdevices.auth.session;
 
-import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
-import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
+ import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
+ import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
+ 
+ import java.util.concurrent.ConcurrentHashMap;
```
jcosentino11, line 11: looks like most lines have an extra space at the start, let's fix the formatting so the diff shows nicely in github

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,6 +35,12 @@ public final class PermissionEvaluationUtils {
             String.format(SERVICE_RESOURCE_FORMAT, SERVICE_PATTERN_STRING, SERVICE_RESOURCE_TYPE_PATTERN_STRING,
                     SERVICE_RESOURCE_NAME_PATTERN_STRING), Pattern.UNICODE_CHARACTER_CLASS);
 
+    private static final String POLICY_VARIABLE_PATTERN_STRING = "\\$\\{iot:(\\w*+.\\w*+.\\w*+)}";
```
jcosentino11, line 38:
```suggestion
    private static final String POLICY_VARIABLE_PATTERN = "\\$\\{iot:(\\w*+.\\w*+.\\w*+)}";
```
style nit: string is redundant

MikeDombo:
@jcosentino11 let's make a new branch for this work instead of merging to main.

jcosentino11:
rebased to `policy-variable-support` branch, @cyyan88 let's start cutting PRs to this branch now

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
jcosentino11: can we just match on `${iot:Connection.Thing.ThingName}` rather than matching on \w?  When we need to support other variables we can revisit the approach, can keep it very simple for now

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionImpl.java
```diff
@@ -12,37 +12,66 @@
...
+  * Check if attribute exists.
+  *
+  * @param attributeNamespace Attribute namespace
+  * @param attributeName      Attribute name
+  * @return True/false
+  */
+ @Override
+ public boolean containsSessionAttribute(String attributeNamespace, String attributeName) {
+     return containsAttributeProvider(attributeNamespace)
+                 && this.getAttributeProvider(attributeNamespace)
```
jcosentino11, line 72: i think can get away w/o adding new `containsAttributeProvider`.  If it doesn't exist `getAttributeProvider` would be null

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -139,14 +150,58 @@ private static Resource parseResource(String resourceStr) {
...
+        Matcher matcher = pattern.matcher(resource);
+
+        while (matcher.find()) { 
+
+            String[] vars = matcher.group(1).split("\\.");
+            String attributeNamespace = vars[1];
+            String attributeName = vars[2];
+
+            // this currently supports the ThingName attribute
+            if (vars[0].equals(PROVIDER)
```
jcosentino11, line 185: since we explicitly match on thing name we don't need to check provider anymore, we know it's `Connection.Thing.ThingName`

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionImpl.java
```diff
@@ -45,4 +45,33 @@ public DeviceAttribute getSessionAttribute(String attributeNamespace, String att
...
+     * Check if attribute exists.
+     *
+     * @param attributeNamespace Attribute namespace
+     * @param attributeName      Attribute name
+     * @return True/false
+     */
+    @Override
+    public boolean containsSessionAttribute(String attributeNamespace, String attributeName) {
+        return containsAttributeProvider(attributeNamespace)
+                && this.getAttributeProvider(attributeNamespace)
```
MikeDombo, line 72: don't do a "contains" check and then do a "get". Instead just do a get and handle the null case. With this current implementation there's a race after checking contains

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -139,14 +150,58 @@ private static Resource parseResource(String resourceStr) {
...
+     * @param session    current device session
+     * @param permission permission to parse
+     * @return updated permission
+     */
+
+    static Permission updateResource(@NonNull Session session, @NonNull Permission permission) {
+
+        String resource = permission.getResource();
+
+        Pattern pattern = Pattern.compile(POLICY_VARIABLE_PATTERN);
```
MikeDombo, line 175: Store the compiled pattern as a constant field. Compiling the pattern every time this method is called is not necessary.

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,6 +36,13 @@ public final class PermissionEvaluationUtils {
             String.format(SERVICE_RESOURCE_FORMAT, SERVICE_PATTERN_STRING, SERVICE_RESOURCE_TYPE_PATTERN_STRING,
                     SERVICE_RESOURCE_NAME_PATTERN_STRING), Pattern.UNICODE_CHARACTER_CLASS);
 
+    private static final String POLICY_VARIABLE_PATTERN = "\\$\\{iot:(Connection.Thing.ThingName)}";
+
+
+    private static final String PROVIDER = "Connection";
+
+    private static final Logger LOGGER = LogManager.getLogger(SessionConfig.class);
```
MikeDombo, line 44: a logger already exists, use the logger we already have.

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -139,14 +150,58 @@ private static Resource parseResource(String resourceStr) {
...
+
+            String[] vars = matcher.group(1).split("\\.");
+            String attributeNamespace = vars[1];
+            String attributeName = vars[2];
+
+            // this currently supports the ThingName attribute
+            if (vars[0].equals(PROVIDER)
+                    && session.containsSessionAttribute(attributeNamespace, attributeName)) {
+
+                String policyVariableValue = session.getSessionAttribute(attributeNamespace, attributeName).toString();
```
MikeDombo, line 188: getSessionAttribute could be null, so don't call `toString` which would throw a null pointer. Make sure to handle the null case.

Using `Coerce.toString` would be safe.

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,6 +36,11 @@ public final class PermissionEvaluationUtils {
             String.format(SERVICE_RESOURCE_FORMAT, SERVICE_PATTERN_STRING, SERVICE_RESOURCE_TYPE_PATTERN_STRING,
                     SERVICE_RESOURCE_NAME_PATTERN_STRING), Pattern.UNICODE_CHARACTER_CLASS);
 
+    private static final String POLICY_VARIABLE_FORMAT = "\\$\\{iot:(Connection.Thing.ThingName)}";
```
MikeDombo, line 39: shouldn't this "format" be more generic?

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -139,14 +148,66 @@ private static Resource parseResource(String resourceStr) {
...
+
+        while (matcher.find()) {
+
+            String policyVariable = matcher.group(1);
+            String[] vars = policyVariable.split("\\.");
+            String attributeNamespace = vars[1];
+            String attributeName = vars[2];
+
+            // this currently supports the ThingName attribute
+            if (policyVariable.equals(THING_NAME_VARIABLE)
```
MikeDombo, line 183: flip this equals around to `THING_NAME_VARIABLE.equals(policyVariable)` that way if policy variable is somehow null, this wouldn't error.

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
cyyan88: I originally had it match all variables in the xx.Namespace.Attribute format (with `"\\$\\{iot:(\\w*+.\\w*+.\\w*+)}"`). Joe suggested I only match on Connection.Thing.ThingName for now to keep it simple. Should I change it back?

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/session/SessionImpl.java
```diff
@@ -45,4 +45,24 @@ public DeviceAttribute getSessionAttribute(String attributeNamespace, String att
...
+     * Check if attribute exists.
+     *
+     * @param attributeNameSpace Attribute namespace
+     * @param attributeName      Attribute name
+     * @return True/false
+     */
+    @Override
+    public boolean containsSessionAttribute(String attributeNameSpace, String attributeName) {
+
+        if (!this.containsKey(attributeNameSpace)) {
```
jcosentino11, line 59: there's still a race between check and get here, as well as in `updateResource`.  i think we can remove the contains method and just use `getSessionAttribute` in `updateResource`

jcosentino11 (approved):

PR #344, opened 2023-06-26
feat: utility method for updating permissions for device

cyyan88:
**Issue #, if available:**

**Description of changes:**

- Add getDevicePolicyPermissions method to update device group permissions with permissions specific to the current device. These updates are based on the device's policy variables
- Unit test for getDevicePolicyPermissions

Required for implementation of CDA policy variables.

- 1 unit test for getDevicePolicyPermissions method
- Passes current integration tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -165,7 +165,6 @@ public final class PermissionEvaluationUtils {
      * @param permission permission to parse
      * @return updated permission
      */
-
     static Permission updateResource(@NonNull Session session, @NonNull Permission permission) {
 
         String resource = permission.getResource();
@@ -180,15 +179,13 @@ public final class PermissionEvaluationUtils {
             String attributeName = vars[2];
 
             // this supports the ThingName attribute only
-            if (THING_NAME_VARIABLE.equals(policyVariable)
-                    && session.containsSessionAttribute(attributeNamespace, attributeName)) {
-
+            if (THING_NAME_VARIABLE.equals(policyVariable)) {
                 String policyVariableValue =
                         Coerce.toString(session.getSessionAttribute(attributeNamespace, attributeName));
 
                 if (policyVariableValue == null) {
                     logger.atWarn().kv("attributeName", attributeName)
-                            .log("No attribute found for current session.");
+                            .log("No attribute found for current session");
                 } else {
                     resource = matcher.replaceFirst(policyVariableValue);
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -97,6 +98,25 @@ public static boolean isAuthorized(String operation, String resource,
...
     }
 
+    /**
+     * utility method to permission with variable value.
+     *
+     * @param session    current device session
+     * @param groupToPermissionsMap set of permissions for each device group
+     * @return permission map with updated resources
+     */
+    public static Map<String, Set<Permission>> getDevicePolicyPermissions(
```
jcosentino11, line 108: I think we can adjust the naming to make the intent clearer---this isn't doing a get, but rather a map or transform, on the input.    

I neglected to mention on the last PR, but same with `updateResource`, it's doing variable replacement, so the name should reflect that

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -97,6 +98,25 @@ public static boolean isAuthorized(String operation, String resource,
         return false;
     }
 
+    /**
+     * utility method to permission with variable value.
```
jcosentino11, line 102: could we provide more description/context here? this doesn't make sense to me

jcosentino11:
also looks like commit-lint is failing, make sure all commits in your branch follow conventional commit format

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -97,6 +98,25 @@ public static boolean isAuthorized(String operation, String resource,
...
     }
 
+    /**
+     * utility method to transform map of group permissions by updating policy variables with device variable values.
+     *
+     * @param session    current device session
+     * @param groupToPermissionsMap set of permissions for each device group
+     * @return permission map with updated resources
+     */
+    public static Map<String, Set<Permission>> transformGroupPermissionsWithVariableValue(
```
jcosentino11, line 108:
```suggestion
    public static Map<String, Set<Permission>> resolveVariablesInResources(
```
suggestion

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -159,13 +179,13 @@ private static Resource parseResource(String resourceStr) {
...
     /**
-     * utility method to permission with variable value.
+     * utility method to replace policy variables in permission with device attribute.
      *
      * @param session    current device session
      * @param permission permission to parse
      * @return updated permission
      */
-    static Permission updateResource(@NonNull Session session, @NonNull Permission permission) {
+    static Permission replaceResourcePolicyVariable(@NonNull Session session, @NonNull Permission permission) {
```
jcosentino11, line 188: suggestion: `resolveVariablesInResource`

jcosentino11 (approved):

PR #346, opened 2023-06-27
fix: add support for replacing multiple policy variables

cyyan88:
**Issue #, if available:**

**Description of changes:**
- Change implementation of replaceResourcePolicyVariable to fix infinite loop when multiple policy variables are in a single resource.
- Add unit tests to cover cases where a resource has multiple policy variables.

Previously, test cases only covered cases where a resource had one policy variable. Introducing a test case for multiple policy variables exposed an error, which was fixed in these code changes.

- 2 unit tests
- Passes all integration tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -165,7 +165,6 @@ public final class PermissionEvaluationUtils {
      * @param permission permission to parse
      * @return updated permission
      */
-
     static Permission updateResource(@NonNull Session session, @NonNull Permission permission) {
 
         String resource = permission.getResource();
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -187,8 +206,10 @@ static Permission updateResource(@NonNull Session session, @NonNull Permission p
                     logger.atWarn().kv("attributeName", attributeName)
                             .log("No attribute found for current session");
                 } else {
+                    // for ThingName support only, we can use .replaceAll()
+                    // to support additional policy variables in the future
                     resource = matcher.replaceFirst(policyVariableValue);
-
+                    matcher = POLICY_VARIABLE_PATTERN.matcher(resource);
                     permission = Permission.builder()
```
jcosentino11, line 213: style: Let's avoid reassigning argument variables, add a new local variable

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -187,8 +206,10 @@ static Permission updateResource(@NonNull Session session, @NonNull Permission p
                     logger.atWarn().kv("attributeName", attributeName)
                             .log("No attribute found for current session");
                 } else {
+                    // for ThingName support only, we can use .replaceAll()
+                    // to support additional policy variables in the future
                     resource = matcher.replaceFirst(policyVariableValue);
-
+                    matcher = POLICY_VARIABLE_PATTERN.matcher(resource);
```
jcosentino11, line 212: This works, but curious is there's a way to iterate over matches and replace without creating a new matcher every time?

cyyan88: I think it's possible but it would require having a new variable for the updated resource bc of the way `matcher` works. `matcher` only has `replaceAll` (which would be fine for only using the ThingName variable) and `replaceFirst` (which is problematic because `matcher` will keep replacing the first instance of the policy variable in every iteration since it was instantiated with the original resource). With a new variable we would have to keep track of indices to append the value + parts of the resource that don't match, not sure if that's preferred (or if we should just `replaceAll` to only account for ThingName for now).

jcosentino11 (approved):

PR #360, opened 2023-07-05
feat(uat): add T20 scenario and new connection step

auarbekov:
**Issue #, if available:**
added t20 scenario and added new connection step

**Description of changes:**
- Add T20 scenario 
- Add new failed connection step

to implement Mqtt 5.0 features

Scenarios run manually and on CI

**Test results:**
```
Given my device is registered as a Thing....................................passed
And my device is running Greengrass.........................................passed
When I create a Greengrass deployment with components.......................passed
And I create client device "clientDeviceTest"...............................passed
When I associate "clientDeviceTest" with ggc................................passed
And I update my Greengrass deployment configuration, setting the component aws.greengrass.clientdevices.Auth configuration to:.passed
And I update my Greengrass deployment configuration, setting the component aws.greengrass.client.Mqtt5JavaSdkClient configuration to:.passed
And I deploy the Greengrass deployment configuration........................passed
Then the Greengrass deployment is COMPLETED on the device after 5 minutes...passed
And the aws.greengrass.clientdevices.mqtt.EMQX log on the device contains the line "is running now!." within 1 minutes.passed
And I discover core device broker as "default_broker" from "clientDeviceTest" in OTF.passed
And I connect device "clientDeviceTest" on aws.greengrass.client.Mqtt5JavaSdkClient to "default_broker" using mqtt "v5".passed
When I create a Greengrass deployment with components.......................passed
And I deploy the Greengrass deployment configuration........................passed
Then the Greengrass deployment is COMPLETED on the device after 299 seconds.passed
And I discover core device broker as "default_broker" from "clientDeviceTest" in OTF.passed
And I force to set broker "default_broker" with port 9443...................passed
Then I wait 60 seconds......................................................passed
And I can not connect device "clientDeviceTest" on aws.greengrass.client.Mqtt5JavaSdkClient to "default_broker" using mqtt "v5".passed
```

### DIFF
--- a/uat/testing-features/src/main/java/com/aws/greengrass/steps/MqttControlSteps.java
+++ b/uat/testing-features/src/main/java/com/aws/greengrass/steps/MqttControlSteps.java
@@ -10,7 +10,6 @@ import com.aws.greengrass.testing.model.RegistrationContext;
 import com.aws.greengrass.testing.model.ScenarioContext;
 import com.aws.greengrass.testing.model.TestContext;
 import com.aws.greengrass.testing.modules.model.AWSResourcesContext;
-import com.aws.greengrass.testing.mqtt.client.Mqtt5ConnAck;
 import com.aws.greengrass.testing.mqtt.client.Mqtt5Disconnect;
 import com.aws.greengrass.testing.mqtt.client.Mqtt5Message;
 import com.aws.greengrass.testing.mqtt.client.Mqtt5Properties;
@@ -737,7 +736,7 @@ public class MqttControlSteps {
      */
     @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.UseObjectForClearerAPI"})
     @And("I connect device {string} on {word} to {string} using mqtt {string}")
-    public Mqtt5ConnAck connect(String clientDeviceId, String componentId, String brokerId, String mqttVersion) {
+    public void connect(String clientDeviceId, String componentId, String brokerId, String mqttVersion) {
 
         // get address information about broker
         final List<MqttBrokerConnectionInfo> bc = brokers.get(brokerId);
@@ -771,7 +770,7 @@ public class MqttControlSteps {
                 log.info("Connection with broker {} established to address {}:{} as Thing {} on {}",
                          brokerId, host, port, clientDeviceThingName, componentId);
                 setConnectionControl(connectionControl, clientDeviceThingName);
-                return connectionControl.getConnAck();
+                return;
             } catch (RuntimeException ex) {
                 lastException = ex;
             }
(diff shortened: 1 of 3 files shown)


uat/testing-features/src/main/java/com/aws/greengrass/steps/MqttControlSteps.java
```diff
@@ -820,6 +843,25 @@ public void discoverCoreDeviceBroker(String brokerId) {
...
     }
 
+
+    /**
+     * Set up IoT core broker port.
+     *
+     * @param brokerId broker name in tests
+     * @param port broker port in tests
+     */
+    @And("I force to set broker {string} with port {int}")
```
MikeDombo, line 853: why is this needed?

MikeDombo: Why isn't this using discovery like our other tests? Our internal testing does not do this. It adds the proper connectivity information with localhost and the custom port.

bgklika: It that scenario we check case when client can connect to custom MQTT port AND can't connect to wrong port. For that we use discovery as usually but after that rewrite port in discovered list of  address information  to wrong port. In Evergreen similar thing did by other way.

MikeDombo: I'm looking at GGAD-1-T20 right now. It does not do this, it is changing the discovery information. What step do you claim is doing this?

bgklika: That inject invalid connectivity information for non-yet actual port number 9001 which resulting second client failed to connect on port 9001.

With OTF we have no step like ''add connectivity information with host "localhost" and port "9001" we use IPdetector instead and it can't send fake port to backend.

MikeDombo: And why don't you have step for that?

bgklika: We can try to implement  by reconfigure IPDetector but I expecting rice between components and scenario. 
We already got a lot of such issues and usually resolved it by adding steps for waiting until string appears in log file. With IPDetector it be hardly possible because IPDetector report action to log several times. Because we can stuck on other waiting step it hard to predict how many times IPDetector send information to backend and  how many entries about it will be logged.

bgklika: We just combine both methods.

MikeDombo: Please update this as suggested to set the connectivity info.

bgklika: Ok, Akezhan will continue on Monday due to national holidays.

MikeDombo (commented):
Is this ready for review?

bgklika:
Under internal review, I do not sure it is ready for your review. Probably close to finish.

uat/testing-features/src/main/java/com/aws/greengrass/steps/MqttControlSteps.java
bgklika: Fixed in latest commits

MikeDombo (approved):

PR #404, opened 2023-09-27
feat: certificate generation for client and server type

junfuchen99:
**Description of changes:**
Add supports for generating custom certificates from other plugins, which include
1. An new type `CLIENT_AND_SERVER` in `CertificateType`;
2. A new request type `CustomGeneratorCertificateRequest` which extends the existing request and adds a field for providing custom certificate generator
3. Extended `Certificate Manager` to handle custom requests;
4. Refactored `CertificateManager` to reduce duplicate codes/improve readability
5. Added unit test.

To allow plugin components request certificate generations from CDA.

1. Added unit test.
2. Manual tested with Cluster Controller component.


src/main/java/com/aws/greengrass/clientdevices/auth/certificate/CertificateHelper.java
```diff
@@ -149,6 +149,35 @@ public static X509Certificate createCACertificate(@NonNull KeyPair keyPair, @Non
...
+     * @param notBefore             Certificate notBefore Date
+     * @param notAfter              Certificate notAfter Date
+     * @param providerType          Type of provider
+     * @return X509 certificate
+     * @throws NoSuchAlgorithmException  NoSuchAlgorithmException
+     * @throws OperatorCreationException OperatorCreationException
+     * @throws CertificateException      CertificateException
+     * @throws IOException               IOException
+     */
+    public static X509Certificate issueClusterCertificate(@NonNull X509Certificate caCert,
```
MikeDombo, line 169: why does CDA need to be aware of "cluster"?

MikeDombo (commented):
Why is all this cluster specific stuff in here? I thought you were just going to open up some of the core APIs to allow cluster to generate what it needs.

jcosentino11:
> Why is all this cluster specific stuff in here? I thought you were just going to open up some of the core APIs to allow cluster to generate what it needs.

+1   i see there's different cert validity for cluster, is that a requirement?   if so we could include validity as part of CDA API instead

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -167,100 +171,102 @@ public String getCaPassPhrase() {
...
-
-        // Add certificate generator to monitors first in order to avoid missing events
-        // that happen while the initial certificate is being generated.
+        // Add certificate generator to monitors
         certExpiryMonitor.addToMonitor(certificateGenerator);
-        cisShadowMonitor.addToMonitor(certificateGenerator);
         caConfigurationMonitor.addToMonitor(certificateGenerator);
 
+        if (type == GetCertificateRequestOptions.CertificateType.SERVER
+                || type == GetCertificateRequestOptions.CertificateType.CUSTOM) {
```
MikeDombo, line 235: not sure about adding a custom type. Really what you want is client+server right? I think that adding that would be more clear in that it is the combination of both of what we have today.

junfuchen99: yes I can rename to ClientAndServer. Passing a `CertificateGenerator` in the request also allows the caller to controll the expiry time, SANs, and other configs, so I think that bit of refactoring is still needed.

jcosentino11: Do we need a new type at all? Would it make sense to modify ClientDeviceAuthApi to accept a list of cert types in the request and all the other customizations (SAN, expiry, etc)?

junfuchen99: I'm trying to keep backwards compatibility so I do not want to change anything in the original `GetCertificateRequest`.

Conceptually, I think today there is an 1:1 mapping between cert type and certificate. Both `CertificateSubscriptionEvent` and `GetCertificateRequestOptions` relies on singleton cert type, which is why I prefer to not use a list of cert types.

jcosentino11: For my understanding how does this break compatibility if we just modify the api class but not the ipc handler?

junfuchen99: If we modify the API class input type, old versions of other components (e.g. moquette server [here](https://github.com/aws-greengrass/aws-greengrass-moquette-mqtt/blob/10d8671b4622a1c388c0f2ad43f3a3f521e084e9/integration/src/main/java/com/aws/greengrass/mqtt/moquette/MQTTService.java#L106)) would not be able to call `subscribeToCertificateUpdates`.

This can be addressed by bumping dependency version range on CDA but I would like to avoid it unless necessary.

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -167,100 +171,102 @@ public String getCaPassPhrase() {
...
+            case CLIENT:
+                // TODO: Should be configurable
+                KeyPair keyPair = CertificateStore.newRSAKeyPair(4096);
+
+                BiConsumer<X509Certificate, X509Certificate[]> consumer = (cert, caCertificates) -> {
+                    CertificateUpdateEvent certificateUpdateEvent =
+                            new CertificateUpdateEvent(keyPair, cert, caCertificates);
+                    request.getCertificateUpdateConsumer().accept(certificateUpdateEvent);
+                };
+                if (type.equals(GetCertificateRequestOptions.CertificateType.SERVER)) {
```
MikeDombo, line 208: flip equals

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
```diff
@@ -167,100 +171,102 @@ public String getCaPassPhrase() {
...
         certSubscriptions.compute(certificateRequest, (k, v) -> {
-            // A subscription already exists, we will replace it so that a new certificate is generated immediately
             if (v != null) {
                 removeCGFromMonitors(v);
             }
             return certificateGenerator;
         });
     }
 
-    private void subscribeToClientCertificateUpdatesNoCSR(@NonNull GetCertificateRequest certificateRequest,
```
MikeDombo, line 241: why is this all gone?

src/main/java/com/aws/greengrass/clientdevices/auth/CertificateManager.java
junfuchen99: updated

MikeDombo (approved):

PR #405, opened 2023-10-23
fix(cis): retry get shadow operation

jcosentino11:
**Issue #, if available:**

**Description of changes:**

On startup, CIS shadow monitor requests the latest CIS shadow via an MQTT request, and will update certificates with host information accordingly.  However, there is a subtle race where nucleus' mqtt client may reset during this CIS shadow get operation (e.g. due to device configuration changing).

This PR adds retry logic when fetching CIS shadows, to ensure we are resilient to failures.  We can simply wait until we receive a get shadow response, which is either:
* get shadow accepted
* get shadow rejected

Accepted is the happy path. We requested a shadow and we received it.  Rejected means the shadow doesn't exist yet; there's no need to retry because we'll get a delta update when the shadow is eventually created.

If get shadow fails when CDA starts up, server certificates may not be populated with the proper SANs, causing TLS connections between clients and local broker to fail.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -79,6 +79,9 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
                     .retryableExceptions(Collections.singletonList(Exception.class))
                     .build();
 
+    // we only need to subscribe to shadow topics once across the lifetime of this monitor
+    private final AtomicBoolean subscribed = new AtomicBoolean();
+
     // we don't need to unsubscribe, because mqtt client reconnects with clean session on startup.
     // instead, we can cancel current operations and block new ones on shutdown.
     private final AtomicBoolean stopped = new AtomicBoolean();
@@ -104,13 +107,13 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
 
     private final Object getShadowLock = new Object();
     private Future<?> getShadowTask;
-    AtomicReference<CompletableFuture<?>> shadowGetResponseReceived = new AtomicReference<>();
+    AtomicReference<CompletableFuture<?>> getShadowResponseReceived = new AtomicReference<>();
+
+    private String lastVersion;
 
     private final Supplier<Integer> mqttOperationTimeoutMillis;
     private MqttClientConnection connection;
     private IotShadowClient iotShadowClient;
-    private String lastVersion;
-    private final AtomicBoolean subscribed = new AtomicBoolean();
     private final NetworkStateProvider networkStateProvider;
     private final List<CertificateGenerator> monitoredCertificateGenerators = new CopyOnWriteArrayList<>();
     private final ExecutorService executorService;
@@ -337,7 +340,7 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
                             GET_CIS_SHADOW_RETRY_CONFIG,
                             () -> {
                                 CompletableFuture<?> shadowGetResponseReceived = new CompletableFuture<>();
-                                this.shadowGetResponseReceived.set(shadowGetResponseReceived);
+                                this.getShadowResponseReceived.set(shadowGetResponseReceived);
                                 publishToGetCISShadowTopic().get();
                                 // await shadow get accepted, rejected
                                 long waitForGetResponseTimeout =
@@ -407,7 +410,7 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
     }
 
     private void reportShadowReceived() {
-        CompletableFuture<?> shadowReceived = this.shadowGetResponseReceived.get();
+        CompletableFuture<?> shadowReceived = this.getShadowResponseReceived.get();
         if (shadowReceived != null) {
             shadowReceived.complete(null);
         }
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -14,7 +14,9 @@ import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.mqttclient.WrapperMqttClientConnection;
 import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.CrashableSupplier;
 import com.aws.greengrass.util.RetryUtils;
+import lombok.NonNull;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnection;
 import software.amazon.awssdk.crt.mqtt.QualityOfService;
 import software.amazon.awssdk.iot.iotshadow.IotShadowClient;
@@ -43,11 +45,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
@@ -79,9 +79,6 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
                     .retryableExceptions(Collections.singletonList(Exception.class))
                     .build();
 
-    // we only need to subscribe to shadow topics once across the lifetime of this monitor
-    private final AtomicBoolean subscribed = new AtomicBoolean();
-
     private final Consumer<ShadowDeltaUpdatedEvent> onShadowDeltaUpdated = this::processCISShadow;
     private final Consumer<GetShadowResponse> onGetShadowAccepted = resp -> {
         signalShadowResponseReceived();
@@ -101,8 +98,49 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
     private final NetworkStateProvider networkStateProvider;
     private final List<CertificateGenerator> monitoredCertificateGenerators = new CopyOnWriteArrayList<>();
     private final ExecutorService executorService;
-    private final String shadowName;
+    private final AtomicReference<String> shadowName = new AtomicReference<>();
     private final ConnectivityInformation connectivityInformation;
+
+    private final SucceedOnceOperation subscribeToShadowUpdateDelta = new SucceedOnceOperation(() -> {
+        ShadowDeltaUpdatedSubscriptionRequest shadowDeltaUpdatedSubscriptionRequest =
+                new ShadowDeltaUpdatedSubscriptionRequest();
+        shadowDeltaUpdatedSubscriptionRequest.thingName = shadowName.get();
+        iotShadowClient.SubscribeToShadowDeltaUpdatedEvents(
+                        shadowDeltaUpdatedSubscriptionRequest,
+                        QualityOfService.AT_LEAST_ONCE,
+                        onShadowDeltaUpdated,
+                        (e) -> LOGGER.atError()
+                                .log("Error processing shadowDeltaUpdatedSubscription Response", e))
+                .get();
+        LOGGER.atDebug().log("Subscribed to shadow update delta topic");
+        return null;
+    });
+
+    private final SucceedOnceOperation subscribeToShadowGetAccepted = new SucceedOnceOperation(() -> {
+        GetShadowSubscriptionRequest getShadowSubscriptionRequest = new GetShadowSubscriptionRequest();
+        getShadowSubscriptionRequest.thingName = shadowName.get();
+        iotShadowClient.SubscribeToGetShadowAccepted(
+                        getShadowSubscriptionRequest,
+                        QualityOfService.AT_LEAST_ONCE,
+                        onGetShadowAccepted,
+                        (e) -> LOGGER.atError().log("Error processing getShadowSubscription Response", e))
+                .get();
+        LOGGER.atDebug().log("Subscribed to shadow get accepted topic");
+        return null;
+    });
+
+    private final SucceedOnceOperation subscribeToShadowGetRejected = new SucceedOnceOperation(() -> {
+        GetShadowSubscriptionRequest getShadowRejectedSubscriptionRequest = new GetShadowSubscriptionRequest();
+        getShadowRejectedSubscriptionRequest.thingName = shadowName.get();
+        iotShadowClient.SubscribeToGetShadowRejected(
+                        getShadowRejectedSubscriptionRequest,
+                        QualityOfService.AT_LEAST_ONCE,
+                        onGetShadowRejected,
+                        (e) -> LOGGER.atError().log("Error processing get shadow rejected response", e))
+                .get();
+        LOGGER.atDebug().log("Subscribed to shadow get rejected topic");
+        return null;
+    });
 
     /**
      * Constructor.
@@ -143,7 +181,7 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
         this.connection = connection;
         this.iotShadowClient = iotShadowClient;
         this.executorService = executorService;
-        this.shadowName = shadowName;
+        this.shadowName.set(shadowName);
         this.connectivityInformation = connectivityInformation;
         this.mqttOperationTimeoutMillis = mqttOperationTimeoutMillis;
     }
@@ -160,7 +198,6 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
      * Stop shadow monitor.
      */
     public void stopMonitor() {
-        subscribed.set(false);
         cancelFetchCISShadow();
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -67,6 +71,21 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
...
+            RetryUtils.RetryConfig.builder()
+                    .initialRetryInterval(Duration.ofSeconds(1L))
+                    .maxRetryInterval(Duration.ofMinutes(30L))
+                    .maxAttempt(Integer.MAX_VALUE)
+                    .retryableExceptions(Collections.singletonList(Exception.class))
+                    .build();
+
+    @Getter(AccessLevel.PACKAGE)
+    @Setter(AccessLevel.PACKAGE)
+    private long mqttOperationTimeoutSeconds = 10L;
```
MikeDombo, line 84: get the timeout from the mqtt client. mqttClient.getOperationTimeout or something like that

MikeDombo: In fact, our MQTT client already has a timeout; so I'm not sure we actually need our own timeout here, it may be fine to just do a blocking `get`

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -158,21 +177,37 @@ private void subscribeToShadowTopics() throws InterruptedException {
                 ShadowDeltaUpdatedSubscriptionRequest shadowDeltaUpdatedSubscriptionRequest =
                         new ShadowDeltaUpdatedSubscriptionRequest();
                 shadowDeltaUpdatedSubscriptionRequest.thingName = shadowName;
-                iotShadowClient.SubscribeToShadowDeltaUpdatedEvents(shadowDeltaUpdatedSubscriptionRequest,
+                iotShadowClient.SubscribeToShadowDeltaUpdatedEvents(
+                                shadowDeltaUpdatedSubscriptionRequest,
                                 QualityOfService.AT_LEAST_ONCE,
                                 this::processCISShadow,
```
MikeDombo, line 183: store the callback as a final field to prevent recreating lambdas.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -158,21 +177,37 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
                 getShadowSubscriptionRequest.thingName = shadowName;
-                iotShadowClient.SubscribeToGetShadowAccepted(getShadowSubscriptionRequest,
-                                QualityOfService.AT_LEAST_ONCE, this::processCISShadow,
+                iotShadowClient.SubscribeToGetShadowAccepted(
+                                getShadowSubscriptionRequest,
+                                QualityOfService.AT_LEAST_ONCE,
+                                resp -> {
+                                    reportShadowReceived();
+                                    processCISShadow(resp);
+                                },
```
MikeDombo, line 197: store callback as a final field to prevent recreating lambdas which may lead to duplicate calls on message receipt

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -158,21 +177,37 @@ private void subscribeToShadowTopics() throws InterruptedException {
...
-                        .get(TIMEOUT_FOR_SUBSCRIBING_TO_TOPICS_SECONDS, TimeUnit.SECONDS);
+                        .get(mqttOperationTimeoutSeconds, TimeUnit.SECONDS);
                 LOGGER.info("Subscribed to shadow get accepted topic");
+
+                GetShadowSubscriptionRequest getShadowRejectedSubscriptionRequest = new GetShadowSubscriptionRequest();
+                getShadowRejectedSubscriptionRequest.thingName = shadowName;
+                iotShadowClient.SubscribeToGetShadowRejected(
+                                getShadowRejectedSubscriptionRequest,
+                                QualityOfService.AT_LEAST_ONCE,
+                                err -> reportShadowReceived(),
```
MikeDombo, line 207: same.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -315,14 +351,60 @@ private void updateCISShadowReportedState(Map<String, Object> reportedState) {
...
+                // operation already in progress
+                return;
+            }
+            getShadowTask = executorService.submit(() -> {
+                try {
+                    RetryUtils.runWithRetry(
+                            GET_CIS_SHADOW_RETRY_CONFIG,
+                            () -> {
+                                CompletableFuture<?> shadowReceived = this.shadowGetResponseReceived.updateAndGet(
+                                        ignore -> new CompletableFuture<>());
```
MikeDombo, line 374: just do a set instead of update?

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -315,14 +351,60 @@ private void updateCISShadowReportedState(Map<String, Object> reportedState) {
...
+
+    private CompletableFuture<Integer> publishToGetCISShadowTopic() {
         LOGGER.atDebug().log("Publishing to get shadow topic");
         GetShadowRequest getShadowRequest = new GetShadowRequest();
         getShadowRequest.thingName = shadowName;
-        iotShadowClient.PublishGetShadow(getShadowRequest, QualityOfService.AT_LEAST_ONCE).exceptionally(e -> {
-            LOGGER.atWarn().cause(e).log("Unable to retrieve CIS shadow");
-            return null;
-        });
+        return iotShadowClient.PublishGetShadow(getShadowRequest, QualityOfService.AT_LEAST_ONCE)
```
MikeDombo, line 403: use qos 0. Otherwise when we're offline this is going to fill up the spooler.

you may also want to not retry when we know that we're offline.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -315,14 +351,60 @@ private void updateCISShadowReportedState(Map<String, Object> reportedState) {
...
+        CompletableFuture<?> shadowReceived = this.shadowGetResponseReceived.get();
+        if (shadowReceived != null) {
+            shadowReceived.complete(null);
+        }
+    }
+
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private void fetchCISShadowWithRetriesAsync() {
+        synchronized (getShadowLock) {
+            if (getShadowTask != null && !getShadowTask.isDone()) {
```
MikeDombo, line 364: check for cancelled? or does done include cancelled

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: ah yeah, will adjust

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: good point. and yeah when cda goes offline the task will be cancelled, but i didn't handle case if we start offline will fix

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: done includes cancelled

jcosentino11:
comments addressed, still need to update the test, had to step out for a little bit

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -33,45 +33,70 @@
...
-    private static final long TIMEOUT_FOR_SUBSCRIBING_TO_TOPICS_SECONDS = Duration.ofMinutes(1).getSeconds();
-    private static final long WAIT_TIME_TO_SUBSCRIBE_AGAIN_IN_MS = Duration.ofMinutes(2).toMillis();
-    private static final Random JITTER = new Random();
     static final String SHADOW_UPDATE_DELTA_TOPIC = "$aws/things/%s/shadow/update/delta";
     static final String SHADOW_GET_ACCEPTED_TOPIC = "$aws/things/%s/shadow/get/accepted";
 
+    private static final RetryUtils.RetryConfig SUBSCRIBE_TO_TOPICS_RETRY_CONFIG =
+            RetryUtils.RetryConfig.builder()
+                    .initialRetryInterval(Duration.ofSeconds(1L))
+                    .maxRetryInterval(Duration.ofMinutes(30L))
```
MikeDombo, line 66: 30 minutes? Seems a bit much, no?

If we're only trying when we know that we're online, then retrying faster should be alright.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -149,49 +208,38 @@ public void removeFromMonitor(CertificateGenerator certificateGenerator) {
...
-                LOGGER.atWarn().setCause(e).log("Subscribe to shadow topics timed out, will retry shortly");
-            }
-            // Wait for sometime and then try to subscribe again
-            Thread.sleep(WAIT_TIME_TO_SUBSCRIBE_AGAIN_IN_MS + JITTER.nextInt(10_000));
-        }
+    private void subscribeToShadowTopics() throws InterruptedException, ExecutionException {
+        ShadowDeltaUpdatedSubscriptionRequest shadowDeltaUpdatedSubscriptionRequest =
+                new ShadowDeltaUpdatedSubscriptionRequest();
+        shadowDeltaUpdatedSubscriptionRequest.thingName = shadowName;
+        iotShadowClient.SubscribeToShadowDeltaUpdatedEvents(
```
MikeDombo, line 215: let's do it in a different PR, but we should stop using the shadow client. It dynamically creates a new message handler every time it is called which will lead to duplicate subscription callbacks.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -149,49 +208,38 @@ public void removeFromMonitor(CertificateGenerator certificateGenerator) {
...
+
+        GetShadowSubscriptionRequest getShadowSubscriptionRequest = new GetShadowSubscriptionRequest();
+        getShadowSubscriptionRequest.thingName = shadowName;
+        iotShadowClient.SubscribeToGetShadowAccepted(
+                        getShadowSubscriptionRequest,
+                        QualityOfService.AT_LEAST_ONCE,
+                        onGetShadowAccepted,
+                        (e) -> LOGGER.atError().log("Error processing getShadowSubscription Response", e))
+                .get();
+        LOGGER.atInfo().log("Subscribed to shadow get accepted topic");
```
MikeDombo, line 232: debug

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: in this case it's retrying even when offline, but could probably be smarter about this, will revise

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: ah geez, yep will follow up w/ a fix

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: fixed. also consolidated subscribing future and get future into a single future (making subscribe only happen once).  also removed unsubscribing since it's unnecessary

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -33,45 +33,88 @@
...
+                    .maxRetryInterval(Duration.ofSeconds(30L))
+                    .maxAttempt(Integer.MAX_VALUE)
+                    .retryableExceptions(Collections.singletonList(Exception.class))
+                    .build();
+
+    // we only need to subscribe to shadow topics once across the lifetime of this monitor
+    private final AtomicBoolean subscribed = new AtomicBoolean();
+
+    // we don't need to unsubscribe, because mqtt client reconnects with clean session on startup.
+    // instead, we can cancel current operations and block new ones on shutdown.
```
MikeDombo, line 86: on reconnect though, the subscriptions which were successful will be re-subscribed by the mqtt client for you.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -33,45 +33,88 @@
...
+    private final AtomicBoolean stopped = new AtomicBoolean();
+    private final Consumer<ShadowDeltaUpdatedEvent> onShadowDeltaUpdated = event -> {
+        if (stopped.get()) {
+            return;
+        }
+        processCISShadow(event);
```
MikeDombo, line 92: if we got an event, why would we ignore it?

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -33,45 +33,88 @@
...
+    };
+    private final Consumer<GetShadowResponse> onGetShadowAccepted = resp -> {
+        if (stopped.get()) {
+            return;
+        }
+        reportShadowReceived();
+        processCISShadow(resp);
```
MikeDombo, line 99: why would we ignore any response?

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -33,45 +33,88 @@
...
+    };
+    private final Consumer<ErrorResponse> onGetShadowRejected = err -> {
+        if (stopped.get()) {
+            return;
+        }
+        reportShadowReceived();
```
MikeDombo, line 105: report shadow received? But this is rejected...

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: this was to handle shutdown w/o unsubscribe

jcosentino11: will add unsubscribe back in, looks to be less confusing---was originally thinking on the terms of not having extra mqtt calls if after shutdown we'd lose the subscription anyway (since nucleus first-time connect is w/ clean session)

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: bad naming, it's to signal that we've received any shadow response, whether accepted or rejected, will think of better name

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
MikeDombo: If we're shutting down, why does it matter then if we process the event or not?

jcosentino11: Was to avoid extra processing if we receive a shadow update after the task was already cancelled. 

I guess it's not a big deal though because nucleus will shutdown the executors eventually

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
jcosentino11: yep, this was talking about shutdown and startup.  but just to confirm my thinking, unsubscribe on shutdown is technically unnecessary right?

MikeDombo: Yes, unsubscribing on shutdown shouldn't really matter.

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -81,54 +108,60 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
...
 
     /**
      * Stop shadow monitor.
      */
     public void stopMonitor() {
-        if (subscribeTaskFuture != null) {
-            subscribeTaskFuture.cancel(true);
-        }
-        unsubscribeFromShadowTopics();
+        subscribed.set(false);
```
MikeDombo, line 163: why set it to false? You aren't ever unsubscribing, so you are still subscribed.

jcosentino11: revised the subscription stuff

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -33,102 +35,169 @@
...
+                    .maxRetryInterval(Duration.ofSeconds(30L))
+                    .maxAttempt(Integer.MAX_VALUE)
+                    .retryableExceptions(Collections.singletonList(Exception.class))
+                    .build();
     private static final RetryUtils.RetryConfig GET_CONNECTIVITY_RETRY_CONFIG =
             RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofMinutes(1L))
                     .maxRetryInterval(Duration.ofMinutes(30L)).maxAttempt(Integer.MAX_VALUE)
                     .retryableExceptions(Arrays.asList(ThrottlingException.class, InternalServerException.class))
                     .build();
+    private static final RetryUtils.RetryConfig GET_CIS_SHADOW_RETRY_CONFIG =
```
sameerzuberi, line 74: nit: do we need to separate the retry configs here? Looks like `SUBSCRIBE_TO_TOPICS` and `GET_CONNECTIVITY` are the same, could be combined in a single variable

vabhasin (approved):

sameerzuberi (approved):

sameerzuberi (approved):

vabhasin (approved):

PR #409, opened 2023-10-31
chore: cleanup exception messaging

jcosentino11:
**Issue #, if available:**

**Description of changes:**

* Have more specific exception messages in `CreateIoTThingSession` to disambiguate failures.
* Make `CloudServiceInteractionException` a checked exception

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
@@ -41,13 +41,13 @@ import java.util.stream.Stream;
 import javax.inject.Inject;
 
 public interface IotAuthClient {
-    Optional<String> getActiveCertificateId(String certificatePem);
+    Optional<String> getActiveCertificateId(String certificatePem) throws CloudServiceInteractionException;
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/iot/IotAuthClient.java
```diff
@@ -41,13 +41,13 @@
 import javax.inject.Inject;
 
 public interface IotAuthClient {
-    Optional<String> getActiveCertificateId(String certificatePem);
+    Optional<String> getActiveCertificateId(String certificatePem) throws CloudServiceInteractionException;
```
MikeDombo, line 44: this breaks out interface doesn't it? Isn't moquette using these methods? If it is, then this is going to break consumers.

jcosentino11: consumers just interact with `ClientDevicesAuthServiceApi` and `AuthenticationException`s from what i can see, so that should be fine?

robcmann (approved):

MikeDombo (approved):

PR #410, opened 2023-11-06
chore(cis): concurrency fixes and cert gen retries

jcosentino11:
**Issue #, if available:**

**Description of changes:**

This PR modifies `CISShadowMonitor` to be resilient to race conditions, such as:
1) receiving shadow delta response around the same time as a get shadow response, during startup up or after network connectivity resumes
2) receiving multiple delta responses in quick succession, due to connectivity shadow being updated quickly

...and the chances increase if the core device is low-powered (takes longer to generate certificates).

Currently, there is no mechanism in place to prevent the interleaving connectivity shadow processing (which includes fetching/updating connectivity info, and generating certificates using such info).  Worst case, this can result in server certificates being generated with old/previous host addresses.

To achieve this, I'm using a producer-consumer approach. Producers are incoming shadow requests (get shadow, or shadow delta update).  Consumer is a single thread that processes one request from the queue at a time.  Incoming requests are filtered out if they are deemed stale (previous or duplicate shadow version, duplicate CIS shadow state).  This approach is more complicated than canceling and replacing task with every incoming shadow request, but it ensures the monitor doesn't perform extra work.

Additionally, retry behavior has been modified so that calls to connectivity service and certificate generation are retried indefinitely, until interruption.  Previously, on failures there were no retries so there was no way for recovery, aside from restarting greengrass or waiting/forcing a shadow update.

:warning: Suggestion to reviewers: view diff in split mode

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -455,6 +455,12 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
             if (Objects.equals(task1, task2)) {
                 return 0;
             }
+            if (task1 == null) {
+                return -1;
+            }
+            if (task2 == null) {
+                return 1;
+            }
 
             // previous or duplicate shadow version
             if (task1.shadowVersion <= task2.shadowVersion) {
@@ -485,7 +491,7 @@ public class CISShadowMonitor implements Consumer<NetworkStateProvider.Connectio
 
         private boolean isNewTask(ProcessCISShadowTask task) {
             return Stream.of(this.tasks.peekLast(), lastProcessedTask)
-                    .allMatch(t -> t == null || taskComparator.compare(task, t) > 0);
+                    .allMatch(t -> taskComparator.compare(task, t) > 0);
         }
 
         synchronized void removeProcessedTask(ProcessCISShadowTask task) {
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
@@ -6,7 +6,6 @@
 package com.aws.greengrass.clientdevices.auth.connectivity;
 
 import com.aws.greengrass.clientdevices.auth.certificate.CertificateGenerator;
-import com.aws.greengrass.clientdevices.auth.exception.CertificateGenerationException;
 import com.aws.greengrass.clientdevices.auth.infra.NetworkStateProvider;
 import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.logging.api.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -29,62 +31,61 @@
...
-                    .build();
     private static final RetryUtils.RetryConfig ALWAYS_RETRY_CONFIG =
             RetryUtils.RetryConfig.builder()
                     .initialRetryInterval(Duration.ofSeconds(1L))
                     .maxRetryInterval(Duration.ofSeconds(30L))
                     .maxAttempt(Integer.MAX_VALUE)
                     .retryableExceptions(Collections.singletonList(Exception.class))
                     .build();
+    private final CISShadowTaskExecutor taskExecutor = new CISShadowTaskExecutor();
+    private final CISShadowTaskQueue taskQueue = new CISShadowTaskQueue();
```
MikeDombo, line 74: do we actually need a queue? Shouldn't we be just using the very latest update all the time?

jcosentino11: yeah, it's effectively a queue of size 1, and requests can only be added if it's for the latest update.  the reason for using a queue is to make request processing truly serial

i considered doing the usual approach of `future.cancel(); executor.submit()` for each request received, but the issue was that with interrupts we could still have overlapping tasks (new one starts while other one, although interrupted, is still finishing execution), and tasks update state of "last processed shadow", so didn't want that to interleave

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -434,6 +310,210 @@ public void accept(NetworkStateProvider.ConnectionState state) {
...
+    }
+
+    /**
+     * A queue for connectivity shadow responses. Old shadows will be rejected from the queue.
+     */
+    class CISShadowTaskQueue  {
+        private final BlockingDeque<ProcessCISShadowTask> tasks = new LinkedBlockingDeque<>();
+        private ProcessCISShadowTask lastProcessedTask;
+        private final Comparator<ProcessCISShadowTask> taskComparator = (task1, task2) -> {
+            if (Objects.equals(task1, task2)) {
```
MikeDombo, line 455: Make sure to handle both null tasks and null versions within those tasks.

jcosentino11: right now null tasks are filtered out before comparing in `isNewTask`:
```
t -> t == null || taskComparator.compare(task, t) > 0
```
could move it in to the comparator so it's more resilient.
---
null cis versions are already handled through Objects.equals. and shadow versions are just `long`

jcosentino11: done

src/main/java/com/aws/greengrass/clientdevices/auth/connectivity/CISShadowMonitor.java
```diff
@@ -29,62 +31,61 @@
...
                     .maxAttempt(Integer.MAX_VALUE)
                     .retryableExceptions(Collections.singletonList(Exception.class))
                     .build();
+    private final CISShadowTaskExecutor taskExecutor = new CISShadowTaskExecutor();
+    private final CISShadowTaskQueue taskQueue = new CISShadowTaskQueue();
 
-    private final Consumer<ShadowDeltaUpdatedEvent> onShadowDeltaUpdated = this::processCISShadow;
+    private final Consumer<ShadowDeltaUpdatedEvent> onShadowDeltaUpdated = resp ->
+            taskQueue.offer(new ProcessCISShadowTask(
+                    resp.version, Coerce.toString(resp.state.get("version")), resp.state));
```
MikeDombo, line 78: resp.state may be null. resp.state.get("version") definitely can be null.

jcosentino11: ah right, was thinking shadow state would always be correct, will fix

jcosentino11: fixed

robcmann (approved):

MikeDombo (approved):

PR #415, opened 2024-01-05
feat: add policy variable support

sameerzuberi:
**Issue #, if available:**

**Description of changes:**

- Adding policy variable support for thing names.
- Updating `PermissionEvaluationUtils` to use non static methods for future wildcard support

Unit and integration tests

### DIFF
--- a/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
+++ b/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
@@ -7,6 +7,7 @@ package com.aws.greengrass.clientdevices.auth.benchmark;
 
 import com.aws.greengrass.clientdevices.auth.AuthorizationRequest;
 import com.aws.greengrass.clientdevices.auth.DeviceAuthClient;
+import com.aws.greengrass.clientdevices.auth.PermissionEvaluationUtils;
 import com.aws.greengrass.clientdevices.auth.configuration.AuthorizationPolicyStatement;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupConfiguration;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupDefinition;
--- a/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
+++ b/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.benchmark;
+
+import com.aws.greengrass.clientdevices.auth.AuthorizationRequest;
+import com.aws.greengrass.clientdevices.auth.DeviceAuthClient;
+import com.aws.greengrass.clientdevices.auth.configuration.AuthorizationPolicyStatement;
+import com.aws.greengrass.clientdevices.auth.configuration.GroupConfiguration;
+import com.aws.greengrass.clientdevices.auth.configuration.GroupDefinition;
+import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
+import com.aws.greengrass.clientdevices.auth.configuration.parser.ParseException;
+import com.aws.greengrass.clientdevices.auth.exception.AuthorizationException;
+import com.aws.greengrass.clientdevices.auth.session.Session;
+import com.aws.greengrass.clientdevices.auth.session.SessionManager;
+import com.aws.greengrass.clientdevices.auth.session.attribute.AttributeProvider;
+import com.aws.greengrass.clientdevices.auth.session.attribute.DeviceAttribute;
+import com.aws.greengrass.clientdevices.auth.session.attribute.StringLiteralAttribute;
+import com.aws.greengrass.clientdevices.auth.session.attribute.WildcardSuffixAttribute;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Warmup(iterations = 3)
+@Measurement(iterations = 5)
+public class AuthorizationBenchmarks {
+
+    @State(Scope.Thread)
+    public static class SimpleAuthRequest extends PolicyTestState {
+
+        final AuthorizationRequest basicRequest = AuthorizationRequest.builder()
+                .operation("mqtt:publish")
+                .resource("mqtt:topic:humidity")
+                .sessionId("sessionId")
+                .build();
+
+        @Setup
+        public void doSetup() throws ParseException, AuthorizationException {
+            sessionManager.registerSession("sessionId", FakeSession.forDevice("MyThingName"));
+            groupManager.setGroupConfiguration(GroupConfiguration.builder()
+                    .definitions(Collections.singletonMap(
+                            "group1", GroupDefinition.builder()
+                                    .selectionRule("thingName: " + "MyThingName")
+                                    .policyName("policy1")
+                                    .build()))
+                    .policies(Collections.singletonMap(
+                            "policy1", Collections.singletonMap(
+                                    "Statement1", AuthorizationPolicyStatement.builder()
+                                                    .statementDescription("Policy description")
+                                                    .effect(AuthorizationPolicyStatement.Effect.ALLOW)
+                                                    .resources(new HashSet<>(Collections.singleton("mqtt:topic:humidity")))
+                                                    .operations(new HashSet<>(Collections.singleton("mqtt:publish")))
+                                                    .build())))
+                    .build());
+        }
+    }
+
+    @Benchmark
+    public boolean GIVEN_single_group_permission_WHEN_simple_auth_request_THEN_successful_auth(SimpleAuthRequest state) throws Exception {
+        return state.deviceAuthClient.canDevicePerform(state.basicRequest);
+    }
+
+    static abstract class PolicyTestState {
+        final FakeSessionManager sessionManager = new FakeSessionManager();
+        final GroupManager groupManager = new GroupManager();
+        final DeviceAuthClient deviceAuthClient = new DeviceAuthClient(sessionManager, groupManager, null);
+    }
+
+    static class FakeSession implements Session {
+        private final String thingName;
+        private final boolean isComponent;
+
+        static FakeSession forComponent() {
+            return new FakeSession(null, true);
+        }
+
+        static FakeSession forDevice(String thingName) {
+            return new FakeSession(thingName, false);
+        }
+
+        private FakeSession(String thingName, boolean isComponent) {
+            this.thingName = thingName;
+            this.isComponent = isComponent;
+        }
+
+        @Override
+        public AttributeProvider getAttributeProvider(String attributeProviderNameSpace) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public DeviceAttribute getSessionAttribute(String ns, String name) {
+            if ("Component".equalsIgnoreCase(ns) && name.equalsIgnoreCase("component")) {
+                return isComponent ? new StringLiteralAttribute("component") : null;
+            }
+            if ("Thing".equalsIgnoreCase(ns) && name.equalsIgnoreCase("thingName")) {
+                   return new WildcardSuffixAttribute(thingName);
+            }
+            throw new UnsupportedOperationException(String.format("Attribute %s.%s not supported", ns, name));
+        }
+    }
+
+    private static class FakeSessionManager extends SessionManager {
+        private final Map<String, Session> sessions = new HashMap<>();
+
+        public FakeSessionManager() {
+            super(null);
+        }
+
+        void registerSession(String id, Session session) {
+            sessions.put(id, session);
+        }
+
+        @Override
+        public Session findSession(String id) {
+            return sessions.get(id);
+        }
+    }
+}
--- a/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
+++ b/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
@@ -80,7 +80,8 @@ public class AuthorizationBenchmarks {
     static abstract class PolicyTestState {
         final FakeSessionManager sessionManager = new FakeSessionManager();
         final GroupManager groupManager = new GroupManager();
-        final DeviceAuthClient deviceAuthClient = new DeviceAuthClient(sessionManager, groupManager, null);
+        final PermissionEvaluationUtils permissionEvaluationUtils = new PermissionEvaluationUtils(groupManager);
+        final DeviceAuthClient deviceAuthClient = new DeviceAuthClient(sessionManager, null, permissionEvaluationUtils);
     }
 
     static class FakeSession implements Session {
--- a/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
+++ b/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
@@ -30,7 +30,6 @@ import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.Warmup;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
--- a/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
+++ b/benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
@@ -30,6 +30,7 @@ import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.Warmup;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -73,9 +74,44 @@ public class AuthorizationBenchmarks {
         }
     }
 
+    @State(Scope.Thread)
+    public static class PolicyVariableAuthRequest extends PolicyTestState {
+
+        final AuthorizationRequest thingNameRequest = AuthorizationRequest.builder()
+                .operation("mqtt:publish")
+                .resource("mqtt:topic:MyThingName/humidity")
+                .sessionId("sessionId")
+                .build();
+
+        @Setup
+        public void doSetup() throws ParseException, AuthorizationException {
+            sessionManager.registerSession("sessionId", FakeSession.forDevice("MyThingName"));
+            groupManager.setGroupConfiguration(GroupConfiguration.builder()
+                    .definitions(Collections.singletonMap(
+                            "group1", GroupDefinition.builder()
+                                    .selectionRule("thingName: " + "MyThingName")
+                                    .policyName("policy1")
+                                    .build()))
+                    .policies(Collections.singletonMap(
+                            "policy1", Collections.singletonMap(
+                                    "Statement1", AuthorizationPolicyStatement.builder()
+                                            .statementDescription("Policy description")
+                                            .effect(AuthorizationPolicyStatement.Effect.ALLOW)
+                                            .resources(new HashSet<>(Collections.singleton("mqtt:topic:${iot:Connection.Thing.ThingName}/humidity")))
+                                            .operations(new HashSet<>(Collections.singleton("mqtt:publish")))
+                                            .build())))
+                    .build());
+        }
+    }
+
     @Benchmark
     public boolean GIVEN_single_group_permission_WHEN_simple_auth_request_THEN_successful_auth(SimpleAuthRequest state) throws Exception {
         return state.deviceAuthClient.canDevicePerform(state.basicRequest);
+    }
+
+    @Benchmark
+    public boolean GIVEN_policy_with_thing_name_variable_WHEN_auth_request_THEN_successful_auth(PolicyVariableAuthRequest state) throws Exception {
+        return state.deviceAuthClient.canDevicePerform(state.thingNameRequest);
     }
 
     static abstract class PolicyTestState {
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/deviceauth/EvaluateClientDeviceActionsWithPolicyVariablesTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/deviceauth/EvaluateClientDeviceActionsWithPolicyVariablesTest.java
@@ -1,0 +1,242 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.integrationtests.deviceauth;
+
+import com.aws.greengrass.clientdevices.auth.ClientDevicesAuthService;
+import com.aws.greengrass.clientdevices.auth.api.ClientDevicesAuthServiceApi;
+import com.aws.greengrass.clientdevices.auth.certificate.CertificateHelper;
+import com.aws.greengrass.clientdevices.auth.helpers.CertificateTestHelpers;
+import com.aws.greengrass.clientdevices.auth.iot.Certificate;
+import com.aws.greengrass.clientdevices.auth.iot.CertificateRegistry;
+import com.aws.greengrass.clientdevices.auth.iot.IotAuthClient;
+import com.aws.greengrass.clientdevices.auth.iot.IotAuthClientFake;
+import com.aws.greengrass.clientdevices.auth.iot.Thing;
+import com.aws.greengrass.clientdevices.auth.iot.infra.ThingRegistry;
+import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.integrationtests.ipc.IPCTestUtils;
+import com.aws.greengrass.lifecyclemanager.GlobalStateChangeListener;
+import com.aws.greengrass.lifecyclemanager.GreengrassService;
+import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.logging.impl.config.LogConfig;
+import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
+import com.aws.greengrass.testcommons.testutilities.GGExtension;
+import com.aws.greengrass.testcommons.testutilities.UniqueRootPathExtension;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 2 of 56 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/DeviceAuthClient.java
```diff
@@ -141,7 +144,8 @@ public boolean canDevicePerform(AuthorizationRequest request) throws Authorizati
             return true;
         }
 
-        return PermissionEvaluationUtils.isAuthorized(request.getOperation(), request.getResource(),
-                groupManager.getApplicablePolicyPermissions(session));
+        return permissionEvaluationUtils.isAuthorized(request.getOperation(), request.getResource(),
+                permissionEvaluationUtils.transformGroupPermissionsWithVariableValue(session,
```
jcosentino11, line 148: while we're at it, we can clean up by making groupManager a dependency of PermissionEvaluationUtils, then we can do something like

```
permissionEvaluationUtils.isAuthorized(request, session)
```

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -60,7 +66,8 @@ public static boolean isAuthorized(String operation, String resource,
             return false;
         }
 
-        for (Map.Entry<String, Set<Permission>> entry : groupToPermissionsMap.entrySet()) {
+        for (Map.Entry<String, Set<Permission>> entry :
+                groupToPermissionsMap.entrySet()) {
```
jcosentino11, line 70: nit: do we need this formatting change?

src/main/java/com/aws/greengrass/clientdevices/auth/DeviceAuthClient.java
jcosentino11: lets also take performance into account, e.g. when evaluating w/ same permissions we shouldn't do variable substitution again

sameerzuberi: done

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
sameerzuberi: fixed

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,34 +39,56 @@ public final class PermissionEvaluationUtils {
...
+     * @param request   Authorization Request
+     * @param session   Session
+     *
+     * @return boolean indicating if the operation requested is authorized
      */
-    public static boolean isAuthorized(String operation, String resource,
-                                       Map<String, Set<Permission>> groupToPermissionsMap) {
-        Operation op = parseOperation(operation);
-        Resource rsc = parseResource(resource);
+    public boolean isAuthorized(AuthorizationRequest request, Session session) {
```
MikeDombo, line 69: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 12. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 86-114 into a separate method.

src/main/java/com/aws/greengrass/clientdevices/auth/DeviceAuthClient.java
jcosentino11: > lets also take performance into account, e.g. when evaluating w/ same permissions we shouldn't do variable substitution again

Thinking through this more, variables are substituted based on current session, so a caching solution for this would be a bit more involved, assuming it would even be useful, so lets table for now

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +179,68 @@ private static Operation parseOperation(String operationStr) {
...
+                    resource = matcher.replaceFirst(policyVariableValue);
+                    matcher = POLICY_VARIABLE_PATTERN.matcher(resource);
+                    permission = Permission.builder()
+                            .principal(permission.getPrincipal())
+                            .operation(permission.getOperation())
+                            .resource(resource).build();
+                }
+            } else {
+                logger.atWarn().kv("policyVariable", policyVariable)
+                        .log("Policy variable detected but could not be parsed");
```
jcosentino11, line 238: policy variable `unsupported`,  and can also guide user w/ next steps (e.g. fix config)

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +179,68 @@ private static Operation parseOperation(String operationStr) {
...
+        Matcher matcher = POLICY_VARIABLE_PATTERN.matcher(resource);
+
+        while (matcher.find()) {
+            String policyVariable = matcher.group(1);
+            String[] vars = policyVariable.split("\\.");
+            String attributeNamespace = vars[1];
+            String attributeName = vars[2];
+
+            // this supports the ThingName attribute only
+            if (THING_NAME_VARIABLE.equals(policyVariable)) {
```
jcosentino11, line 219: i think we'd want to be case insensitive, would need to configure the matcher too

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +179,68 @@ private static Operation parseOperation(String operationStr) {
...
+                String policyVariableValue =
+                        Coerce.toString(session.getSessionAttribute(attributeNamespace, attributeName));
+
+                if (policyVariableValue == null) {
+                    logger.atWarn().kv("attributeName", attributeName)
```
jcosentino11, line 224: i think we'd want to error here rather than log, it's a failure mode--not doing substitution would mean the policy is acting differently than intended

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +174,68 @@ private static Operation parseOperation(String operationStr) {
...
+            String attributeNamespace = vars[1];
+            String attributeName = vars[2];
+
+            // this supports the ThingName attribute only
+            if (THING_NAME_VARIABLE.equalsIgnoreCase(policyVariable)) {
+                String policyVariableValue =
+                        Coerce.toString(session.getSessionAttribute(attributeNamespace, attributeName));
+
+                if (policyVariableValue == null) {
+                    throw new IllegalArgumentException("Not attribute found for current session");
```
robcmann, line 219: Nit: typo

sameerzuberi: ah good catch fixed

robcmann (dismissed):

src/integrationtests/resources/com/aws/greengrass/integrationtests/deviceauth/config.yaml
```diff
@@ -0,0 +1,72 @@
...
+            policyName: "thingAccessPolicy"
+        policies:
+          thingAccessPolicy:
+            policyStatement1:
+              statementDescription: "mqtt connect"
+              effect: ALLOW
+              operations:
+                - "mqtt:connect"
+              resources:
+                - "mqtt:${iot:Connection.Thing.ThingName}:foo"
```
jcosentino11, line 26: i don't think we should allow substitution here, just in the third part of the resource

MikeDombo: Isn't this the primary use case @jcosentino11?

MikeDombo: While I agree this doesn't make practical sense (since it can only be filled with `mqtt:topic`, `mqtt:clientId`), is it actually reasonable for us to prevent it? Or would preventing it actually be more difficult

benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
```diff
@@ -80,7 +81,8 @@ public boolean GIVEN_single_group_permission_WHEN_simple_auth_request_THEN_succe
     static abstract class PolicyTestState {
         final FakeSessionManager sessionManager = new FakeSessionManager();
         final GroupManager groupManager = new GroupManager();
-        final DeviceAuthClient deviceAuthClient = new DeviceAuthClient(sessionManager, groupManager, null);
+        final PermissionEvaluationUtils permissionEvaluationUtils = new PermissionEvaluationUtils(groupManager);
+        final DeviceAuthClient deviceAuthClient = new DeviceAuthClient(sessionManager, null, permissionEvaluationUtils);
```
MikeDombo, line 85: make sure we're benchmarking utilizing variables.

src/integrationtests/resources/com/aws/greengrass/integrationtests/deviceauth/config.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - BrokerWithAuthorizeClientDeviceActionPermission
+  BrokerWithGetClientDeviceAuthTokenPermission:
+    dependencies:
+      - aws.greengrass.clientdevices.Auth
+    lifecycle:
+      run:
+        windows:
+          powershell -command sleep 1
+        posix:
+          sleep 1
```
MikeDombo, line 45: remove

src/integrationtests/resources/com/aws/greengrass/integrationtests/deviceauth/config.yaml
```diff
@@ -0,0 +1,72 @@
...
+              - '*'
+  BrokerWithAuthorizeClientDeviceActionPermission:
+    dependencies:
+      - aws.greengrass.clientdevices.Auth
+    lifecycle:
+      run:
+        windows:
+          powershell -command sleep 1
+        posix:
+          sleep 1
```
MikeDombo, line 63: remove

src/integrationtests/java/com/aws/greengrass/integrationtests/deviceauth/EvaluateClientDeviceActionsWithPolicyVariablesTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+            authzClientDeviceAction(ipcClient, connectRequest, false);
+            authzClientDeviceAction(ipcClient, publishRequest, false);
+        }
+    }
+
+    @Test
+    void GIVEN_permissiveGroupPolicyWithThingNameVariable_WHEN_ClientAuthorizesWithInvalidThingNameResource_THEN_ClientNotAuthorized() throws Exception {
+
+        try (EventStreamRPCConnection connection = IPCTestUtils.getEventStreamRpcConnection(kernel,
```
MikeDombo, line 224: why are we using IPC here? Can't we just call the internal API?

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,29 +39,46 @@ public final class PermissionEvaluationUtils {
             String.format(SERVICE_RESOURCE_FORMAT, SERVICE_PATTERN_STRING, SERVICE_RESOURCE_TYPE_PATTERN_STRING,
                     SERVICE_RESOURCE_NAME_PATTERN_STRING), Pattern.UNICODE_CHARACTER_CLASS);
 
+    private static final String POLICY_VARIABLE_FORMAT = "\\$\\{iot:(Connection.Thing.ThingName)}";
```
MikeDombo, line 42: not much of a format if we're only supporting literally 1 replacement.

MikeDombo: Can we make this more generic?

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +174,68 @@ private static Operation parseOperation(String operationStr) {
...
+
+        String resource = permission.getResource();
+
+        Matcher matcher = POLICY_VARIABLE_PATTERN.matcher(resource);
+
+        while (matcher.find()) {
+            String policyVariable = matcher.group(1);
+            String[] vars = policyVariable.split("\\.");
+            String attributeNamespace = vars[1];
+            String attributeName = vars[2];
```
MikeDombo, line 211: check for length before indexing

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +174,68 @@ private static Operation parseOperation(String operationStr) {
...
+
+            // this supports the ThingName attribute only
+            if (THING_NAME_VARIABLE.equalsIgnoreCase(policyVariable)) {
+                String policyVariableValue =
+                        Coerce.toString(session.getSessionAttribute(attributeNamespace, attributeName));
+
+                if (policyVariableValue == null) {
+                    throw new IllegalArgumentException("No attribute found for current session");
```
MikeDombo, line 219: who is handling this exception?

src/integrationtests/resources/com/aws/greengrass/integrationtests/deviceauth/config.yaml
jcosentino11: > since it can only be filled with

i don't think we enforce this though?  the above policy right now is considered valid and auth passes, (`mqtt:myThing:foo` in this case)

src/integrationtests/java/com/aws/greengrass/integrationtests/deviceauth/EvaluateClientDeviceActionsWithPolicyVariablesTest.java
jcosentino11: true, originally suggested IPC since that's the typical customer entrypoint, but yeah not strictly required

src/integrationtests/resources/com/aws/greengrass/integrationtests/deviceauth/config.yaml
MikeDombo: It can only be useful when filled with those values.

jcosentino11: gotcha. yeah agreed that it'll only complicate things, fine w/ leaving as is

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
jcosentino11: yeah this is my bad, just wanted to keep it laser focused for the original impl :P

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,29 +39,53 @@ public final class PermissionEvaluationUtils {
...
+     * @param request   Authorization Request
+     * @param session   Session
+     *
+     * @return boolean indicating if the operation requested is authorized
      */
-    public static boolean isAuthorized(String operation, String resource,
-                                       Map<String, Set<Permission>> groupToPermissionsMap) {
-        Operation op = parseOperation(operation);
-        Resource rsc = parseResource(resource);
+    public boolean isAuthorized(AuthorizationRequest request, Session session) {
```
MikeDombo, line 68: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 11. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 83-116 into a separate method.

src/integrationtests/resources/com/aws/greengrass/integrationtests/deviceauth/config.yaml
jbutler: Glad to see this feature being implemented!

+1 to Joe's initial thought - I'd suggest validating config when it's loaded and explicitly forbidding this (then you don't pay a penalty in the authZ path). Should be trivial, right? It's just a prefix match against the supported resource types.

> i don't think we enforce this though? the above policy right now is considered valid and auth passes, (`mqtt:myThing:foo` in this case)

Joe - are you saying that this still gets interpreted as a client ID resource type and allows a connection? I hope not!

jcosentino11: yep what michael said, hi @jbutler :)

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,29 +43,59 @@ public final class PermissionEvaluationUtils {
...
```
MikeDombo, line 78: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 11. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 93-126 into a separate method.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Component.java
```diff
@@ -23,7 +23,7 @@ public String getNamespace() {
     }
 
     @Override
-    public Map<String, DeviceAttribute> getDeviceAttributes() {
-        return ATTRIBUTES;
+    public DeviceAttribute getDeviceAttribute(String attributeName) {
```
jcosentino11, line 26: do we need this change? this would give us less capability than before and looks like it's a cosmetic change.  if we do need this, then i think we should add it as part of the AttributeProvider interface

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,29 +43,59 @@ public final class PermissionEvaluationUtils {
...
-    private PermissionEvaluationUtils() {
+    private static final Pattern POLICY_VARIABLE_PATTERN = Pattern.compile(POLICY_VARIABLE_FORMAT,
+            Pattern.CASE_INSENSITIVE);
+
+    private final GroupManager groupManager;
+
+    private static final Map<String, Pair<String,String>> policyVariableResolver =
+            new HashMap<String, Pair<String,String>>() {
+        {
+            put("iot:Connection.Thing.ThingName", new Pair<>("Thing", "ThingName"));
```
jcosentino11, line 56: lets make the attr namespace and properties constants

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -33,29 +43,59 @@ public final class PermissionEvaluationUtils {
...
 
-    private PermissionEvaluationUtils() {
+    private static final Pattern POLICY_VARIABLE_PATTERN = Pattern.compile(POLICY_VARIABLE_FORMAT,
+            Pattern.CASE_INSENSITIVE);
+
+    private final GroupManager groupManager;
+
+    private static final Map<String, Pair<String,String>> policyVariableResolver =
+            new HashMap<String, Pair<String,String>>() {
+        {
```
jcosentino11, line 55: nit: avoid {{}} initializers

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Component.java
robcmann: I modified the`AttributeProvider#getDeviceAttribute`interface to return only the DeviceAttribute instead of a map since in `SessionImpl#getSessionAttribute` we were getting the map just to return the DeviceAttribute

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -132,21 +191,65 @@ private static Operation parseOperation(String operationStr) {
...
+                    resource = matcher.replaceFirst(policyVariableValue);
+                    matcher = POLICY_VARIABLE_PATTERN.matcher(resource);
+                    permission = Permission.builder()
+                            .principal(permission.getPrincipal())
+                            .operation(permission.getOperation())
+                            .resource(resource).build();
+                }
+            } else {
+                logger.atWarn().kv("policyVariable", policyVariable)
+                        .log("Policy variable unsupported. Only thing name variables are supported, please fix the "
```
jcosentino11, line 246: > Only thing name variables are supported

let's remove, otherwise we'll have to update this every time we add new ones

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Component.java
jcosentino11: yeah, i guess we could always add back `getDeviceAttributes() ` if we have a real need

jcosentino11:
Putting here for later reference:

With regex on the policy eval path, throughput is 219222.33 ops/s (for variable substitution test case). Before we were hitting 400k.

We talked about doing variable eval during config time, which should get those perf numbers back up

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -32,30 +35,40 @@ public final class PermissionEvaluationUtils {
...
+     * @param request   Authorization Request
+     * @param session   Session
+     *
+     * @return boolean indicating if the operation requested is authorized
      */
-    public static boolean isAuthorized(String operation, String resource,
-                                       Map<String, Set<Permission>> groupToPermissionsMap) {
-        Operation op = parseOperation(operation);
-        Resource rsc = parseResource(resource);
+    public boolean isAuthorized(AuthorizationRequest request, Session session) {
```
MikeDombo, line 58: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 11. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test. We recommend that you simplify this method or break it into multiple methods.

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Utility method to replace policy variables in the resource with device attributes.
+     *
+     * @param policyVariables list of policy variables in resource
+     * @param resource resource to resolve
+     * @param session current device session
+     * @return updated resource
+     */
+    public static String resolvePolicyVariables(List<String> policyVariables, String resource, Session session) {
```
jcosentino11, line 43: note: in follow up we'll also need to add more validations to the policy (e.g. bad policy variable name, etc), after https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/418 is merged too

jcosentino11, line 43:
```suggestion
    public static String resolvePolicyVariables(List<String> policyVariables, String format, Session session) {
```
nit: doesn't necessarily have to be a resource

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
```diff
@@ -0,0 +1,61 @@
...
+            return resource;
+        }
+        String substitutedResource = resource;
+        for (String policyVariable : policyVariables) {
+            String attributeNamespace = policyVariableMap.get(policyVariable).getLeft();
+            String attributeName = policyVariableMap.get(policyVariable).getRight();
+            String policyVariableValue = Coerce.toString(session.getSessionAttribute(attributeNamespace,
+                    attributeName));
+            if (policyVariableValue == null) {
+                throw new IllegalArgumentException("No attribute found for current session");
```
jcosentino11, line 54: lets avoid unchecked exceptions, may need to make your own if there's not a suitable one

src/main/java/com/aws/greengrass/clientdevices/auth/exception/AttributeProviderException.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.clientdevices.auth.exception;
+
+public class AttributeProviderException extends Exception {
```
jcosentino11, line 8: nit: how about `PolicyException`?

src/main/java/com/aws/greengrass/clientdevices/auth/iot/Thing.java
```diff
@@ -164,11 +164,8 @@ public String getNamespace() {
 
     @Override
     public DeviceAttribute getDeviceAttribute(String attributeName) {
-        if ("ThingName".equals(attributeName)) {
-            return new WildcardSuffixAttribute(thingName);
-        }
-        // TODO: Return other possible DeviceAttributes
-        return null;
+        // TODO: Support other DeviceAttributes
+        return new WildcardSuffixAttribute(thingName);
```
jcosentino11, line 168: from convo: if we want to still have attributeName passed in we'll need to use it, otherwise would need to be reverted back to just `getDeviceAttributes()` that returns a map

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
```diff
@@ -5,57 +5,53 @@
...
+            return format;
         }
-        String substitutedResource = resource;
+        String substitutedFormat = format;
         for (String policyVariable : policyVariables) {
-            String attributeNamespace = policyVariableMap.get(policyVariable).getLeft();
-            String attributeName = policyVariableMap.get(policyVariable).getRight();
-            String policyVariableValue = Coerce.toString(session.getSessionAttribute(attributeNamespace,
-                    attributeName));
+            String attributeNamespace = policyVariableToAttributeProvider.get(policyVariable).getLeft();
```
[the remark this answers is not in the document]
robcmann: Should handling unsupported policy vars happen at config time with the validation in #418 so that we don't need to worry about it during eval time?

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
```diff
@@ -0,0 +1,59 @@
...
+     * Policy variables need to be validated when reading the policy document.
+     * This method does not handle unsupported policy variables.
+     *
+     * @param policyVariables list of policy variables in permission format
+     * @param format permission format to resolve
+     * @param session current device session
+     * @return updated format
+     * @throws PolicyException when unable to find a policy variable value
+     */
+    public static String resolvePolicyVariables(List<String> policyVariables, String format, Session session)
```
jcosentino11, line 39: you mentioned before, but we can make this a set now

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
robcmann: done

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
```diff
@@ -0,0 +1,59 @@
...
+    public static String resolvePolicyVariables(Set<String> policyVariables, String format, Session session)
+            throws PolicyException {
+        if (policyVariables.isEmpty()) {
+            return format;
+        }
+        String substitutedFormat = format;
+        for (String policyVariable : policyVariables) {
+            String attributeNamespace = policyVariableToAttributeProvider.get(policyVariable).getLeft();
+            String attributeName = policyVariableToAttributeProvider.get(policyVariable).getRight();
+            String policyVariableValue = session.getSessionAttribute(attributeNamespace, attributeName).toString();
```
MikeDombo, line 48: use `Coerce.toString` to avoid NPE

MikeDombo (approved):

jcosentino11 (approved):

PR #416, opened 2024-01-09
feat: basic authn benchmark

jcosentino11:
**Issue #, if available:**

**Description of changes:**

* New benchmark that tests CDA's authn flow "canDevicePerform"
* Remove example benchmarks that test AST since those aren't actually testing a hot path
* Add github action to run benchmarks and upload results
image


benchmark/pom.xml
```diff
@@ -28,6 +28,17 @@
...
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.22</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.aws.greengrass</groupId>
+            <artifactId>nucleus</artifactId>
+            <version>2.6.0-SNAPSHOT</version>
```
MikeDombo, line 40: for benchmarking, any reason not to use a newer version?

jcosentino11: Just wanted to keep nucleus version consistent with what CDA uses--jmh uses CDA jar, and nucleus is a "provided" dep so we include it explicitly here

benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
```diff
@@ -37,7 +37,7 @@
 
 
 @BenchmarkMode(Mode.AverageTime)
-@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@OutputTimeUnit(TimeUnit.SECONDS)
```
MikeDombo, line 40: this will still give seconds per operation right? Rather than ops per second

jcosentino11: oh right, read it too quickly, so throughput then? yep will adjust

MikeDombo:
~400K ops/s sounds pretty good. This is the current code base right which contains none of the extra wildcard and policy variable stuff right?

jcosentino11:
> ~400K ops/s sounds pretty good. This is the current code base right which contains none of the extra wildcard and policy variable stuff right?

yep, no wildcard/var stuff yet

benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Warmup(iterations = 0) // TODO for testing
+@Measurement(iterations = 1)
```
MikeDombo, line 42: I think we'll want more than 1 iteration to remove outliers.

jcosentino11: yeah i had it turned down for now since im cleaning up the script /testing

jcosentino11: looks like we're good now, script auto cleans up comments now

MikeDombo (dismissed):

robcmann (approved):

benchmark/src/main/java/com/aws/greengrass/clientdevices/auth/benchmark/AuthorizationBenchmarks.java
MikeDombo: Rather than cleaning, it would be best to replace any existing comment like the code coverage does. This will prevent emails for each new benchmark result.

jcosentino11: Ah right, was just thinking about keeping the latest results at the bottom but the emails are for sure annoying ;p will fix

MikeDombo (approved):

robcmann (approved):

PR #418, opened 2024-01-19
refactor: policy validation on config change

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Add an explicit `GroupConfiguration#validate` method, which is used to validate incoming policy config changes before applying them to CDA.

Making this change as the foundation for adding more validations in the future (for policy variable feature).


MikeDombo:

robcmann (dismissed):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -68,6 +71,15 @@
 @ImplementsService(name = ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME)
 public class ClientDevicesAuthService extends PluginService {
     public static final String CLIENT_DEVICES_AUTH_SERVICE_NAME = "aws.greengrass.clientdevices.Auth";
+    private static final ObjectMapper MAPPER = JsonMapper.builder()
```
MikeDombo, line 74: use the mappers we already have in the utils. We have one already for recipes which does case insensitive stuff

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -174,7 +186,7 @@ private void configChangeHandler(WhatHappened whatHappened, Node node) {
         if (whatHappened == WhatHappened.timestampUpdated || whatHappened == WhatHappened.interiorAdded) {
             return;
         }
-        logger.atDebug().kv("why", whatHappened).kv("node", node).log();
+        logger.atDebug().kv("why", whatHappened).kv(KV_NODE, node).log();
```
MikeDombo, line 189: drop this to trace while we're here

src/test/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthServiceTest.java
```diff
@@ -171,8 +171,7 @@ void GIVEN_bad_group_configuration_WHEN_start_service_THEN_service_in_error_stat
         ignoreExceptionOfType(context, IllegalArgumentException.class);
         ignoreExceptionOfType(context, UnrecognizedPropertyException.class);
 
-        startNucleusWithConfig("badGroupConfig.yaml", State.ERRORED);
-
+        startNucleusWithConfig("badGroupConfig.yaml", State.RUNNING);
```
MikeDombo, line 174: so we're changing the behavior now that it will be running rather than errored? Is that desired?

MikeDombo: I think we should be erroring out to cause a failed and rolled back deployment rather than silently failing.

jcosentino11: will adjust

pom.xml
```diff
@@ -76,6 +76,12 @@
             <version>2.6.0-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
MikeDombo, line 79: shouldn't need to add this, it comes via the nucleus dependency chain

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -68,6 +70,8 @@
 @ImplementsService(name = ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME)
 public class ClientDevicesAuthService extends PluginService {
     public static final String CLIENT_DEVICES_AUTH_SERVICE_NAME = "aws.greengrass.clientdevices.Auth";
+    private static final ObjectMapper MAPPER = SerializerFactory.getRecipeSerializerJson();
```
MikeDombo, line 73: I think you want `getConfigurationSerializerJson` to avoid changing capitalization

pom.xml
jcosentino11: yeah was trying w/o it at first but my IDE wouldn't recognize, maybe i missed something

MikeDombo: If you remove, does it still compile?

jcosentino11: Yeah I removed component-common from my maven cache and did a maven build and it worked, and looks like it compiled fine in GitHub too

MikeDombo (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
MikeDombo: We shouldn't fail on unknown though. That prevents forward compatibility

jcosentino11: yeah, ok if we're fine with that behavior change then yeah can add it in

src/main/java/com/aws/greengrass/clientdevices/auth/configuration/GroupConfiguration.java
```diff
@@ -114,4 +107,21 @@ private Set<String> findPolicyVariables(String resource) {
         }
         return policyVariables;
     }
+
+    /**
+     * Validate the deviceGroups configuration.
+     *
+     * @throws PolicyException if an invalid policy is detected
+     */
+    public void validate() throws PolicyException {
```
robcmann, line 116: Should we also validate that policy variables in policy document are supported by the `PolicyVariableResolver` since the resolver doesn't handle unsupported policy variables?

jcosentino11: yep, wanted to do that in a follow-up, this PR is meant to be a refactor

src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
jcosentino11: done

robcmann (approved):

sameerzuberi (approved):

PR #422, opened 2024-02-07
chore: validate policy variables

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Validate that policy resources only contain expected policy variables. Had to modify the `${}` regex to accept anything in order to detect bad cases.  Also made some changes to support case-insensitive variables (willing to remove if we think we don't need case insensitivity)

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
@@ -20,7 +20,7 @@ public final class PolicyVariableResolver {
     private static final String THING_NAME_ATTRIBUTE = "ThingName";
 
     private static final Map<String, Pair<String,String>> policyVariableToAttributeProvider = ImmutableMap.of(
-            "${iot:Connection.Thing.ThingName}".toLowerCase(), new Pair<>(THING_NAMESPACE, THING_NAME_ATTRIBUTE)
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/clientdevices/auth/configuration/PolicyVariableResolver.java
```diff
@@ -44,8 +44,8 @@ public static String resolvePolicyVariables(Set<String> policyVariables, String
         }
         String substitutedFormat = format;
         for (String policyVariable : policyVariables) {
-            String attributeNamespace = policyVariableToAttributeProvider.get(policyVariable).getLeft();
-            String attributeName = policyVariableToAttributeProvider.get(policyVariable).getRight();
+            String attributeNamespace = policyVariableToAttributeProvider.get(policyVariable.toLowerCase()).getLeft();
```
MikeDombo, line 47: is iot core case insensitive? I'm not so sure we want to be.

jcosentino11: just manually tested, case-sensitive, will remove this

jcosentino11: done

robcmann (approved):

MikeDombo (approved):

PR #425, opened 2024-02-13
fix: move config validation to startup stage so bad configs will break the component

robcmann:
**Description of changes:**
Move config subscription and validation from `install` to `startup` stage.

Moving the config validation to the startup stage will prevent CDA from ever reaching the RUNNING state if the config is bad.  


src/main/java/com/aws/greengrass/clientdevices/auth/ClientDevicesAuthService.java
```diff
@@ -214,6 +213,7 @@ private void configChangeHandler(WhatHappened whatHappened, Node node) {
     @Override
     protected void startup() throws InterruptedException {
         context.get(CertificateManager.class).startMonitors();
+        subscribeToConfigChanges();
```
MikeDombo, line 216: subscribe registers a config change callback. That callback needs to be saved as a final field so that we don't duplicate callbacks

jcosentino11 (approved):

MikeDombo (approved):

PR #430, opened 2024-03-22
fix: remove invalid certificates if validation fails

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Remove stored certificate if validation fails. This would happen automatically on background refresh, but take care of it immediately now.

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
@@ -55,7 +55,7 @@ public class VerifyIotCertificate implements UseCases.UseCase<Boolean, String> {
         // so this guarantees that we at least try once.
         // Else, rely on whatever is in the local registry.
         Optional<Certificate> cloudCert = Optional.empty();
-        Certificate cert;
+        Certificate cert = null;
 
         try {
             cert = certificateRegistry.getOrCreateCertificate(certificatePem);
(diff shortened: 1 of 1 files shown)


alter-mage (dismissed):

alter-mage (approved):

jcosentino11 (approved):

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -72,10 +75,11 @@ public Boolean apply(String certificatePem) {
             cert = cloudCert.get();
             if (cert.isActive()) {
                 certificateRegistry.updateCertificate(cloudCert.get());
-            } else {
-                certificateRegistry.deleteCertificate(cloudCert.get());
             }
         }
+        if (!cert.isActive()) {
+            certificateRegistry.deleteCertificate(cert);
+        }
```
saranyailla, line 82: If the `cloudCert` is active at L76 but becomes inactive at L80, then we're only removing the `cloudCert` that was added in L77. The previous `cert` may still exist in store because of L75.

src/main/java/com/aws/greengrass/clientdevices/auth/iot/usecases/VerifyIotCertificate.java
```diff
@@ -64,6 +64,9 @@ public Boolean apply(String certificatePem) {
             }
         } catch (InvalidCertificateException e) {
             logger.atWarn().kv("certificatePem", certificatePem).log("Unable to process certificate", e);
+            if (cert != null) {
+                certificateRegistry.deleteCertificate(cert);
+            }
```
saranyailla, line 69: Nit: Not sure if this is ever gonna be used as L61 is going to throw this exception first if the `certificatePem` is invalid. Although L63 can throw this exception for the same reason.

MikeDombo: Unlikely, but covering all the cases.

PR #435, opened 2024-08-09
chore(fips): update endpoint overwrite method

yitingb:
**Issue #, if available:**

**Description of changes:**
1. Updating the dependencies
2. Update method to provide ggv2dataclient endpoint
3. Ignoring device offline exceptions in unit testing since we're targeting authentication and the device offline related exceptions are expected

It's used to support fips feature where the endpoint should be override earlier than built-in useFips function
With UAT

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
@@ -20,6 +20,7 @@ import com.aws.greengrass.clientdevices.auth.iot.IotAuthClientFake;
 import com.aws.greengrass.clientdevices.auth.iot.NetworkStateFake;
 import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.lifecyclemanager.Kernel;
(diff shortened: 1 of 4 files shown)


junfuchen99 (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/certificateauthority/OfflineAuthenticationTest.java
```diff
@@ -260,6 +266,8 @@ void GIVEN_clientConnectsWhileOnline_WHEN_offlineAndCertificateRevoked_THEN_back
     void GIVEN_clientConnectsWhileOnline_WHEN_offlineAndCertDetachedFromThing_THEN_backOnlineAndClientRejected(
             ExtensionContext context) throws Exception {
         ignoreExceptionOfType(context, NoSuchFileException.class);
+        ignoreExceptionOfType(context, MqttRequestException.class);
+        ignoreExceptionOfType(context, ExecutionException.class);
```
saranyailla, line 270: Nit: we can assert the exception as well.

saranyailla (approved):

saranyailla (approved):

junfuchen99 (approved):

PR #436, opened 2024-09-11
feat: mqtt-style resource wildcard matching

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Proposes a new configuration option, `enableMqttWildcardEvaluation`, to allow policy resources to be evaluated using MQTT wildcard.

e.g.

```
       policies:
          thingAccessPolicy:
            subscribe:
              statementDescription: "mqtt subscribe"
              operations:
                - "mqtt:subscribe"
              resources:
                - "mqtt:topic:${iot:Connection.Thing.ThingName}/+/test/#"
```

Allows client devices to subscribe to `thingName/path/test/path1/path2`, for example

### DIFF
--- a/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
+++ b/src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
@@ -6,6 +6,7 @@
 package com.aws.greengrass.clientdevices.auth;
 
 import com.aws.greengrass.authorization.WildcardTrie;
+import com.aws.greengrass.clientdevices.auth.configuration.CDAConfiguration;
 import com.aws.greengrass.clientdevices.auth.configuration.GroupManager;
 import com.aws.greengrass.clientdevices.auth.configuration.Permission;
 import com.aws.greengrass.clientdevices.auth.exception.PolicyException;
@@ -14,6 +15,7 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Utils;
 import lombok.Builder;
+import lombok.Setter;
 import lombok.Value;
 import org.apache.commons.lang3.StringUtils;
 
@@ -36,6 +38,8 @@ public final class PermissionEvaluationUtils {
             "Resource is malformed, must be of the form: "
             + "([a-zA-Z]+):([a-zA-Z]+):" + RESOURCE_NAME_PATTERN.pattern();
     private final GroupManager groupManager;
+    @Setter
+    private volatile CDAConfiguration cdaConfiguration;
 
     /**
      * Constructor for PermissionEvaluationUtils.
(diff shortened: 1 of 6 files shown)


MikeDombo (commented):
This was purposefully not included because it will differ from how IoT Core works. If you have permission to subscribe to `A/#` then you have permission to subscribe to that and _exactly_ that.

jcosentino11:
> This was purposefully not included because it will differ from how IoT Core works. If you have permission to subscribe to `A/#` then you have permission to subscribe to that and _exactly_ that.

Ah I see https://docs.aws.amazon.com/iot/latest/developerguide/pub-sub-policy.html now, looks like `?` is used for pseduo-`+` matching

MikeDombo:
> > This was purposefully not included because it will differ from how IoT Core works. If you have permission to subscribe to `A/#` then you have permission to subscribe to that and _exactly_ that.
> 
> Ah I see https://docs.aws.amazon.com/iot/latest/developerguide/pub-sub-policy.html now, looks like `?` is used for pseduo-`+` matching

Well yeah, but that's for IPC which isn't MQTT :)
So.... reasons, but maybe not good reasons.

jcosentino11:
> > > This was purposefully not included because it will differ from how IoT Core works. If you have permission to subscribe to `A/#` then you have permission to subscribe to that and _exactly_ that.
> > 
> > 
> > Ah I see https://docs.aws.amazon.com/iot/latest/developerguide/pub-sub-policy.html now, looks like `?` is used for pseduo-`+` matching
> 
> Well yeah, but that's for IPC which isn't MQTT :) So.... reasons, but maybe not good reasons.

Should be for MQTT/HTTP/Websockets?  IoT Core publish/subscribe, unless im missing something here

MikeDombo:
> > > > This was purposefully not included because it will differ from how IoT Core works. If you have permission to subscribe to `A/#` then you have permission to subscribe to that and _exactly_ that.
> > > 
> > > 
> > > Ah I see https://docs.aws.amazon.com/iot/latest/developerguide/pub-sub-policy.html now, looks like `?` is used for pseduo-`+` matching
> > 
> > 
> > Well yeah, but that's for IPC which isn't MQTT :) So.... reasons, but maybe not good reasons.
> 
> Should be for MQTT/HTTP/Websockets? IoT Core publish/subscribe, unless im missing something here

Not too sure what you mean. I made the IPC policy more lenient to avoid common customer misunderstandings when compared with V1. But we wanted to keep CDA aligned with IoT Core as much as possible.

jcosentino11:
> > > > > This was purposefully not included because it will differ from how IoT Core works. If you have permission to subscribe to `A/#` then you have permission to subscribe to that and _exactly_ that.
> > > > 
> > > > 
> > > > Ah I see https://docs.aws.amazon.com/iot/latest/developerguide/pub-sub-policy.html now, looks like `?` is used for pseduo-`+` matching
> > > 
> > > 
> > > Well yeah, but that's for IPC which isn't MQTT :) So.... reasons, but maybe not good reasons.
> > 
> > 
> > Should be for MQTT/HTTP/Websockets? IoT Core publish/subscribe, unless im missing something here
> 
> Not too sure what you mean. I made the IPC policy more lenient to avoid common customer misunderstandings when compared with V1. But we wanted to keep CDA aligned with IoT Core as much as possible.

I think we're talking about two different things 😅   Was saying that https://docs.aws.amazon.com/iot/latest/developerguide/pub-sub-policy.html  and `?` matching is IoT Core specific

src/main/java/com/aws/greengrass/clientdevices/auth/PermissionEvaluationUtils.java
```diff
@@ -134,9 +138,27 @@ private boolean compareResource(Resource requestResource, String policyResource)
             return true;
         }
 
-        WildcardTrie wildcardTrie = new WildcardTrie();
-        wildcardTrie.add(policyResource);
-        return wildcardTrie.matchesStandard(requestResource.getResourceStr());
+        if (matchMqttWildcards()) {
```
MikeDombo, line 141: Switching comment to threads to make it more sane.


My point is that IoT Core for MQTT policies does not let you subscribe to A/B when the policy is `A/#`. Greengrass PubSub does allow this, but CDA does not in order to match the behavior of IoT Core.

jcosentino11: Okay yeah I understand your point now.  So to have parity with IoT Core then, `?` single char matching would be needed, so then CDA knows about `*` and `?`

MikeDombo: Yeah guess so, if iot core just added that in.

MikeDombo: But wildcards `#` and `+` shouldn't change the way they currently work.

jcosentino11: Yep will either close or revise this. Reason it came up was customer ask so was exploring the idea

jcosentino11:
Closing in favor of `?` matching: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/437

